PR #4, opened 2023-08-03
fix: split role tests on daemonset

fmoral2:
#### Proposed Changes ####

- fix: split role tests on daemonset

#### Types of Changes ####

- Change in the daemonset test logic

#### Verification ####

- Tests ran 5 times in both products

#### Testing ####

- Ran each test with all configurations 4 times- OK
- Code lint: OK
- Make file | Docker Run | Jenkins updated: NOK  -> see here https://github.com/rancher/qa-tasks/issues/854


VestigeJ (dismissed):
lgtm

mdrahman-suse (dismissed):
LGTM

entrypoint/versionbump/cniplugin_test.go
```diff
@@ -49,8 +49,8 @@ var _ = Describe("VersionTemplate Upgrade:", func() {
 			TestCombination: &template.RunCmd{
 				Run: []template.TestMap{
 					{
-						Cmd: "kubectl get -n test-pod-bandwidth  pod test-pod -o yaml : | grep -A2 annotations ," +
-							" k3s -v",
+						Cmd: "/var/lib/rancher/k3s/data/current/bin/cni, /var/lib/rancher/k3s/data/current/bin/flannel",
```
mdrahman-suse, line 52: Will this work for rke2?

fmoral2: nope, since rke cni version is another path , or we should add a new test inside for rke2, or use the bump version test  not any particular   {subject}_test.go in this folder.

i just added one test for k3s ( cni ) and one for rke2 ( runc ) as an working example on how to create a test on the bump version folder. 

But TBH this is just one more reason that i dont like that....i prefer use the version_bump.test  because of that also,  you can pass any cmd you want not being tied to a test like this one.

if its somehow confused we cant talk later

VestigeJ (approved):
LGTM

mdrahman-suse (approved):
LGTM

entrypoint/versionbump/cniplugin_test.go
```diff
@@ -44,13 +44,18 @@ var _ = Describe("VersionTemplate Upgrade:", func() {
 		}
 	})
 
+	var cmd string
+	if cfg.Product == "rke2" {
+		cmd = "sudo cat /etc/cni/net.d/10-canal.conflist"
```
rancher-max, line 49: I think for now it might be better to just skip the test for rke2 entirely, since technically rke2 supports:
- canal
- calico
- cilium
- multus (with any of the three above)
- sriov (with multus)
and these could all have relevant data to validate in different locations 😭

pkg/testcase/daemonset.go
```diff
@@ -17,17 +21,77 @@ func TestDaemonset(deployWorkload bool) {
...
+	nodeNames, err := shared.RunCommandHost(cmd)
+	if err != nil {
+		return
+	}
+
+	var nodes []string
+	n := strings.Split(nodeNames, "\n")
+	for _, nodeName := range n {
+		if nodeName != "" {
+			nodes = append(nodes, nodeName)
```
rancher-max, line 41: We can probably use `ParsePods` func here, but maybe modify it to accept a namespace. I don't know if it's necessary, but it has some re-usability in that case. I'm fine with this though as it works and is clear what data it is grabbing too 👍

pkg/testcase/daemonset.go
```diff
@@ -17,17 +21,77 @@ func TestDaemonset(deployWorkload bool) {
...
 	Eventually(func(g Gomega) int {
 		return shared.CountOfStringInSlice("test-daemonset", pods)
-	}, "420s", "5s").Should(Equal(len(allNodes)),
+	}, "400s", "5s").Should(Equal(len(nodes)),
 		"Daemonset pod count does not match node count")
+
+}
+
+// validateNodes checks if the nodes in the two strings are equal (ignoring order through sorting).
+func validateNodes(taints, names string) bool {
```
ShylajaDevadiga, line 66: We could name it to something more specific on what it is doing.

pkg/testcase/daemonset.go
```diff
@@ -17,17 +21,77 @@ func TestDaemonset(deployWorkload bool) {
...
+		--kubeconfig="%s" | grep '<none>'
+		`,
+		shared.KubeConfigFile,
+	)
+	taints, err := shared.RunCommandHost(cmd)
+	if err != nil {
+		return
+	}
+	Expect(taints).To(ContainSubstring("<none>"))
+	Expect(validateNodes(strings.TrimSpace(nodeNames), strings.TrimSpace(taints))).To(BeTrue())
```
ShylajaDevadiga, line 56: params are passed in a different order than it is received. functionality would not be affected irrespective i guess

README.md
```diff
@@ -313,7 +307,7 @@ $ make test-create
 
 
 - Run upgrade cluster test:
-$ make test-upgrade-manual INSTALLTYPE=INSTALL_K3S_COMMIT=257fa2c54cda332e42b8aae248c152f4d1898218
+$ make test-upgrade-manual INSTALLTYPE=257fa2c54cda332e42b8aae248c152f4d1898218
```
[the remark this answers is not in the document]
fmoral2: agreed

entrypoint/versionbump/cniplugin_test.go
fmoral2: ok i will just leave for k3s then

pkg/testcase/daemonset.go
ShylajaDevadiga: can we change names to nodeNames. That way we don't have to go back to see what is being passed.

pkg/testcase/daemonset.go
fmoral2: parse pods for doing what? sorry

pkg/testcase/daemonset.go
fmoral2: sure

pkg/testcase/daemonset.go
fmoral2: yeah, but better correct,  doing now.
tks

README.md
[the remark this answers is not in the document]
rancher-max: https://github.com/rancher/qa-tasks/issues/587

pkg/testcase/daemonset.go
rancher-max: That function currently looks like it retrieves the Node name as well, so it could potentially save a bit of logic in this case if that function could take in a namespace and return the struct of pods as it does already. For easy reference: https://github.com/rancher/distros-test-framework/blob/main/shared/cluster.go#L263-L292

fmoral2: isnt the same as here  ? 
```

	cmd := fmt.Sprintf(`
		kubectl get pods -n test-daemonset -o wide --kubeconfig="%s" \
		| grep -A10 NODE | awk 'NR>1 {print $7}'
		`,
		shared.KubeConfigFile,
	)
	nodeNames, err := shared.RunCommandHost(cmd)
	if err != nil {
		return
	}
```

rancher-max: yeah the end result is still achieved in this case! I don't think you need to change anything if you don't want to here, but just thought it might be helpful for this and potentially future work if we ever need to retrieve similar data on pods, then we could just use ParsePods and pass the desired namespace

rancher-max (approved):

ShylajaDevadiga (approved):

PR #6, opened 2023-08-04
[READY FOR REVIEW] Add windows test and modules

mdrahman-suse:
#### Proposed Changes ####

Adding ability to create cluster and join Windows agent. FYI this works only with non-hardened cluster setup as hardening flags are not supported in Windows, yet!
ONLY SUPPORTED IN RKE2

Yes, Additional parameters needed for creating the Windows instance in AWS
In `rke2.tfvars` file, add the below vars
```
server_flags   = "cni: calico"
windows_ec2_instance_class = "<use t3.xlarge or higher>"
windows_aws_ami = "<windows ami>"
no_of_windows_worker_nodes = <count of windows node>
```

#### Types of Changes ####

No changes in RKE2 development code, only in acceptance testing code

#### Verification ####

Execute the test locally using the tfvars file or use jenkins


config/config.yaml
```diff
@@ -0,0 +1,2 @@
+ENV_PRODUCT: rke2
```
fmoral2, line 1: on the current master this is already ignored on git to facilitate our lifes not needing to add this everytime on PRs

Makefile
```diff
@@ -42,8 +47,8 @@ test-complete: test-env-clean test-env-down remove-tf-state test-env-up test-run
 
 .PHONY: remove-tf-state
 remove-tf-state:
-	@rm -rf ./modules/k3scluster/.terraform
-	@rm -rf ./modules/k3scluster/.terraform.lock.hcl ./modules/k3scluster/terraform.tfstate ./modules/k3scluster/terraform.tfstate.backup
+	@rm -rf ./modules/rke2/.terraform
+	@rm -rf ./modules/rke2/.terraform.lock.hcl ./modules/rk2/terraform.tfstate ./modules/rke2/terraform.tfstate.backup
```
fmoral2, line 52: since for now we still have both products in model dir , we should or have to make commands like :

remove-tf-state-rke2
remove-tf-state-k3s

or if you want one command we should pass the product as arg, and based on that go to the specific dir.

config/config.yaml
fmoral2, line 1: we should add on the README. , the informations to running this like....rke2 should not be hardened and etc ...

entrypoint/createcluster/createcluster_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 )
 
+var arch = customflag.ServiceFlag.ClusterConfig.Arch.String()
```
fmoral2, line 14: lets follow the same approach in all suites...if you want to use this as a var, create it on the suite file, not in the test it self....lets try at max keep this file clean as possible it can be.

Makefile
mdrahman-suse: Yeah I vote for one command, will update

shared/cluster.go
```diff
@@ -34,52 +35,64 @@ type Pod struct {
...
+			if err != nil {
+				return "", fmt.Errorf("failed to apply workload %s: %s", workload, err)
 			}
-			return res, err
+		} else if action == "delete" {
+			err = deleteWorkload(workload, filename)
+			if err != nil {
+				return "", fmt.Errorf("failed to delete workload %s: %s", workload, err)
+			}
+		} else {
```
fmoral2, line 68: same as above....you can return the error directly , because you are repeating your self from line 55 

something like 
```

} else {
				err = deleteWorkload(workload, filename)
	if err != nil {
			return "", fmt.Errorf("failed to delete workload %s: %s", workload, err)
		}
```

shared/cluster.go
```diff
@@ -34,52 +35,64 @@ type Pod struct {
...
 
-	return "", fmt.Errorf("workload %s not found", workload)
+	return res, err
 }
 
-func createWorkload(workload, filename string) (string, error) {
-	fmt.Println("\nDeploying", workload)
+// applyWorkload applies a workload to the cluster.
+func applyWorkload(workload, filename string) (error) {
+	fmt.Println("\nApplying ", workload)
```
fmoral2, line 78: add one line from space the code here

shared/cluster.go
```diff
@@ -34,52 +35,64 @@ type Pod struct {
...
 	cmd := "kubectl apply -f " + filename + " --kubeconfig=" + KubeConfigFile
+	out, err := RunCommandHost(cmd)
```
fmoral2, line 80: this RunCommandHost functions, accepts variadcs inputs args. 
you can pass both commands in the same call , if you dont need any extra stuff between both

entrypoint/createcluster/createcluster_test.go
fmoral2: please also add this on the other tests places that uses it....to stabilish the same pattern.

factory/config.go
```diff
@@ -79,22 +81,27 @@ func addClusterConfig(
...
-	ServerIPs := strings.Split(terraform.Output(g, terraformOptions, "master_ips"), ",")
-	c.ServerIPs = ServerIPs
-
-	shared.AwsUser = terraform.GetVariableAsStringFromVarFile(g, varDir, "aws_user")
-	shared.AccessKey = terraform.GetVariableAsStringFromVarFile(g, varDir, "access_key")
+	rawAgentIPs = terraform.Output(g, terraformOptions, "windows_worker_ips")
+	if rawAgentIPs != "" {
+		agentIPs = strings.Split(rawAgentIPs, ",")
+	}
+	c.WinAgentIPs = agentIPs
```
fmoral2, line 104: this logic here will not work when u are not creating windowns ips.....

you should do something like
```
func addClusterConfig(
	cfg *config.ProductConfig,
	g GinkgoTInterface,
	varDir string,
	terraformOptions *terraform.Options,
) (*Cluster, error) {
	c := &Cluster{}
	var agentIPs []string
	var winAgentIPs []string

	if cfg.Product == "k3s" {
		c.ClusterType = terraform.GetVariableAsStringFromVarFile(g, varDir, "cluster_type")
		c.ExternalDb = terraform.GetVariableAsStringFromVarFile(g, varDir, "external_db")
		c.RenderedTemplate = terraform.Output(g, terraformOptions, "rendered_template")
	}

	shared.KubeConfigFile = terraform.Output(g, terraformOptions, "kubeconfig")
	shared.AwsUser = terraform.GetVariableAsStringFromVarFile(g, varDir, "aws_user")
	shared.AccessKey = terraform.GetVariableAsStringFromVarFile(g, varDir, "access_key")
	c.ProductType = cfg.Product

	serverIPs := strings.Split(terraform.Output(g, terraformOptions, "master_ips"), ",")
	c.ServerIPs = serverIPs

	rawAgentIPs := terraform.Output(g, terraformOptions, "worker_ips")
	if rawAgentIPs != "" {
		agentIPs = strings.Split(rawAgentIPs, ",")
	}
	c.AgentIPs = agentIPs

	rawAgentWinIPs := terraform.Output(g, terraformOptions, "windows_worker_ips")
	if rawAgentWinIPs != "" {
		winAgentIPs = strings.Split(rawAgentWinIPs, ",")
	}
	c.WinAgentIPs = winAgentIPs

	return c, nil
}
```

fmoral2 (changes requested):
https://github.com/rancher/distros-test-framework/pull/6/files#r1286215353

workloads/pod_client.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apps/v1
```
fmoral2, line 1: should we create a folder for windowns or no need ?

Makefile
```diff
@@ -68,6 +73,10 @@ test-version-bump:
 	  -installVersionOrCommit ${INSTALLTYPE} -channel ${CHANNEL} \
 	  -testCase "${TESTCASE}" -deployWorkload ${DEPLOYWORKLOAD} -workloadName ${WORKLOADNAME} -description "${DESCRIPTION}"
 
+.PHONY: test-create-mixedos
+test-create-mixedos:
+	@go test -timeout=45m -v ./entrypoint/mixedoscluster/...
+
```
fmoral2, line 79: please delete all commands bellow if they are not going to be used anymore

pkg/testcase/node.go
```diff
@@ -33,8 +32,70 @@ func TestNodeStatus(
...
+	}, "1200s", "60s").Should(Succeed())
 
+	fmt.Println("\nRetrieving Cluster nodes:")
 	_, err := shared.ParseNodes(true)
 	Expect(err).NotTo(HaveOccurred())
 }
+
+// TestInternodeConnectivityMixedOS Deploys services in the cluster and validates communication between linux and windows nodes
+func TestInternodeConnectivityMixedOS() {
+	_, err := shared.ManageWorkload("apply", "",
```
fmoral2, line 44: here you should be passing the arch value, not empty

entrypoint/mixedoscluster/mixedoscluster_suite_test.go
```diff
@@ -0,0 +1,37 @@
...
+	os.Exit(m.Run())
+}
+
+func TestMixedOSClusterCreateSuite(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecs(t, "Create Mixed OS Cluster Test Suite")
+}
+
+var _ = AfterSuite(func() {
```
fmoral2, line 30: can you add on the after suite or somewhere else you think is suitable, a automatic cleanup of the sonobuoy? 
instead of only being called if i say delete caliing the mixedos_sonobuoy.sh ?

Makefile
mdrahman-suse: Updated

workloads/pod_client.yaml
mdrahman-suse: no need atm and I was leaning towards even moving out the amd64 folder as lot of workloads are already copied in to arm64

shared/cluster.go
mdrahman-suse: ok will check that

shared/cluster.go
```diff
@@ -34,52 +35,64 @@ type Pod struct {
 	Node      string
 }
 
-// ManageWorkload creates or deletes a workload based on the action: create or delete.
-func ManageWorkload(action, workload, arch string) (string, error) {
+// ManageWorkload applies or deletes a workload based on the action: apply or delete.
+func ManageWorkload(action string, arch string, workloads ...string) (string, error) {
+	if action != "apply" && action != "delete" {
```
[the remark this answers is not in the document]
mdrahman-suse: good call, updated

entrypoint/mixedoscluster/mixedoscluster_test.go
```diff
@@ -0,0 +1,51 @@
...
+		)
+	})
+
+	It("Validates internode connectivity over the vxlan tunnel", func() {
+		testcase.TestInternodeConnectivityMixedOS()
+	})
+
+	It("Validates cluster by running sonobuoy mixed OS plugin", func() {
+		testcase.TestSonobuoyMixedOS(sonobuoyVersion, true)
+		shared.ManageWorkload("delete", "",
```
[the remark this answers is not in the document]
fmoral2: the point is not that is not needed....is that is a parameter that we shouldnt passing around empty since we always have a default value. 

if the point is having this passing as empty , when u pass the flag to start the arch == arm, how does this parameter will know that ? 

even if this will not be working for arm.

some examples 

```
func TestIngress(deployWorkload bool) {
	if deployWorkload {
		_, err := shared.ManageWorkload(
			"create",
			"ingress.yaml",
			customflag.ServiceFlag.ClusterConfig.Arch.String(),
		)
		Expect(err).NotTo(HaveOccurred(),
			"Ingress manifest not deployed")
	}
```

```	
if deployWorkload {
		_, err := shared.ManageWorkload(
			"create",
			"local-path-provisioner.yaml",
			customflag.ServiceFlag.ClusterConfig.Arch.String(),
		)
		Expect(err).NotTo(HaveOccurred(),
			"local-path-provisioner manifest not deployed")
	}
```

```
func TestServiceNodePort(deployWorkload bool) {
	if deployWorkload {
		_, err := shared.ManageWorkload(
			"create",
			"nodeport.yaml",
			customflag.ServiceFlag.ClusterConfig.Arch.String(),
		)
		Expect(err).NotTo(HaveOccurred(),
		
```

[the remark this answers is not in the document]
fmoral2: on the coredns.go file, i am passing it wrongly 
```
	if deployWorkload {
		_, err := shared.ManageWorkload("create", "dnsutils.yaml", "arch")
		Expect(err).NotTo(HaveOccurred(),
			"dnsutils manifest not deployed", err)
	}

```

can u fiz for me in your PR ?

entrypoint/mixedoscluster/mixedoscluster_suite_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"testing"
+
+	"github.com/rancher/distros-test-framework/pkg/customflag"
+	"github.com/rancher/distros-test-framework/factory"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var sonobuoyVersion string
```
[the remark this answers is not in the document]
mdrahman-suse: This is fine here

shared/cluster.go
mdrahman-suse: Actually I wanted to get some suggestion on this. I want to wait for sometime after the workload is applied as in some cases it takes a while for the container to be created... so I was thinking about using Eventually... wdyt?

entrypoint/mixedoscluster/mixedoscluster_test.go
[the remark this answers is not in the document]
mdrahman-suse: Lets discuss this offline or on standup

shared/cluster.go
mdrahman-suse: Updated

Makefile
```diff
@@ -1,39 +1,44 @@
...
 test-run:
 	@docker run -d --name acceptance-test${IMGNAME} -t \
       -e AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}" \
       -e AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}" \
-      -v ${ACCESS_KEY_LOCAL}:/go/src/github.com/k3s-io/k3s/tests/acceptance/modules/k3scluster/config/.ssh/aws_key.pem \
-      k3s-automated-${TAGNAME} sh -c 'cd ./tests/acceptance/entrypoint; \
+      -v ${ACCESS_KEY_LOCAL}:/go/src/github.com/rancher/distros-test-framework/config/.ssh/aws_key.pem \
+      acceptance-test-${TAGNAME} sh -c 'cd ./entrypoint; \
     if [ -n "${TESTDIR}" ]; then \
         if [ "${TESTDIR}" = "upgradecluster" ]; then \
```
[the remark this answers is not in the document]
mdrahman-suse: It will be done as part of the Makefile changes, I guess

Makefile
mdrahman-suse: Will leave that to be updated as part of the Makefile changes

entrypoint/mixedoscluster/mixedoscluster_suite_test.go
mdrahman-suse: I added that in the same script. Let me know if you think otherwise: `scripts/mixedos_sonobuoy.sh`

pkg/testcase/cluster.go
```diff
@@ -6,44 +6,100 @@ import (
...
+			}
 		}
 	}
 
-	err := shared.PrintFileContents(shared.KubeConfigFile)
+	fmt.Println("\nKubeconfig file:\n")
+	err = shared.PrintFileContents(shared.KubeConfigFile)
 	if err != nil {
 		return
 	}
```
fmoral2, line 47: can we break a line here? to better visualize the code scope

shared/aux.go
```diff
@@ -91,14 +92,26 @@ func PrintFileContents(f ...string) error {
...
+// PrintBase64Encoded prints the base64 encoded contents of the file as string.
+func PrintBase64Encoded(filepath string) error {
+    file, err := os.ReadFile(filepath)
+	if err != nil {
+		return fmt.Errorf("failed to encode file %s: %w", file, err)
+	}
+    encoded := base64.StdEncoding.EncodeToString(file)
+
+    fmt.Println(encoded)
+	return nil
```
fmoral2, line 112: add a space before return

entrypoint/mixedoscluster/mixedoscluster_suite_test.go
```diff
@@ -0,0 +1,59 @@
...
+		fmt.Println("Unsupported product to execute tests: " + cfg.Product)
+		os.Exit(1)
+	}
+	
+}
+
+func TestMixedOSClusterCreateSuite(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	if cfg.Product == "rke2" {
```
fmoral2, line 44: We dont need this validation here., since you added above, where it should be

shared/aux.go
```diff
@@ -237,3 +250,13 @@ func JoinCommands(cmd, kubeconfigFlag string) string {
...
 }
+
+// fileExists Checks if a file exists in a directory
+func fileExists(files []os.DirEntry, workload string) bool {
+	for _, file := range files {
+		if file.Name() == workload {
+			return true
+		}
+	}
+	return false
```
fmoral2, line 261: add a space before return here

Makefile
```diff
@@ -1,39 +1,44 @@
...
             go test -timeout=45m -v ./upgradecluster/... -installVersionOrCommit "${INSTALLTYPE}"; \
         elif [ "${TESTDIR}" = "versionbump" ]; then \
             go test -timeout=45m -v -tags=versionbump ./versionbump/... -cmd "${CMD}" -expectedValue "${VALUE}" \
```
fmoral2, line 20: can you just delete the comments ? and them in another task we fix the whole makefile

scripts/mixedos_sonobuoy.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+installation(){
+    echo "Installing sonobuoy version ${version} for mixedos validation"
+    git clone https://github.com/phillipsj/my-sonobuoy-plugins.git
+    wait
+    wget -q https://github.com/vmware-tanzu/sonobuoy/releases/download/v${version}/sonobuoy_${version}_linux_${arch}.tar.gz
+    wait
+    tar -xvf sonobuoy_${version}_linux_amd64.tar.gz
+    chmod +x sonobuoy && mv sonobuoy /usr/local/bin/sonobuoy
```
fmoral2, line 27: shouldnt we need the w permission as well ?

pkg/testcase/pod.go
```diff
@@ -44,8 +46,9 @@ func TestPodStatus(
 				}
 			}
 		}
-	}, "1000s", "5s").Should(Succeed())
+	}, "120s", "30s").Should(Succeed())
```
fmoral2, line 49: we should increate this 120s

scripts/mixedos_sonobuoy.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
```
fmoral2, line 2: i am getting this error when i try to run whitout delete the sonobuoy 
```
•failed to execute install action sonobuoy: exit status 1
Output: Installing sonobuoy version 0.56.16 for mixedos validation
fatal: destination path 'my-sonobuoy-plugins' already exists and is not an empty directory.
x LICENSE
x sonobuoy
```

shoulndt we need an automatic deletion  ?   or is just on my end ?

mdrahman-suse: ok let me check that, I did not run it locally so did not have that issue... likely need to add a check for the folder

scripts/mixedos_sonobuoy.sh
mdrahman-suse: I dont think we need to, I was able to run it with just x permission

pkg/testcase/node.go
```diff
@@ -33,8 +32,79 @@ func TestNodeStatus(
 				nodeAssertVersion(g, node)
 			}
 		}
-	}, "900s", "3s").Should(Succeed())
+	}, "900s", "150s").Should(Succeed())
```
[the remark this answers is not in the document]
mdrahman-suse: I went ahead and updated it to be variables based on the product as k3s takes less time than rke2... let me know wdyt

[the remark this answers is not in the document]
fmoral2: i dont think that its a real need for that, on my view its just adding more code without a real gain.

pkg/testcase/cluster.go
```diff
@@ -6,44 +6,102 @@ import (
...
 	"github.com/rancher/distros-test-framework/factory"
 	"github.com/rancher/distros-test-framework/shared"
+	"github.com/rancher/distros-test-framework/pkg/customflag"
+
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
+var arch = customflag.ServiceFlag.ClusterConfig.Arch.String()
```
fmoral2, line 16: this var here it would not work....bc this should be instatiate on the test main. did you tested it ?

shared/cluster.go
```diff
@@ -34,52 +35,63 @@ type Pod struct {
...
 	var err error
 
-	if action != "create" && action != "delete" {
-		return "", fmt.Errorf("invalid action: %s. Must be 'create' or 'delete'", action)
-	}
-
-	resourceDir := BasePath() + "/distros-test-framework/workloads/amd64/"
-	if arch == "arm64" {
-		resourceDir = BasePath() + "/distros-test-framework/workloads/arm/"
+	resourceDir := BasePath() + "/distros-test-framework/workloads/"
```
fmoral2, line 47: this path here should have the / amd ,  right ?

mdrahman-suse: I can just remove the if below and have default to arch

Makefile
mdrahman-suse: Updated

pkg/testcase/node.go
```diff
@@ -33,8 +32,70 @@ func TestNodeStatus(
...
 	_, err := shared.ParseNodes(true)
 	Expect(err).NotTo(HaveOccurred())
 }
+
+// TestInternodeConnectivityMixedOS Deploys services in the cluster and validates communication between linux and windows nodes
+func TestInternodeConnectivityMixedOS() {
+	_, err := shared.ManageWorkload("apply", "",
+		"pod_client.yaml","windows_app_deployment.yaml")
+	if err != nil {
+		fmt.Errorf("Error applying workload: ", err)
```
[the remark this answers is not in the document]
mdrahman-suse: Updated

scripts/mixedos_sonobuoy.sh
mdrahman-suse: Updated the script, please check and let me know if that worked

pkg/testcase/node.go
```diff
@@ -33,8 +32,70 @@ func TestNodeStatus(
...
+	}
+	
+	assert.ValidatePodIPByLabel([]string{"app=client","app=windows-app"},[]string{"10.42","10.42"})
+
+	err = testCrossNodeService(
+		[]string{"client-curl", "windows-app-svc"}, 
+		[]string{"8080", "3000"}, 
+		[]string{"Welcome to nginx", "Welcome to PSTools"})
+	if err != nil {
+		fmt.Errorf("Error checking cross node service: ", err)
```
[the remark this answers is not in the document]
mdrahman-suse: Updated

Makefile
fmoral2: i meant the comments bellow , after the file .

pkg/testcase/pod.go
```diff
@@ -44,8 +46,9 @@ func TestPodStatus(
 				}
 			}
 		}
-	}, "1000s", "5s").Should(Succeed())
+	}, "180s", "30s").Should(Succeed())
```
fmoral2, line 49: still wrong

factory/config.go
```diff
@@ -23,15 +23,19 @@ type Cluster struct {
...
 	RenderedTemplate string
 	ExternalDb       string
 	ClusterType      string
+	ProductType      string
+	ArchType		 string
 }
 
 func loadConfig() (*config.ProductConfig, error) {
-	cfg, err := config.LoadConfigEnv("./config")
+	cfg, err := config.LoadConfigEnv(shared.BasePath() + "/distros-test-framework/config")
```
fmoral2, line 38: why doing this ?

pkg/testcase/cluster.go
```diff
@@ -11,39 +11,105 @@ import (
...
 
-	Expect(shared.KubeConfigFile).ShouldNot(BeEmpty())
-	Expect(cluster.ServerIPs).ShouldNot(BeEmpty())
+	fmt.Println("Base64 Encoded Kubeconfig file:\n")
+	err = shared.PrintBase64Encoded(shared.KubeConfigFile)
+	if err != nil {
+		return
+	}
 
 	fmt.Println("Server Node IPS:", cluster.ServerIPs)
```
fmoral2, line 54: can you please add a  \n here   fmt.Println("Server Node IPS:",  ?  

locally gets odd being in sequence

pkg/testcase/node.go
```diff
@@ -16,10 +16,17 @@ func TestNodeStatus(
 	nodeAssertReadyStatus assert.NodeAssertFunc,
 	nodeAssertVersion assert.NodeAssertFunc,
 ) {
+	timeout := "300s"
+	interval := "30s"
 	cluster := factory.GetCluster(GinkgoT())
-	fmt.Println("\nFetching node status")
+	expectedNodeCount := cluster.NumServers + cluster.NumAgents 
+
+	if cluster.ProductType == "rke2" {
```
fmoral2, line 24: isnt better get in a between range ? i really think all those extra code only for this , kind of unnecessary

scripts/mixedos_sonobuoy.sh
fmoral2: checking

pkg/testcase/cluster.go
```diff
@@ -11,39 +11,105 @@ import (
...
-	fmt.Println("Agent Node IPS:", cluster.AgentIPs)
 
-	if cluster.NumAgents > 0 {
-		Expect(cluster.AgentIPs).ShouldNot(BeEmpty())
+	checkAndPrintAgentNodeIPs(cluster.NumAgents, cluster.AgentIPs, false)
+
```
fmoral2, line 57: what u think of always print  all ? like so 
```
Server Node IPS: [3.140.188.209 3.144.29.56]
Agent Node IPS: [3.137.163.226]
Windows Agent Node IPS: []
```

in this way we dont need to add this if and also dont need to add a function to it and repeat yourself calling the funciont 2 times

Makefile
mdrahman-suse: Not sure if I understand the ask, please advise

pkg/testcase/pod.go
mdrahman-suse: Why do you think it's wrong? What should be the right approach?

factory/config.go
mdrahman-suse: Updated thanks, I was trying out `.env` approach instead of config but it was not working...

pkg/testcase/cluster.go
mdrahman-suse: Sure, updated

pkg/testcase/node.go
mdrahman-suse: I need to understand the request here bit better, we can discuss offline

scripts/mixedos_sonobuoy.sh
mdrahman-suse: Updated again, added sudo now...

Makefile
fmoral2: it was not updated on my end! nw , sorry

pkg/testcase/service.go
```diff
@@ -112,4 +97,9 @@ func TestServiceLoadBalancer(deployWorkload bool) {
 			GinkgoT().Errorf("%v", err)
 		}
 	}
+
+	if delete {
+		_, err := shared.ManageWorkload("apply", arch, "loadbalancer.yaml")
```
ShylajaDevadiga, line 102: We need to pass "delete" here

pkg/testcase/service.go
```diff
@@ -39,19 +28,17 @@ func TestServiceClusterIp(deployWorkload bool) {
 			GinkgoT().Errorf("Error: %v", err)
 		}
 	}
```
ShylajaDevadiga, line 30: missing if delete check here

pkg/testcase/localpathstorage.go
```diff
@@ -47,7 +39,7 @@ func TestLocalPathProvisionerStorage(deployWorkload bool) {
 		g.Expect(err).NotTo(HaveOccurred())
 		g.Expect(res).Should(ContainSubstring("testing local path"))
 		g.Expect(err).NotTo(HaveOccurred())
-	}, "420s", "2s").Should(Succeed())
+	}, "420s", "30s").Should(Succeed())
```
ShylajaDevadiga, line 42: Would prefer shorter intervals to check, maybe 5s?

pkg/testcase/service.go
mdrahman-suse: sure updating

pkg/testcase/pod.go
fmoral2: a lot of times we have pods that took more than 180s to get ready. ...thats why the value was 1000. !

fmoral2: here locally if i dont increase this , always return error.

scripts/mixedos_sonobuoy.sh
```diff
@@ -0,0 +1,55 @@
...
+# Usage: sh mixedos_sonobouy.sh install 0.56.16 amd64
+# Usage: sh mixedos_sonobouy.sh delete
+
+if [ -z "$2" ]
```
ShylajaDevadiga, line 6: if version is not passed arch would be $2

scripts/mixedos_sonobuoy.sh
```diff
@@ -0,0 +1,55 @@
...
+    version="0.56.16"
+else
+    version=$2
+fi
+
+if [ -z "$3" ]
+then
+    arch="amd64"
+else
+    arch=$3
```
ShylajaDevadiga, line 17: check for keyword arm64

entrypoint/mixedoscluster/mixedoscluster_test.go
```diff
@@ -0,0 +1,48 @@
...
+	It("Validate Pods", func() {
+		testcase.TestPodStatus(
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus(),
+		)
+	})
+
+	It("Validates internode connectivity over the vxlan tunnel", func() {
+		testcase.TestInternodeConnectivityMixedOS(true)
```
ShylajaDevadiga, line 34: Are we going to reuse this method?

pkg/testcase/node.go
```diff
@@ -33,8 +40,82 @@ func TestNodeStatus(
...
-	}, "900s", "3s").Should(Succeed())
+	}, timeout, interval).Should(Succeed())
 
+	fmt.Println("\n\nCluster nodes:")
 	_, err := shared.ParseNodes(true)
 	Expect(err).NotTo(HaveOccurred())
 }
+
+// TestInternodeConnectivityMixedOS Deploys services in the cluster and validates communication between linux and windows nodes
+func TestInternodeConnectivityMixedOS(delete bool) {
```
ShylajaDevadiga, line 51: Can we create networkconnectivity.go and these there?

factory/config.go
```diff
@@ -23,11 +23,15 @@ type Cluster struct {
 	Status           string
 	ServerIPs        []string
 	AgentIPs         []string
+	WinAgentIPs		 []string
```
ShylajaDevadiga, line 26: alignment

pkg/testcase/ingressdns.go
```diff
@@ -44,31 +38,34 @@ func TestIngress(deployWorkload bool) {
...
-	if deployWorkload {
-		_, err := shared.ManageWorkload(
-			"create",
-			"dnsutils.yaml",
-			customflag.ServiceFlag.ClusterConfig.Arch.String(),
-		)
-		Expect(err).NotTo(HaveOccurred(),
-			"dnsutils manifest not deployed")
+	if delete {
+		_, err := shared.ManageWorkload("apply", arch, "ingress.yaml")
```
fmoral2, line 43: here should it be   "delete"  instead of apply .

pkg/testcase/cluster.go
```diff
@@ -11,39 +11,105 @@ import (
...
+		checkAndPrintAgentNodeIPs(cluster.NumWinAgents, cluster.WinAgentIPs, true)
+	}	
+}
+
+// TestSonobuoyMixedOS runs sonobuoy tests for mixed os cluster (linux + windows) node
+func TestSonobuoyMixedOS(version string, delete bool) {
+	err := shared.SonobuoyMixedOS("install", version)
+	if err != nil {
+		fmt.Println(err)
+		return
```
fmoral2, line 68: on those returns on tests , since all of our test functions do not return errors , in return statements that we need to break the flow because of an error like here, you can do something like :

```

if err != nil {
GinkgoT().Errorf("error {your custom message if needed }: %v", err)
}
```

pkg/testcase/node.go
```diff
@@ -33,8 +40,82 @@ func TestNodeStatus(
...
+	_, err := shared.ManageWorkload("apply", arch,
+		"pod_client.yaml","windows_app_deployment.yaml")
+	if err != nil {
+		fmt.Println(err)
```
fmoral2, line 55: https://github.com/rancher/distros-test-framework/pull/6/files#r1293488769

pkg/testcase/node.go
```diff
@@ -33,8 +40,82 @@ func TestNodeStatus(
...
+	
+	assert.ValidatePodIPByLabel([]string{"app=client","app=windows-app"},[]string{"10.42","10.42"})
+
+	err = testCrossNodeService(
+		[]string{"client-curl", "windows-app-svc"}, 
+		[]string{"8080", "3000"}, 
+		[]string{"Welcome to nginx", "Welcome to PSTools"})
+	if err != nil {
+		fmt.Println("Error checking cross node service: ", err)
+
```
fmoral2, line 68: https://github.com/rancher/distros-test-framework/pull/6/files#r1293488769

pkg/testcase/node.go
```diff
@@ -33,8 +40,82 @@ func TestNodeStatus(
...
+		return
+	}
+
+	if delete {
+		_, err := shared.ManageWorkload("delete", arch,
+			"pod_client.yaml","windows_app_deployment.yaml")
+		if err != nil {
+			fmt.Println(err)
+
```
fmoral2, line 77: https://github.com/rancher/distros-test-framework/pull/6/files#r1293488769

entrypoint/mixedoscluster/mixedoscluster_test.go
mdrahman-suse: I am not sure if this can be re-used in any test other that mixed OS test.. may be when we add upgrade tests for Windows, this can be re-used

fmoral2:
if you implemented the ARCH being loaded through vars u must update the entire project as: 

- suites files , to remove the arch flags
- model - > /Users/moral/suse/distros-test-framework/pkg/customflag/model.go

pkg/testcase/cluster.go
mdrahman-suse: Updated

pkg/testcase/node.go
mdrahman-suse: updated

scripts/mixedos_sonobuoy.sh
mdrahman-suse: as discussed, will address this later

pkg/testcase/networkconnectivity.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+// testCrossNodeService Perform testing cross node communication via service exec call
+//
+// services Slice Takes service names as parameters in the array
+//
+// ports	Slice Takes service ports needed to access the services
+//
+// expected	array Takes the expected substring from the curl response
```
fmoral2, line 50: slice instead of array here

pkg/testcase/localpathstorage.go
```diff
@@ -47,7 +39,7 @@ func TestLocalPathProvisionerStorage(deployWorkload bool) {
 		g.Expect(err).NotTo(HaveOccurred())
 		g.Expect(res).Should(ContainSubstring("testing local path"))
 		g.Expect(err).NotTo(HaveOccurred())
-	}, "420s", "2s").Should(Succeed())
+	}, "300s", "10s").Should(Succeed())
```
fmoral2, line 42: i found on my testing of this , that 10s is to much .

pkg/testcase/networkconnectivity.go
```diff
@@ -0,0 +1,82 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// TestInternodeConnectivityMixedOS Deploys services in the cluster and validates communication between linux and windows nodes
+func TestInternodeConnectivityMixedOS(deleteWorkload bool) {
+	_, err := shared.ManageWorkload("apply",
+		"pod_client.yaml","windows_app_deployment.yaml")
+	if err != nil {
+		GinkgoT().Errorf("error: %v", err)
```
fmoral2, line 18: this func already return the test as failed. 
is similar to `	Expect(err).NotTo(HaveOccurred(), "xxxxxx  not deployed")`

so no need to return as well here

pkg/testcase/networkconnectivity.go
```diff
@@ -0,0 +1,82 @@
...
+		return
+	}
+
+	if deleteWorkload {
+		_, err := shared.ManageWorkload("delete",
+			"pod_client.yaml","windows_app_deployment.yaml")
+		if err != nil {
+			GinkgoT().Errorf("error: %v", err)
+			return
```
fmoral2, line 39: https://github.com/rancher/distros-test-framework/pull/6/files#r1294514719

pkg/testcase/coredns.go
```diff
@@ -10,14 +10,11 @@ import (
...
-			"dnsutils manifest not deployed", err)
-	}
+func TestCoredns(deleteWorkload bool) {
+	_, err := shared.ManageWorkload("apply", "dnsutils.yaml")
+	Expect(err).NotTo(HaveOccurred(),"dnsutils manifest not deployed")
 
-	_, err := shared.AddHelmRepo("traefik", "https://helm.traefik.io/traefik")
+	_, err = shared.AddHelmRepo("traefik", "https://helm.traefik.io/traefik")
 	if err != nil {
 		log.Fatalf("failed to add Helm repo: %v", err)
```
fmoral2, line 19: https://github.com/rancher/distros-test-framework/pull/6/files#r1294514719

pkg/testcase/coredns.go
```diff
@@ -37,4 +34,9 @@ func TestCoredns(deployWorkload bool) {
 	if err != nil {
 		return
```
fmoral2, line 35: https://github.com/rancher/distros-test-framework/pull/6/files#r1294514719

pkg/testcase/localpathstorage.go
```diff
@@ -14,22 +13,15 @@ import (
...
+	_, err := shared.ManageWorkload("apply", "local-path-provisioner.yaml")
+	Expect(err).NotTo(HaveOccurred(), "local-path-provisioner manifest not deployed")
 
 	getPodVolumeTestRunning := "kubectl get pods -n local-path-storage" +
 		" --field-selector=status.phase=Running --kubeconfig=" + shared.KubeConfigFile
-	err := assert.ValidateOnHost(
+	err = assert.ValidateOnHost(
 		getPodVolumeTestRunning,
-		Running,
+		statusRunning,
```
fmoral2, line 24: on line 32 as well. please 
https://github.com/rancher/distros-test-framework/pull/6/files#r1294514719

pkg/testcase/networkconnectivity.go
mdrahman-suse: Should generalize the approach then, having different approach is kind of confusing... so should I go with this or the Expect? Ideally we can go with Expect being this a test itself, wdyt?

fmoral2 (approved):

pkg/testcase/coredns.go
```diff
@@ -1,40 +1,33 @@
...
-		log.Fatalf("failed to add Helm repo: %v", err)
-	}
+	_, err = shared.AddHelmRepo("traefik", "https://helm.traefik.io/traefik")
+	Expect(err).NotTo(HaveOccurred(), "failed to add Helm repo: %v", err)
 
 	kubeconfigFlag := " --kubeconfig=" + shared.KubeConfigFile
 	fullCmd := shared.JoinCommands("helm list --all-namespaces ", kubeconfigFlag)
 	err = assert.CheckComponentCmdHost(
 		fullCmd,
 		"rke2-coredns-1.19.402",
```
mdrahman-suse, line 21: Should this be hardcoded? @fmoral2

fmoral2: definitely not lol. i can correct on my PR, seen that i have no idea what should be going here   :)

pkg/testcase/service.go
```diff
@@ -1,101 +1,73 @@
...
-		_, err := shared.ManageWorkload(
-			"create",
-			"clusterip.yaml",
-			customflag.ServiceFlag.ClusterConfig.Arch.String(),
-		)
-		Expect(err).NotTo(HaveOccurred(),
-			"Cluster IP manifest not deployed")
-	}
+func TestServiceClusterIP(deleteWorkload bool) {		
+	_, err := shared.ManageWorkload("apply", "clusterip.yaml")
```
rancher-max, line 11: Apply could overwrite -- is this definitely what we want for ALL test cases that use this function? 

Also, do we ALWAYS want to apply with this? Are there any cases that call this function that are expecting the resource to not be there and attempt a delete?


(same questions for all similar changes in this file)

pkg/testcase/networkconnectivity.go
```diff
@@ -0,0 +1,71 @@
...
+	. "github.com/onsi/gomega"
+)
+
+// TestInternodeConnectivityMixedOS Deploys services in the cluster and validates communication between linux and windows nodes
+func TestInternodeConnectivityMixedOS(deleteWorkload bool) {
+	_, err := shared.ManageWorkload("apply",
+		"pod_client.yaml","windows_app_deployment.yaml")
+	Expect(err).NotTo(HaveOccurred(), err)
+	
+	assert.ValidatePodIPByLabel([]string{"app=client","app=windows-app"},[]string{"10.42","10.42"})
```
rancher-max, line 18: What if we're passing in a specific clusterCIDR or serviceCIDR where it's not in the `10.42` range but rather the `10.43` or something? Would this test then fail?

If so, is there a way we could get this value dynamically? Or validate some other part of it?

scripts/mixedos_sonobuoy.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# Usage: sh mixedos_sonobouy.sh install 0.56.17 amd64
+# Usage: sh mixedos_sonobouy.sh delete
+
+action=$1
+
```
rancher-max, line 7: Can we have variables for $2 and $3 as `version` and `arch` so that this shell file is more clear too and easier to make updates to?

rancher-max (changes requested):
Some general questions and one area for changes

pkg/testcase/service.go
mdrahman-suse: As long as the deployment file is unchanged, it should be fine, right? This is one of the reason I did not want to have one method to handle the operations as it makes the Test more abstract

We can definitely add a check for the resource to exist before performing deletion but atm it does not exist

pkg/testcase/networkconnectivity.go
mdrahman-suse: Will check for that option for improvements but atm this is how it can be validated

scripts/mixedos_sonobuoy.sh
mdrahman-suse: Updated

factory/cluster.go
```diff
@@ -79,18 +88,15 @@ func DestroyCluster(g GinkgoTInterface) (string, error) {
 		return "", err
 	}
 
-	if cfg.Product == "rke2" {
-		varDir, err = filepath.Abs(shared.BasePath() + "/distros-test-framework/config/rke2.tfvars")
-	} else if cfg.Product == "k3s" {
-		varDir, err = filepath.Abs(shared.BasePath() + "/distros-test-framework/config/rke2.tfvars")
+	if cfg.Product == "rke2" || cfg.Product == "k3s"{
```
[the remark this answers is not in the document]
mdrahman-suse: Good thought... Simply running it would work, doing this in addtion to handle a case where it wont be either or those product, which is less likely to happen

pkg/testcase/service.go
rancher-max: Yeah abstract tests aren't great and there is definitely room for improvement overall in that regard. You're right in that if the file doesn't change it's fine, but I'm wondering more for the case of "Let's try to delete something that we never applied in the first place" -- in that case we'd have to use a different method altogether I guess, which is fine!

I think it's okay as is 👍

entrypoint/mixedoscluster/mixedoscluster_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/rancher/distros-test-framework/pkg/assert"
+	"github.com/rancher/distros-test-framework/pkg/testcase"
+
+	. "github.com/onsi/ginkgo/v2"
+)
+
+var _ = Describe("Test: Mixed OS Cluster", func() {
+
+	It("Starts Up with no issues", func() {
+		testcase.TestBuildCluster(GinkgoT())
```
ShylajaDevadiga, line 15: Can we avoid creating the cluster if the config contains hardened parameters?

factory/cluster.go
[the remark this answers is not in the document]
mdrahman-suse: Updated thanks @aganesh-suse

entrypoint/mixedoscluster/mixedoscluster_test.go
mdrahman-suse: In that case we have to be able to read the server configs. We don't have that ability now but can add it in later iterations

rancher-max (approved):
LGTM!

entrypoint/mixedoscluster/mixedoscluster_test.go
mdrahman-suse: Updated Jenkinsfile to handle this condition, please review

scripts/Jenkinsfile
```diff
@@ -80,6 +80,10 @@ node {
                   def testdata = readFile(file: filename)
                   println("\n\nTFVARS TEST DATA")
                   println(testdata)
+                  if ((env.PRODUCT_NAME == 'k3s' && TESTCASE == 'mixedoscluster') || (env.PRODUCT_NAME == 'rke2' && testdata.contains('cis'))) {
+                    error('Unsupported test execution found, Aborting the build!')
```
aganesh-suse, line 84: We can explicitly specify which test combination is not allowed in the error message for better comprehension of why this errored.

ShylajaDevadiga (approved):
LGTM

scripts/Jenkinsfile
mdrahman-suse: Updated, please review

aganesh-suse (dismissed):

VestigeJ (approved):
lgtm

aganesh-suse (approved):

ShylajaDevadiga (approved):

mdrahman-suse:
Thanks y'll 👍

PR #12, opened 2023-08-15
chore: add lint on gh actions

fmoral2:
#### Proposed Changes ####

- add go lint on pushes 

#### Types of Changes ####

- Addition of golang ci lint on pushes 

#### Verification ####

- Pushing new code

#### Testing ####
- n/a

### DIFF
--- a/.github/workflows/run-distros.yaml
+++ b/.github/workflows/run-distros.yaml
@@ -1,0 +1,39 @@
+name: Run Distros Framework
+
+on:
+  push:
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build and Lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+      - name: Checkout distros
+        uses: actions/checkout@v3
+
+      - name: Cache Go Modules
+        uses: actions/cache@v3
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - name: Build Docker image
+        run: docker build . -q -f scripts/Dockerfile.build -t acceptance-tests
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.20'
+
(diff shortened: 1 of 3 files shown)


.github/workflows/run-distros.yaml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  push:
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build and Lint
+    runs-on: ubuntu-20.04
```
VestigeJ, line 12: Do we want to keep these ubuntu-latest for the freemium run time?

fmoral2: Any objections?

fmoral2: I just got the default one from actions, I can add other

scripts/Dockerfile.build
```diff
@@ -22,6 +22,8 @@ RUN apk update && \
     chmod +x /usr/local/bin/terraform && \
     rm -rf tmp
 
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin v1.53.3
```
osodracnai, line 25:
```suggestion
```
actions will already run golangci-liint, install on docker is not needed

fmoral2: added there to have available running here locally with docker

.github/workflows/run-distros.yaml
VestigeJ: doing ubuntu-latest keeps it a free run if we’re under our monthly allowance from github for the org if i recall correctly.

fmoral2: sure

mdrahman-suse:
I don't see any step for cleanup, is that not needed or will be addressed later?

VestigeJ (approved):
LGTM

scripts/Dockerfile.build
mdrahman-suse: In that case does it makes sense to have separate dockerfile for local and CI?

mdrahman-suse (approved):
LGTM

PR #13, opened 2023-08-22
[Not Ready] Add dual stack tests

mdrahman-suse:
#### Proposed Changes ####

- This PR adds the ability to run dual stack tests (Cluster with both IPV4 and IPV6 IPs) which includes validation with ingress, nodeport, clusterIP and multi-stack deployments.
- The test is configured to run with a Bastion instance created on the same subnet as the server and agent nodes as traditional way of test execution were causing failures when trying to validate IPV6 IPs
- The test needs to be executed on AWS us-west-1 region as the security group, vpc and subnet ids specific to dual stack setting is already configured in that region
- Additional variables are added keeping in mind for airgap and ipv6 testing, which will be added in upcoming PRs

- Yes, the changes require updating documentation

#### Types of Changes ####

- Adding dual stack tests
- Updating existing vars: 
  - In k3s.tfvars: `instance_class` -> `db_instance_class` and `environment` -> `db_environment`
- Additional Variables in *.tfvars:
```
no_of_bastion_nodes = 1
bastion_subnets     = "<subnet-of-bastion-instance>"
enable_public_ip   = true
enable_ipv6        = true
```
- Required Variable: 
```
server_flags = "cluster-cidr: <IPV4-CIDR-1>,<IPV6-CIDR-1>\nservice-cidr: <IPV4-CIDR-2>,<IPV6-CIDR-2>\n"
```

#### Testing ####

Checklist:
- Ran the PR on both k3s and rke2 Jenkins job and shared the results

- Locally having some issue so depending in the CI linting

- TBA

- Updated makefile

- The code continues to work with existing settings

- Only ran with Ubuntu in us-west-1, if needed will execute on other OS types

#### Linked Issues ####

Ref Ticket: https://github.com/rancher/qa-tasks/issues/882

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,11 @@ include ./config/.env
 TAG_NAME := $(if $(TAG_NAME),$(TAG_NAME),distros)
 
 
+#========================= Run tests with Docker locally =========================#
+
+.PHONY: test-env-up
 test-env-up:
-	@docker build . -q -f ./scripts/Dockerfile.build -t acceptance-test-${TAG_NAME}
+	@docker build . -q -f ./scripts/Dockerfile.build -t acceptance-test-${TAGNAME}
 
 test-run:
 	@docker run -dt --name acceptance-test-${IMG_NAME} \
--- a/Makefile
+++ b/Makefile
@@ -7,21 +7,21 @@ TAG_NAME := $(if $(TAG_NAME),$(TAG_NAME),distros)
 
 .PHONY: test-env-up
 test-env-up:
-	@docker build . -q -f ./scripts/Dockerfile.build -t acceptance-test-${TAGNAME}
+	@docker build . -q -f ./scripts/Dockerfile.build -t acceptance-test-${TAG_NAME}
 
 test-run:
 	@docker run -dt --name acceptance-test-${IMG_NAME} \
 	  -e AWS_ACCESS_KEY_ID=$${AWS_ACCESS_KEY_ID} \
 	  -e AWS_SECRET_ACCESS_KEY=$${AWS_SECRET_ACCESS_KEY} \
+	  -e TEST_DIR=${TEST_DIR} \
+	  -e IMG_NAME=${IMG_NAME} \
+	  -e TEST_TAG=${TEST_TAG} \
 	  --env-file ./config/.env \
 	  -v ${ACCESS_KEY_LOCAL}:/go/src/github.com/rancher/distros-test-framework/config/.ssh/aws_key.pem \
-	  -v ./scripts/test-runner.sh:/go/src/github.com/rancher/distros-test-framework/scripts/test-runner.sh \
-	  acceptance-test-${TAG_NAME} && \
-	  make image-stats IMG_NAME=${IMG_NAME} && \
-	  make test-logs USE=IMG_NAME acceptance-test-${IMG_NAME}
+	  acceptance-test-${TAG_NAME} 
 
 
-## Use this to run on the same environement + cluster from the previous last container -${TAGNAME} created
+## Use this to run on the same environement + cluster from the previous last container -${TAG_NAME} created
 test-run-state:
 	DOCKER_COMMIT=$$? \
 	CONTAINER_ID=$(shell docker ps -a -q --filter ancestor=acceptance-test-${TAG_NAME} | head -n 1); \
--- a/entrypoint/dualstack/dualstackcluster_test.go
+++ b/entrypoint/dualstack/dualstackcluster_test.go
@@ -1,0 +1,55 @@
+package dualstack
+
+import (
+	"fmt"
+
+	"github.com/rancher/distros-test-framework/pkg/assert"
+	"github.com/rancher/distros-test-framework/pkg/testcase"
+
+	. "github.com/onsi/ginkgo/v2"
+)
+
+var _ = Describe("Test:", func() {
+	It("Start Up with no issues", func() {
+		testcase.TestBuildCluster(GinkgoT())
+	})
+
+	It("Validate Nodes", func() {
+		testcase.TestNodeStatus(
+			assert.NodeAssertReadyStatus(),
+			nil,
+		)
+	})
+
+	It("Validate Pods", func() {
+		testcase.TestPodStatus(
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus(),
+		)
+	})
+
+	It("Validate Dual-Stack Ingress Service", func() {
+		testcase.TestDualStackIngress(true)
+	})
+
+	It("Validate Dual-Stack NodePort Service", func() {
+		testcase.TestDualStackNodePort(true)
+	})
+
+	It("Validate Dual-Stack ClusterIPs in CIDR range", func() {
+		testcase.TestDualStackClusterIPsInCIDRRange(true)
+	})
+
+	It("Validate Single and Dual-Stack IPFamilies", func() {
+		testcase.TestDualStackIPFamilies(true)
+	})
+})
+
+var _ = AfterEach(func() {
+	if CurrentSpecReport().Failed() {
+		fmt.Printf("\nFAILED! %s\n", CurrentSpecReport().FullText())
+	} else {
+		fmt.Printf("\nPASSED! %s\n", CurrentSpecReport().FullText())
+	}
+})
--- a/entrypoint/dualstack/dualstackcluster_test.go
+++ b/entrypoint/dualstack/dualstackcluster_test.go
@@ -29,20 +29,20 @@ var _ = Describe("Test:", func() {
 		)
 	})
 
-	It("Validate Dual-Stack Ingress Service", func() {
-		testcase.TestDualStackIngress(true)
+	It("Validate Ingress Service in Dual-Stack", func() {
+		testcase.TestDualStackIngress(false)
 	})
 
-	It("Validate Dual-Stack NodePort Service", func() {
-		testcase.TestDualStackNodePort(true)
+	It("Validate NodePort Service in Dual-Stack", func() {
+		testcase.TestDualStackNodePort(false)
 	})
 
-	It("Validate Dual-Stack ClusterIPs in CIDR range", func() {
+	It("Validate ClusterIPs in CIDR range in Dual-Stack", func() {
 		testcase.TestDualStackClusterIPsInCIDRRange(true)
 	})
 
-	It("Validate Single and Dual-Stack IPFamilies", func() {
-		testcase.TestDualStackIPFamilies(true)
+	It("Validate Single and Dual-Stack IPFamilies in Dual-Stack", func() {
+		testcase.TestDualStackIPFamilies(false)
 	})
 })
 
--- a/factory/cluster.go
+++ b/factory/cluster.go
@@ -94,7 +94,7 @@ func DestroyCluster(g GinkgoTInterface) (string, error) {
 	tfDir, err := filepath.Abs(shared.BasePath() +
 		fmt.Sprintf("/distros-test-framework/modules/%s", cfg.Product))
 	if err != nil {
-		return "", shared.ReturnLogError("no module found for product: %s\n", cfg.Product)
+		return "", shared.ReturnLogError("invalid product: %s\n", cfg.Product)
 	}
 
 	terraformOptions := terraform.Options{
--- a/modules/install/join_k3s_agent.sh
+++ b/modules/install/join_k3s_agent.sh
@@ -4,26 +4,55 @@
 #set -x
 # PS4='+(${LINENO}): '
 
+set -x
+echo "$@"
+
+node_os=${1}
+server_ip=${2}
+token=${3}
+public_ip=${4}
+private_ip=${5}
+ipv6_ip=${6}
+install_mode=${7}
+version=${8}
+channel=${9}
+worker_flags=${10}
+rhel_username=${12}
+rhel_password=${13}
+
 mkdir -p /etc/rancher/k3s
 
 create_config() {
-  local server_ip="${1}"
-  local token="${2}"
-
+  hostname=$(hostname -f)
   cat <<EOF >>/etc/rancher/k3s/config.yaml
-server: https://${1}:6443
-token:  "${2}"
+server: "https://$server_ip:6443"
+token:  "$token"
+node-name: $hostname
 EOF
 }
 
 add_config() {
-  local worker_flags="${1}"
-
   if [[ -n "$worker_flags" ]] && [[ "$worker_flags" == *":"* ]]
-    then
-      echo -e "$worker_flags" >> /etc/rancher/k3s/config.yaml
-      cat /etc/rancher/k3s/config.yaml
+  then
+    echo -e "$worker_flags" >> /etc/rancher/k3s/config.yaml
   fi
+
+  if [[ "$worker_flags" != *"cloud-provider-name"* ]]
+  then
+    if [ -n "$ipv6_ip" ] && [ -n "$public_ip" ] && [ -n "$private_ip" ]
+    then
+      echo -e "node-external-ip: $public_ip,$ipv6_ip" >> /etc/rancher/k3s/config.yaml
+      echo -e "node-ip: $private_ip,$ipv6_ip" >> /etc/rancher/k3s/config.yaml
+    elif [ -n "$ipv6_ip" ]
+    then
+      echo -e "node-external-ip: $ipv6_ip" >> /etc/rancher/k3s/config.yaml
+      echo -e "node-ip: $ipv6_ip" >> /etc/rancher/k3s/config.yaml
+    else
+      echo -e "node-external-ip: $public_ip" >> /etc/rancher/k3s/config.yaml
+      echo -e "node-ip: $private_ip" >> /etc/rancher/k3s/config.yaml
+    fi
+  fi
+  cat /etc/rancher/k3s/config.yaml
 
   if [[ -n "$worker_flags" ]] && [[ "$worker_flags" == *"protect-kernel-defaults"* ]]
     then
@@ -35,65 +64,47 @@ add_config() {
 }
 
 rhel() {
-   local node_os="${1}"
-   local username="${2}"
-   local password="${3}"
-
   if [ "$node_os" = "rhel" ]
-    then
-      subscription-manager register --auto-attach --username="$username" --password="$password"
-      subscription-manager repos --enable=rhel-7-server-extras-rpms
+  then
+    subscription-manager register --auto-attach --username="$rhel_username" --password="$rhel_password"
+    subscription-manager repos --enable=rhel-7-server-extras-rpms
   fi
 }
 
 disable_cloud_setup() {
-   local node_os="${1}"
-
-if  [[ "$node_os" = *"rhel"* ]] || [[ "$node_os" = *"centos"* ]]
+  if  [[ "$node_os" = *"rhel"* ]] || [[ "$node_os" = *"centos"* ]]
   then
     NM_CLOUD_SETUP_SERVICE_ENABLED=$(systemctl status nm-cloud-setup.service | grep -i enabled)
     NM_CLOUD_SETUP_TIMER_ENABLED=$(systemctl status nm-cloud-setup.timer | grep -i enabled)
 
     if [ "${NM_CLOUD_SETUP_SERVICE_ENABLED}" ]; then
-    systemctl disable nm-cloud-setup.service
+      systemctl disable nm-cloud-setup.service
     fi
 
     if [ "${NM_CLOUD_SETUP_TIMER_ENABLED}" ]; then
-    systemctl disable nm-cloud-setup.timer
+      systemctl disable nm-cloud-setup.timer
     fi
-fi
+  fi
 }
 
-export "${2}"="${3}"
+export "$install_mode"="$version"
 
 install(){
-  local version="${1}"
-  local worker_flags="${2}"
-  local install_mode="${3}"
-  local ip="${4}"
-  local server_ip="${5}"
-  local token="${6}"
-  local channel="${7}"
-
-if [[ "$version" == *"v1.18"* ]] || [[ "$version" == *"v1.17"* ]] && [[ -n "$worker_flags" ]]
+  if [[ -n "$channel"  ]]
   then
-    curl -sfL https://get.k3s.io | sh -s - "$install_mode" --node-external-ip="$ip" --server https://"$server_ip":6443 --token "$token"
-else
-    if [[ -n "$channel"  ]]
-    then
-      curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=$channel sh -s - agent --node-external-ip="$ip"
-    else
-      curl -sfL https://get.k3s.io | sh -s - agent --node-external-ip="$ip"
-    fi
-    sleep 10
-fi
+    curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=$channel sh -s - agent
+  else
+    curl -sfL https://get.k3s.io | sh -s - agent
+  fi
+  sleep 10
+  
 }
 
 main() {
-  create_config "$4" "$5"
-  add_config "$7"
-  rhel "$1" "$8" "$9"
-  disable_cloud_setup "$1"
-  install "$3" "$7" "$2" "$6" "$4" "$5" "${10}"
+  create_config
+  add_config
+  rhel
+  disable_cloud_setup
+  install
 }
 main "$@"
--- a/modules/install/join_rke2_agent.sh
+++ b/modules/install/join_rke2_agent.sh
@@ -85,7 +85,7 @@ fi
 
 if [ "$channel" != "null" ]
 then
-    curl -sfL https://get.rke2.io | INSTALL_channel="$channel" INSTALL_RKE2_TYPE='agent' sh -
+    curl -sfL https://get.rke2.io | INSTALL_RKE2_CHANNEL="$channel" INSTALL_RKE2_TYPE='agent' sh -
 else
     curl -sfL https://get.rke2.io | INSTALL_RKE2_TYPE='agent' sh -
 fi
--- a/modules/install/join_rke2_agent.sh
+++ b/modules/install/join_rke2_agent.sh
@@ -1,23 +1,25 @@
 #!/bin/bash
 # This script is used to join one or more nodes as agents
-#set -x
+set -x
 echo "$@"
 
-node_os=$1
-server_ip=$2
-token=$3
-rke2_version=$4
-public_ip=$5
-rke2_channel=$6
-worker_flags=$7
-install_mode=$8
-rhel_username=$9
-rhel_password=${10}
-install_method=${11}
+node_os=${1}
+server_ip=${2}
+token=${3}
+public_ip=${4}
+private_ip=${5}
+ipv6_ip=${6}
+version=${7}
+channel=${8}
+install_mode=${9}
+install_method=${10}
+worker_flags=${11}
+rhel_username=${12}
+rhel_password=${13}
 
 hostname=$(hostname -f)
 mkdir -p /etc/rancher/rke2
-cat <<EOF >>/etc/rancher/rke2/config.yaml
+cat << EOF >/etc/rancher/rke2/config.yaml
 server: https://${server_ip}:9345
 token:  "${token}"
 node-name: "${hostname}"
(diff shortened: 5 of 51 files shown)


shared/cluster.go
```diff
@@ -200,6 +204,23 @@ func FetchClusterIP(namespace string, serviceName string) (string, string, error
 	return ip, port, err
 }
 
+// FetchClusterIPs returns the cluster IPs and port of the service.
+func FetchClusterIPs(namespace string, svc string) (string, string, error) {
```
fmoral2, line 208: we already had this func, why adding another ?

config/reader.go
```diff
@@ -49,7 +49,7 @@ func loadConfigEnv(path string) (config *ProductConfig, err error) {
 		os.Exit(1)
 	}
 
-	if config.TFVars == "" || (config.TFVars != "k3s.tfvars" && config.TFVars != "rke2.tfvars") {
+	if config.TFVars == "" || (!strings.Contains(config.TFVars, "tfvars")) {
```
fmoral2, line 52: we should keep the product validation

entrypoint/dualstack/dualstackcluster_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+		_, err = shared.ManageWorkload("delete", "dualstack-ingress.yaml")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("Validate Dual-Stack NodePort Service", func() {
+		_, err := shared.ManageWorkload("apply", "dualstack-nodeport.yaml")
+		Expect(err).NotTo(HaveOccurred())
+
+		testdata := map[string]string{
```
fmoral2, line 56: here it should be only the calling of the tests no data or logic.

https://github.com/rancher/distros-test-framework/blob/2ce312df8081eb1d6a8d2043cf9013afe75daa76/README.md?plain=1#L46

entrypoint/dualstack/dualstackcluster_test.go
```diff
@@ -0,0 +1,126 @@
...
+			"namespace":"default",
+			"label":"app=dualstack-nodeport",
```
fmoral2, line 58: all testcases functionalities must comply the same signature 
https://github.com/rancher/distros-test-framework/blob/2ce312df8081eb1d6a8d2043cf9013afe75daa76/pkg/customflag/model.go#L50C5-L50C5

wich is only deploy workload as param.

entrypoint/dualstack/dualstackcluster_test.go
```diff
@@ -0,0 +1,126 @@
...
+		_,err := shared.ManageWorkload("apply","dualstack-clusterip.yaml")
+		Expect(err).NotTo(HaveOccurred())
+
+		testdata := map[string]string{
+			"namespace":"default",
+			"label":"app=clusterip-demo",
+			"svc":"clusterip-svc-demo",
+		}
+
+		testcase.TestPodStatusIsRunningByNamespaceAndLabel(testdata["namespace"], testdata["label"])
```
fmoral2, line 81: all testcases functionalities must comply the same signature 
https://github.com/rancher/distros-test-framework/blob/2ce312df8081eb1d6a8d2043cf9013afe75daa76/pkg/customflag/model.go#L50C5-L50C5

wich is only deploy workload as param.

factory/cluster.go
```diff
@@ -71,14 +71,14 @@ func DestroyCluster(g GinkgoTInterface) (string, error) {
 		return "", shared.ReturnLogError("error getting config: %w", err)
 	}
 
-	varDir, err = filepath.Abs(shared.BasePath() + fmt.Sprintf("/distros-test-framework/config/%s.tfvars", cfg.Product))
+	tfDir, err := filepath.Abs(shared.BasePath() + fmt.Sprintf("/distros-test-framework/modules/%s", cfg.Product))
```
fmoral2, line 74: why getting from modules? we should maintain the pattern and best practice to load config from config. not from anywhere else.

Makefile
```diff
@@ -2,13 +2,15 @@ include ./config.mk
 
 TAGNAME ?= default
 
+#========================= Run tests with Docker locally =========================#
```
fmoral2, line 5: if you want u can push no changes to make file, since on the next PR it will overwrite almost all .

factory/cluster.go
```diff
@@ -71,14 +71,14 @@ func DestroyCluster(g GinkgoTInterface) (string, error) {
...
 	if err != nil {
-		return "", shared.ReturnLogError("invalid product: %s\n", cfg.Product)
+		return "", err
```
fmoral2, line 76: why taking out the logs ?

pkg/assert/pod.go
```diff
@@ -98,6 +99,52 @@ func ValidatePodIPByLabel(labels, expected []string) {
 			}
 		}
 		return nil
-	}, "180s", "30s").Should(Succeed(),
+	}, "180s", "5s").Should(Succeed(), 
 		"failed to validate expected: %s on %s", expected, labels)
 }
+
+// ValidatePodIPsByLabel validates expected pod IPs by label 
+func ValidatePodIPsByLabel(label string, expected []string) {
```
fmoral2, line 107: same comment from test functionalities signature

pkg/testcase/ingressdns.go
```diff
@@ -59,3 +61,18 @@ func TestDnsAccess(deleteWorkload bool) {
 		Expect(err).NotTo(HaveOccurred(), "dnsutils manifest not deleted")
 	}
 }
+
+func TestIngressDualStack(testinfo map[string]string) {
```
fmoral2, line 65: https://github.com/rancher/distros-test-framework/pull/13/files#r1339253526

pkg/testcase/networkconnectivity.go
```diff
@@ -31,6 +31,19 @@ func TestInternodeConnectivityMixedOS(deleteWorkload bool) {
 	}
 }
 
+// TestCIDRDualStack Validates Pod IPs and Cluster IPs in CIDR range
```
fmoral2, line 34: the doc name should be the same as the func

pkg/testcase/pod.go
```diff
@@ -59,3 +59,13 @@ func processPodStatus(
 		}
 	}
 }
+
+// TestPodStatusIsRunningByLabel checks status of Running pods when searched by label 
+func TestPodStatusIsRunningByNamespaceAndLabel(namespace, label string) {
```
fmoral2, line 64: this is a generic func, should not be linving in the same file as the test.

please lets use shorter and concise names as golang patterns.

pkg/testcase/service.go
```diff
@@ -88,3 +90,43 @@ func TestServiceLoadBalancer(deleteWorkload bool) {
...
+			clusterIP = shared.IPv6IPEncloser(clusterIP)
+		}
+		err := assert.ValidateOnNode(nodeExternalIPs[0], 
+			"curl -sL --insecure http://"+ clusterIP +":"+ port, testInfo["expected"])
+		Expect(err).NotTo(HaveOccurred(), err)
+	}
+}
+
+func TestServiceSpecContainsSubstring(svc, expected string) {
+	cmd := "kubectl get svc "+svc+" -n default"+
```
fmoral2, line 127: also should not be living in the test file, and its not a test actually right.

shared/aux.go
```diff
@@ -325,3 +326,59 @@ func fileExists(files []os.DirEntry, workload string) bool {
 
 	return false
 }
+
+func IPv6IPEncloser(ip string)string {
```
fmoral2, line 330: if the func using this it will be in  the same file no need to export

shared/cluster.go
```diff
@@ -35,6 +36,8 @@ type Pod struct {
 	Node      string
 }
 
+type ArgsMap map[string]interface{}
```
fmoral2, line 39: if you are using this on the aux. should be living there

shared/cluster.go
mdrahman-suse: Its not the same, we have a func called `FetchClusterIP`... this func is calling `FetchClusterIPs` and if you see the implementation, they make look same but they are calling different resources

config/reader.go
mdrahman-suse: This I added because I had multiple *.tfvars file with different settings so that I don't have to change the same file

entrypoint/dualstack/dualstackcluster_test.go
mdrahman-suse: The test data is needed for the test. I think we should revisit this to enable verbosity in the test itself

shared/cluster.go
fmoral2: ok then, so parametrize it to have only one.

factory/cluster.go
mdrahman-suse: Sorry I am not getting the request. I am using modules path for tf dir

config/reader.go
fmoral2: dont understood 

but why pushing it ?

Makefile
mdrahman-suse: If your PR gets merged before mine, I can rebase, its not biggie

entrypoint/dualstack/dualstackcluster_test.go
fmoral2: you should add in the test itself
not here

factory/cluster.go
mdrahman-suse: Will add it back, need to check how it got overwritten

entrypoint/dualstack/dualstackcluster_test.go
fmoral2: the purpose of this file is exact the opposite, it was built to not have any verbosity

fmoral2 (changes requested):
func signatures and data in the entrypoint

pkg/testcase/service.go
mdrahman-suse: This is kind of asserting and contributes to the test... can move it to assert

shared/aux.go
mdrahman-suse: not same file but may be same package, can check and update accordingly

factory/cluster.go
fmoral2: sorry, i see that wrong

mdrahman-suse:
> func signatures and data in the entrypoint

I was not aware that we are strict about using function signature. I would like to discuss more on this topic

config/reader.go
mdrahman-suse: Is there anything wrong with pushing it?

mdrahman-suse: do we really want to restrict to these file names only?

fmoral2: yeah because we are using only rke2 or k3s tfvars across everything

besides in this way if i pass  `test.tfvars` it will accept ...and it shouldnt if i dont have i file like this.

fmoral2: we need to leave place for min error prone possible, especially on load

entrypoint/dualstack/dualstackcluster_test.go
```diff
@@ -0,0 +1,126 @@
...
+		testdata := map[string]string{
+			"namespace":"default",
+			"label":"app=MyDualApp",
+			"deployment":"httpd-deployment",
+			"expected": "It works!",
+		}
+
+		testcase.TestPodStatusIsRunningByNamespaceAndLabel(testdata["namespace"], testdata["label"])
+		
+		for i, svc := range services {
```
fmoral2, line 108: .

modules/install/join_k3s_agent.sh
```diff
@@ -4,26 +4,56 @@
...
+node-name: $hostname
 EOF
 }
 
 add_config() {
-  local worker_flags="${1}"
-
   if [[ -n "$worker_flags" ]] && [[ "$worker_flags" == *":"* ]]
+  then
+    echo "$worker_flags"
```
rancher-max, line 37: Can probably remove this line -- we don't need worker flags to be output here since we cat the entire config.yaml anyway

modules/install/join_k3s_agent.sh
```diff
@@ -35,65 +65,47 @@ add_config() {
...
+    subscription-manager register --auto-attach --username="$rhel_username" --password="$rhel_password"
+    subscription-manager repos --enable=rhel-7-server-extras-rpms
   fi
 }
 
 disable_cloud_setup() {
-   local node_os="${1}"
-
-if  [[ "$node_os" = *"rhel"* ]] || [[ "$node_os" = *"centos"* ]]
+  if  [[ "$node_os" = *"rhel"* ]] || [[ "$node_os" = *"centos"* ]]
```
rancher-max, line 76: Is this something we want to do regardless of node_os? I don't think it needs to change as part of this PR as this is just a refactor to what is existing, but thought it might be a good discussion point...

modules/install/k3s_master.sh
```diff
@@ -3,123 +3,129 @@
...
 cluster-init: true
+node-name: ${hostname}
 EOF
 }
 
 add_config() {
-  local server_flags="${1}"
+  if [ -n "$server_flags" ] && [[ "$server_flags" == *":"* ]]
+  then
+    echo "$server_flags"
```
rancher-max, line 43: No need to echo server flags since we cat the config.yaml later on

pkg/assert/service.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/rancher/distros-test-framework/shared"
+
+	. "github.com/onsi/gomega"
+)
+
+func CheckServiceSpecContainsSubstring(svc, expected string) {
+	cmd := "kubectl get svc "+svc+" -n default"+
+		" -o jsonpath='{range .items[*]}{.spec}' --kubeconfig="
+	res, err := shared.RunCommandHost(cmd + shared.KubeConfigFile)
```
rancher-max, line 12: This is fine as it ends up being the same thing, but nit:
cmd should specify what the full command actually is, so we should add the `+ shared.KubeConfigFile` to the cmd itself instead of adding it when passing `cmd` to the other function.

pkg/testcase/dualstack.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	assert.CheckPodStatusRunningByNSLabel(testdata["namespace"], testdata["label"])
+	
+	for i, svc := range services {
+		testdata["svc"] = "my-service-"
+		testdata["svc"] += svc
+		TestServiceClusterIPDualStack(testdata)
+		assert.CheckServiceSpecContainsSubstring(testdata["svc"], expectedIPFamily[i])
+	}
```
rancher-max, line 96: Will this fail if the first service in the range fails? If so, is that desired, or do we want to see the full list of services that fail? If we want the full list of services, then we might need to refactor this block a bit.

pkg/testcase/ingressdns.go
```diff
@@ -59,3 +61,18 @@ func TestDnsAccess(deleteWorkload bool) {
...
+
+	for _, ingressIP := range ingressIPs {
+		if strings.Contains(ingressIP,":") {
+			ingressIP = IPv6IPEncloser(ingressIP)
+		}
+		err := assert.ValidateOnNode(shared.BastionIP,
+			"curl -sL -H 'Host: test1.com' http://"+ ingressIP +"/name.html",
+			testinfo["expected"])
+		Expect(err).NotTo(HaveOccurred(), err)
+	}
```
rancher-max, line 77: Same comment as above -- basically: fail on first or fail after all are checked?

rancher-max (commented):
I have not finished review yet, but wanted to get this posted today. Left some comments on specific areas.

Also, nit: Some of the assertion function names are switching between "Validate..." and "Check...". We should probably come up with some standard here across the board.

config/reader.go
mdrahman-suse: Ok, updated

entrypoint/dualstack/dualstackcluster_test.go
mdrahman-suse: Ok, Updated

entrypoint/dualstack/dualstackcluster_test.go
mdrahman-suse: Ok, updated

pkg/testcase/networkconnectivity.go
mdrahman-suse: Updated

pkg/testcase/ingressdns.go
mdrahman-suse: Please review

pkg/testcase/pod.go
mdrahman-suse: Updated

pkg/testcase/service.go
mdrahman-suse: Updated

shared/cluster.go
mdrahman-suse: Updated

entrypoint/dualstack/dualstackcluster_test.go
mdrahman-suse: ?

modules/install/join_k3s_agent.sh
mdrahman-suse: Agrees, lets discuss and I can add that change

shared/cluster.go
```diff
@@ -200,6 +202,23 @@ func FetchClusterIP(namespace string, serviceName string) (string, string, error
...
+		" -o jsonpath='{.spec.clusterIPs[*]}' --kubeconfig=" + KubeConfigFile)
+	if err != nil {
+		return "", "", err
+	}
+
+	port, err := RunCommandHost("kubectl get svc " + svc + " -n " + namespace +
+		" -o jsonpath='{.spec.ports[0].port}' --kubeconfig=" + KubeConfigFile)
+	if err != nil {
+		return "", "", err
+	}
```
rancher-max, line 217: Some example places where maybe using the `KubectlCommand` function would be good

pkg/testcase/ingressdns.go
mdrahman-suse: Ideally should fail on first but I am on the fence on how we want to go about doing it. If we continue even on fail, there is currently no way to run tests with specific data unless we want to separate the tests itself and we have to re-run the test anyways if anything fails...

shared/cluster.go
mdrahman-suse: Actually I think I will move away from using `KubectlCommand` and write the cmd directly so that its easier to understand what command was run as part of the test

mdrahman-suse: and also printing out the command on failure would help with debugging

pkg/assert/service.go
mdrahman-suse: Thanks, updated

modules/install/k3s_master.sh
mdrahman-suse: Updated

modules/install/join_k3s_agent.sh
mdrahman-suse: Updated

modules/install/join_rke2_agent.sh
```diff
@@ -1,39 +1,50 @@
...
-   echo "$worker_flags"
-   echo -e "$worker_flags" >> /etc/rancher/rke2/config.yaml
-   if [[ "$worker_flags" != *"cloud-provider-name"* ]]
-   then
-     echo -e "node-external-ip: $public_ip" >> /etc/rancher/rke2/config.yaml
-   fi
-   cat /etc/rancher/rke2/config.yaml
-else
-  echo -e "node-external-ip: $public_ip" >> /etc/rancher/rke2/config.yaml
+  echo "$worker_flags"
```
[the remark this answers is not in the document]
mdrahman-suse: updated

modules/install/join_rke2_master.sh
```diff
@@ -1,44 +1,55 @@
...
-   echo "$server_flags"
-   echo -e "$server_flags" >> /etc/rancher/rke2/config.yaml
-   if [[ "$server_flags" != *"cloud-provider-name"* ]]
-   then
-     echo -e "node-external-ip: $public_ip" >> /etc/rancher/rke2/config.yaml
-   fi
-   cat /etc/rancher/rke2/config.yaml
-else
-  echo -e "node-external-ip: $public_ip" >> /etc/rancher/rke2/config.yaml
+  echo "$server_flags"
```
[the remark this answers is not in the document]
mdrahman-suse: updated

modules/install/rke2_master.sh
```diff
@@ -5,43 +5,55 @@
...
-   echo "$server_flags"
-   echo -e "$server_flags" >> /etc/rancher/rke2/config.yaml
-   if [[ "$server_flags" != *"cloud-provider-name"* ]]
-   then
-     echo -e "node-external-ip: $public_ip" >> /etc/rancher/rke2/config.yaml
-   fi
-   cat /etc/rancher/rke2/config.yaml
-else
-  echo -e "node-external-ip: $public_ip" >> /etc/rancher/rke2/config.yaml
+  echo "$server_flags"
```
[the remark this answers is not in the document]
mdrahman-suse: updated

pkg/testcase/networkconnectivity.go
```diff
@@ -31,6 +31,19 @@ func TestInternodeConnectivityMixedOS(deleteWorkload bool) {
 	}
 }
 
+// TestIPsInCIDRRangeDualStack Validates Pod IPs and Cluster IPs in CIDR range
+func TestIPsInCIDRRangeDualStack(label, svc string){
+	nodeArgs,err := shared.GetNodeArgs("server")
+	Expect(err).NotTo(HaveOccurred(), err)
+
+	// TODO: Decide whether to loop through or not as the info will be same across all nodes
```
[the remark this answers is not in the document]
mdrahman-suse: updated

shared/aux.go
```diff
@@ -9,12 +9,15 @@ import (
...
 	"runtime"
 	"strings"
+	"regexp"
 
 	"github.com/rancher/distros-test-framework/config"
 	"github.com/rancher/distros-test-framework/pkg/log"
 	"golang.org/x/crypto/ssh"
 )
 
+type ArgsMap map[string]interface{}
```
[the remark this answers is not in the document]
mdrahman-suse: updated

shared/cluster.go
```diff
@@ -148,6 +149,7 @@ func KubectlCommand(destination, action, source string, args ...string) (string,
 	}
 
 	cmd := cmdPrefix + " " + source + " " + strings.Join(args, " ") + kubeconfigFlag
+	fmt.Println("Kubectl command: ", cmd)
```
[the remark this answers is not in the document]
mdrahman-suse: I think it was useful when debugging, removed and updated

modules/install/join_k3s_agent.sh
rancher-max: We discussed internally and decided we might add to a future PR, but not necessary as part of this one.

shared/cluster.go
rancher-max: That's fine with me -- we should standardize in that case across the repo (in a future PR)

rancher-max (approved):

pkg/testcase/ingressdns.go
rancher-max: We decided internally that this is fine as is!

shared/cluster.go
fmoral2: as per private discussion MD will implement a better approach that he provided.
create a generic function that acts based on a key to get something from spec.

pkg/assert/service.go
```diff
@@ -0,0 +1,40 @@
...
+	"net"
+	"strings"
+
+	"github.com/rancher/distros-test-framework/shared"
+
+	. "github.com/onsi/gomega"
+)
+
+// CheckServiceSpecContainsSubstring retrives service specs and validates expected substring
+func CheckServiceSpecContainsSubstring(svc, expected string) {
```
fmoral2, line 13: can we keep naming short and concise ?  idk, something like " CheckSvcSpec " ?  and then in the documentation you can do verbose if you want
// CheckSvcSpec retrives service specs and validates expected substring

pkg/testcase/service.go
```diff
@@ -88,3 +90,37 @@ func TestServiceLoadBalancer(deleteWorkload bool) {
 		Expect(err).NotTo(HaveOccurred(), "Loadbalancer manifest not deleted")
 	}
 }
+
+func TestServiceNodePortDualStack(testinfo map[string]string) {
```
fmoral2, line 94: https://github.com/rancher/distros-test-framework/pull/13/files#r1349128583

pkg/testcase/service.go
```diff
@@ -88,3 +90,37 @@ func TestServiceLoadBalancer(deleteWorkload bool) {
...
+		}
+		err = assert.CheckComponentCmdNode(
+			"curl -sL --insecure http://"+ip+":"+nodeport+"/name.html",
+			testinfo["expected"],
+			shared.BastionIP)
+		Expect(err).NotTo(HaveOccurred(), err)
+	}
+}
+
+func TestServiceClusterIPDualStack(testInfo map[string]string) {
```
fmoral2, line 111: https://github.com/rancher/distros-test-framework/pull/13/files#r1349128583

pkg/testcase/dualstack.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/rancher/distros-test-framework/pkg/assert"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestDualStackIngress(deleteWorkload bool) {
+	_, err := shared.ManageWorkload("apply", "dualstack-ingress.yaml")
+	Expect(err).NotTo(HaveOccurred())
+	
+	testdata := map[string]string{
```
fmoral2, line 14: since u wanted to have somehitng more verbose and kind of data driven
in go is a best practice create table driven tests

someething like : 

```
type testData struct {
	TestDataMap      anyname
	ExpectedIPFamily []string
}

type anyname struct {
	namespace string
	label     string
	svc       string
	exp       string
}


and then use it across your functions. 
its more organized 

This is only a advise, not a must do.
```

shared/aux.go
```diff
@@ -325,3 +326,59 @@ func fileExists(files []os.DirEntry, workload string) bool {
...
+		if argsMap["node-type"] == nodeType {
+			return argsMap, nil
+		}
+	}
+
+	return nil, nil
+}
+
+func processNodeArgsResponse(res string) ([]map[string]string){
+	nodeArgsMapSlice := []map[string]string{}
```
fmoral2, line 361: here normally declaring a empty data use
	var nodeArgsMapSlice []map[string]string

shared/aux.go
```diff
@@ -325,3 +326,59 @@ func fileExists(files []os.DirEntry, workload string) bool {
...
+	for _, argsMap := range argsMapSlice {
+		if argsMap["node-type"] == nodeType {
+			return argsMap, nil
+		}
+	}
+
+	return nil, nil
+}
+
+func processNodeArgsResponse(res string) ([]map[string]string){
```
fmoral2, line 360: ([]map[string]string)

no need for this ()

shared/aux.go
```diff
@@ -325,3 +326,59 @@ func fileExists(files []os.DirEntry, workload string) bool {
...
+func GetNodeArgs(nodeType string) (map[string]string, error){
+	product,err := GetProduct()
+	if err != nil{
+		return nil, err
+	}
+	res,err := KubectlCommand(
+		"host",
+		"get",
+		"nodes " + 
+		fmt.Sprintf(`-o jsonpath='{range .items[*]}{.metadata.annotations.%s\.io/node-args}{end}'`, product),
```
fmoral2, line 343: passing line lenght

shared/aux.go
```diff
@@ -325,3 +326,59 @@ func fileExists(files []os.DirEntry, workload string) bool {
...
+
+func processNodeArgsResponse(res string) ([]map[string]string){
+	nodeArgsMapSlice := []map[string]string{}
+	resSlice := strings.Split(res,"]")
+	
+	for _,resItem := range(resSlice[:(len(resSlice)-1)]){
+		resItemSlice := strings.Split(resItem,`","`)
+		nodeArgsMap := map[string]string{}
+
+		for range(resItemSlice[1:]){
```
fmoral2, line 368: you probably dont need this for resitemslice

fmoral2:
how is your lint on github actions?

mdrahman-suse:
> how is your lint on github actions?

Here: https://github.com/mdrahman-suse/distros-test-framework/actions

pkg/testcase/networkconnectivity.go
```diff
@@ -31,6 +31,18 @@ func TestInternodeConnectivityMixedOS(deleteWorkload bool) {
 	}
 }
 
+// TestIPsInCIDRRangeDualStack Validates Pod IPs and Cluster IPs in CIDR range
+func TestIPsInCIDRRangeDualStack(label, svc string){
```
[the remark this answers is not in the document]
mdrahman-suse: So if it is `validate` then I presume its better suited to be in `assert` module?

shared/aux.go
mdrahman-suse: I need to iterate starting from second item and skip the first item in the slice, is it not how it should be done?

pkg/testcase/dualstack.go
fmoral2: event its just here , since u are using a lot of maps its a way to organize better and have more readability .
its a improvement, no need if u dont want to.

shared/aux.go
fmoral2: those assign 
nodeArgsMap["node-type"] = strings.Trim(resItemSlice[0],`["`)
			regxCompile := regexp.MustCompile(`--|"`)

u probably can done only once since in all iterations of 2 second loop the resItemSlice[0]  , does not change.

fmoral2: try remove 

for range(resItemSlice[1:]){.  and the close }

pkg/testcase/networkconnectivity.go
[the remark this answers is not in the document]
fmoral2: not necessarily , if this is a validation function scpecif only to this test, should live with the test.

if it is a generic validation that can be used if more than 1 test, then yes. Assert package

modules/k3s/main.tf
```diff
@@ -1,58 +1,102 @@
...
-   instance_class=var.instance_class
-   external_db_version=var.external_db_version
-   engine_mode=var.engine_mode
-   environment=var.environment
-   create_lb=var.create_lb
-   k3s_channel = var.k3s_channel
+   source = "./master"
+
+   # Basic Variables
+   username            = var.username
```
ShylajaDevadiga, line 5: can we change this to rhel_username to be consistent with variable name?

ShylajaDevadiga (commented):
Areas to verify before merging the PR
For k3s
1. etcd datastore/external datastore
For both
1. Pass channel/skip channel
2. Pass version/skip version (test locally, stable channel version should be used)
3. ipv4 only
4. ipv6 only
5. dual stack
6. rpm based installs

PR #16, opened 2023-08-23
Revert "[READY FOR REVIEW] Add windows test and modules"

ShylajaDevadiga:
Reverts rancher/distros-test-framework#6


mdrahman-suse (approved):

Makefile
```diff
@@ -47,8 +42,8 @@ test-complete: test-env-clean test-env-down remove-tf-state test-env-up test-run
 
 .PHONY: remove-tf-state
 remove-tf-state:
-	@rm -rf ./modules/${PRODUCT}/.terraform
-	@rm -rf ./modules/${PRODUCT}/.terraform.lock.hcl ./modules/${PRODUCT}/terraform.tfstate ./modules/${PRODUCT}/terraform.tfstate.backup
+	@rm -rf ./modules/k3scluster/.terraform
```
aganesh-suse, line 45: curious: our goal is to have a common front for both k3s and rke2 ? so should nt  we try to use "PRODUCT" variable as much as possible - instead of hard coding 'k3s' name wherever possible? or would we be having a separate makefile for k3s and rke2?

mdrahman-suse: Yeah that is the plan and this will be handled in the Makefile change task: https://github.com/rancher/qa-tasks/issues/854

aganesh-suse (approved):

Issue #19, opened 2023-09-05
k3s not installing properly on rocky  because of cloud setup enabled


fmoral2:
- Steps: 

1. Create an HA cluster with rocky AMIs
2. Run any test 

Expected behavior:
- creates the cluster normally

Actual behavior:
- Cluster fails with cloud setup enabled message

PR #20, opened 2023-09-12
 fix(modules): cloud setup enabled

fmoral2:
#### Proposed Changes ####

fix https://github.com/rancher/distros-test-framework/issues/19 

#### Types of Changes ####

- adding the disable systemctl disable nm-cloud-setup.service  and systemctl disable nm-cloud-setup.timer
  for rhel and rocky  

- improvements on parsing nodes 

- adding volume size to terraform 

#### Verification ####

- Testing OS.s with different versions

#### Testing ####
- i have tested k3s and rke2 in 1.26,1.27 and 1.28 with rhel ,sles ,ubuntu and oracle. 

#### Linked Issues ####

https://github.com/rancher/distros-test-framework/issues/19


modules/k3s/main.tf
```diff
@@ -16,6 +16,7 @@ module "master" {
    server_flags=var.server_flags
    availability_zone=var.availability_zone
    sg_id=var.sg_id
+   volume_size=var.volume_size
```
mdrahman-suse, line 19: Suggestion to add a default value for volume size in case it's not provided.
Similar changes where applicable

mdrahman-suse (changes requested):
Please review the comment

modules/k3s/main.tf
fmoral2: right now we dont have that even on rke2, this is a fix PR that i added this to be able to tests on minor.
we should open a task for those behaviors!

pkg/assert/node.go
```diff
@@ -30,7 +30,8 @@ func NodeAssertVersionTypeUpgrade(c customflag.FlagConfig) NodeAssertFunc {
 func assertVersion(c customflag.FlagConfig) NodeAssertFunc {
 	fmt.Printf("Asserting Version: %s\n", c.InstallMode.Version)
 	return func(g Gomega, node shared.Node) {
-		g.Expect(node.Version).Should(ContainSubstring(c.InstallMode.Version),
+		version := strings.Split(c.InstallMode.Version, "-")
```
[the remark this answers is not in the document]
fmoral2: the split func it will ignore, and return the correct value

modules/k3s/main.tf
mdrahman-suse: Sure sound good to me

shared/cluster.go
```diff
@@ -273,54 +273,95 @@ func SonobuoyMixedOS(action, version string) error {
 	return err
 }
 
-// ParseNodes returns nodes parsed from kubectl get nodes.
-func ParseNodes(print bool) ([]Node, error) {
-	nodes := make([]Node, 0, 10)
+// GetNodes returns nodes parsed from kubectl get nodes.
+func GetNodes(print bool) ([]Node, error) {
```
ShylajaDevadiga, line 277: Is this change solving any issues seen with ParseNodes? I'm trying to see if we can keep it simple and small when possible. We can avoid printing the connection error with print boolean false and run it again with print true.

fmoral2: i also do agree....but the only thing with that is like if we do not treat the errors separately ( errors in general, and errors because the cluster is down due to upgrades) when we found some "errror", that we will find in all upgrade scenarios , we are going to do what we were doing, return an error or fmt.Errorf() , but its not actually an error, its just the cluster temporarily unavailable. 
and then on the execution itself, even not printing , it will return the controler of the function as error.

But again, this is not a blocking error, the running of the tests will not abort or even quit the test as failed....its just a matter of nor return error or printing it to the logs, when its not .

With we think that here is more important simplicity than cohesion , then we should just mantain as it was though , like retrying get nodes until eventually it succeeds and in this meantime time in the logs will be there like something as failed ( the way is now the main ) 

https://github.com/rancher/distros-test-framework/blob/67d29dbe12b3fe8703d4528a711012ca3db0211f/shared/aux.go#L36C17-L36C17

pkg/assert/node.go
```diff
@@ -57,7 +58,8 @@ func assertCommit(c customflag.FlagConfig) NodeAssertFunc {
 // NodeAssertVersionUpgraded custom assertion func that asserts that node version is as expected
 func NodeAssertVersionUpgraded() NodeAssertFunc {
 	return func(g Gomega, node shared.Node) {
-		g.Expect(&customflag.ServiceFlag.SUCUpgradeVersion).Should(ContainSubstring(node.Version),
+		version := strings.Split(customflag.ServiceFlag.SUCUpgradeVersion.String(), "-")
```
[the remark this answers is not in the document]
fmoral2: the split func it will ignore, and return the correct value

rancher-max (approved):
LGTM

VestigeJ (approved):
LGTM

mdrahman-suse (approved):
LGTM

PR #21, opened 2023-09-13
Change Readme Steps and add example config files

aganesh-suse:
Changes involved: 
1. Added steps in README.md file:
    a. Setup changes to get the test runs started in local environments.
    b. Help for macOS M2 chip machine to run tests.
2. Added following files to config/examples folder: 
   a. k3s.tfvars file
   b. rke2.tfvars file
   c. config.yaml file

https://github.com/rancher/qa-tasks/issues/587


config/k3s.tfvars.example
```diff
@@ -0,0 +1,83 @@
...
+
+install_mode   = "INSTALL_K3S_VERSION"
+# INSTALL_K3S_VERSION or INSTALL_K3S_COMMIT
+
+create_lb      = false
+arch           = "amd64"
+
+## Custom Vars
+node_os            = "sles15"
+aws_ami            = "ami-046cd3113e0c1b581"
```
mdrahman-suse, line 22: aws_ami info should not be listed here

config/rke2.tfvars.example
```diff
@@ -0,0 +1,59 @@
...
+
+## Windows agent variables
+#server_flags   = "cni: calico\n"
+windows_ec2_instance_class  = "t3.xlarge"
+windows_aws_ami             = "ami-05a418fd6eb36fd5b"
+no_of_windows_worker_nodes  = 0
+
+## Custom Vars
+node_os            = "sles15"
+aws_ami            = "ami-046cd3113e0c1b581"
```
mdrahman-suse, line 33: Redact ami info

config/rke2.tfvars.example
```diff
@@ -0,0 +1,59 @@
...
+optional_files = ""
+# INSTALL_RKE2_VERSION or INSTALL_RKE2_COMMIT
+install_mode   = "INSTALL_RKE2_VERSION"
+# leave blank or choose 'tar' or 'rpm'; For selinux testing, set to 'rpm' mode of install
+install_method = ""
+
+## Windows agent variables
+#server_flags   = "cni: calico\n"
+windows_ec2_instance_class  = "t3.xlarge"
+windows_aws_ami             = "ami-05a418fd6eb36fd5b"
```
mdrahman-suse, line 28: redact ami info

config/.ssh/aws_key.pem.example
```diff
@@ -0,0 +1,3 @@
+-----BEGIN RSA PRIVATE KEY-----
+ABCD
```
fmoral2, line 2: tbh i dont a see a value , specifying a path to the user creating to add his key. 
we can just give the example without doing this.

fmoral2: i prefer to use this key in a different place , even though its this path we use in the container, locally i like to use another path.

config/config.yaml.example
```diff
@@ -0,0 +1,2 @@
+ENV_PRODUCT: k3s
```
fmoral2, line 1: add the value as variable = {product}

config/config.yaml.example
```diff
@@ -0,0 +1,2 @@
...
+ENV_TFVARS: k3s.tfvars
```
fmoral2, line 2: same here

config/.ssh/aws_key.pem.example
fmoral2: so if i like to do this, maybe others as well

config/config.yaml.example
aganesh-suse: isnt this supposed to be a yaml file format? we should be doing 'variable: value' in which case?

config/.ssh/aws_key.pem.example
aganesh-suse: 2 reasons i added this file.
1) to have created the .ssh folder and checked the same in. 
2) To have the permission of the file preset to 600 - so we can simply rename the file and use the same. 
Without this permission, the test run fails - and asks us to have strict permissions. this is simply to avoid that headache.

config/config.yaml.example
aganesh-suse: same as previous comment. sticking with yaml format of `variable: value`

config/.ssh/aws_key.pem.example
fmoral2: `If someone chooses to not use the same, they can of course. It is always a good practise to provide a solution-when it is possible, than let people figure out their own solutions.`

Exactly!! thats why i mentioned to add this as an "example" that were it could be in comments....not actually creating the folder....got it ?

config/config.yaml.example
fmoral2: why do you think use a var in the value is not a yaml ?

config/config.yaml.example
fmoral2: same as above

config/config.yaml.example
aganesh-suse: i think we both misunderstood each other. my problem was with using "=" instead of the ":" in a yaml. - since that is not yaml formatting. got where you are coming from now 👍

aganesh-suse: Fixed.

config/config.yaml.example
aganesh-suse: Fixed.

config/.ssh/aws_key.pem.example
aganesh-suse: This has been added as an "example" file. 
I do not see harm in leaving this in there as an option, since there are not actual secrets shared. 
The main points being that the file permissions have been preset and one does not have to mount files in the lima environment. I do not feel there is a good enough reason to remove this file. Please feel free to not use the same.

aganesh-suse: Since we moved all the files into an examples folder - removing this .pem file example from this location.

rancher-max (approved):

modules/k3s/variables.tf
```diff
@@ -14,7 +14,9 @@ variable "instance_class" {}
 variable "ec2_instance_class" {}
 variable "db_group_name" {}
 variable "username" {}
-variable "password" {}
+variable "password" {
+  default = "Pa$$w0rd"
```
ShylajaDevadiga, line 18: Is default required? Can we leave it blank?

ShylajaDevadiga (approved):
One small comment. Otherwise LGTM

modules/k3s/variables.tf
mdrahman-suse: +1

mdrahman-suse (approved):
LGTM, once Shylaja's comment is addressed

mdrahman-suse:
Also I see 8 commits, don't forget to squash your commits before merging!

modules/k3s/variables.tf
aganesh-suse: fixed. removed this change.

rancher-max (approved):
LGTM -- when merging, select the "Squash and Merge" option

rancher-max (approved):

config/examples/k3s.tfvars
```diff
@@ -0,0 +1,83 @@
...
+k3s_version    = "v1.27.4+k3s1"
+k3s_channel    = "latest"
+server_flags   = "protect-kernel-defaults: true\n"
+worker_flags   = "protect-kernel-defaults: true\n"
+no_of_server_nodes = 3
+no_of_worker_nodes = 1
+resource_name  = "<prefix_name_for_your_resources>"
+key_name       = "jenkins-rke-validation"
+access_key     = "/go/src/github.com/rancher/distros-test-framework/config/.ssh/aws_key.pem"
+password       = "<enter_your_password>"
```
ShylajaDevadiga, line 12: What is password for?

ShylajaDevadiga (approved):

ShylajaDevadiga (approved):

rancher-max (approved):

mdrahman-suse (approved):
LGTM, dont forget to squash and merge

PR #22, opened 2023-09-15
add: coredns bump version 

fmoral2:
#### Proposed Changes ####

- add coredns bump version

#### Types of Changes ####

- add coredns bump version

 

#### Verification ####

- Testing the file 

#### Linked Issues ####

https://github.com/rancher/distros-test-framework/issues/19


entrypoint/versionbump/coredns.go
```diff
@@ -0,0 +1,63 @@
...
+	})
+
+	It("Validate Pods", func() {
+		testcase.TestPodStatus(
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus())
+	})
+
+	It("Verifies bump version for coredns on rke2", func() {
```
mdrahman-suse, line 34: Do we need to add a check for product? This test should not be executed for k3s, right?

rancher-max (changes requested):
Can we also add an "It" test case here at the end to validate that DNS queries are still successful? Copying this one should be fine: https://github.com/rancher/distros-test-framework/blob/main/entrypoint/createcluster/createcluster_test.go#L48-L50

entrypoint/versionbump/coredns.go
fmoral2: rke2 only

entrypoint/versionbump/coredns.go
```diff
@@ -0,0 +1,63 @@
...
+		template.VersionTemplate(template.VersionTestTemplate{
+			TestCombination: &template.RunCmd{
+				Run: []template.TestMap{
+					{
+						Cmd: "kubectl get all -l k8s-app=kube-dns -n kube-system -o wide," +
+							"kubectl exec -n dnsutils -t dnsutils : -- nslookup kubernetes.default",
```
fmoral2, line 40: .

entrypoint/versionbump/coredns.go
```diff
@@ -0,0 +1,63 @@
...
+						ExpectedValue:        template.TestMapTemplate.ExpectedValue,
```
fmoral2, line 41: .

entrypoint/versionbump/coredns.go
fmoral2, line 40: .

rancher-max (commented):
Nevermind on my previous review -- it's already covered based on the args passed when running the test: `-workloadName "dnsutils.yaml" -deployWorkload true`

Just MD's comment then this LGTM

entrypoint/versionbump/coredns.go
fmoral2: you mean like break if it is k3s ? 

this we will add later on when improved the main entrypoint , right now all version bumps that are for a single product we cant do much expect 

`if cfg.product == {product} { os.exit()} ` 

but thats not a good option

fmoral2: task related 
https://github.com/rancher/qa-tasks/issues/803

mdrahman-suse (approved):
LGTM

rancher-max (approved):

PR #23, opened 2023-09-19
[NOT READY] - RKE2 Automation to assign a Elastic IP to a Cluster and perform a reboot test

est-suse:
Being able to automate the reboot process, this is currently perform on Linode, but it can be done using ElasticIP on AWS. Once the cluster is created the will execute the same test we have for create cluster.

Considerations:

Please create the variable "create_eip         = true" in your rke2.tfvars file, if true will create cluster and assign IP static to every one of the nodes, if the value is set to false, the create cluster creation is perform as we have it now.

https://github.com/rancher/qa-tasks/issues/591
https://github.com/rancher/qa-tasks/issues/592


modules/install/rke2_stop_start_instace.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
fmoral2, line 1: typo on the file name = "instace"

modules/install/install_rke2_master_eip.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
fmoral2, line 1: dont create a new file for this, use the existing ones
refactor those to accept your needs parametrized

modules/install/join_rke2_agent_eip.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
fmoral2, line 1: https://github.com/rancher/distros-test-framework/pull/23/files#r1330632639

modules/install/join_rke2_master_eip.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
fmoral2, line 1: https://github.com/rancher/distros-test-framework/pull/23/files#r1330632639

modules/install/rke2_stop_start_instace.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash  
+#set -x
+echo "$@"
+type=$1
+resource_name=$2
+
+cd /tmp
+if [ "$type" == "stop" ]; then
```
fmoral2, line 8: we should not hard code the number of nodes, instead use as var and treat that here and tf module

opinions here @rancher-max @ShylajaDevadiga  ?

modules/rke2/master/instances_server.tf
```diff
@@ -64,6 +64,97 @@ resource "aws_instance" "master" {
   }
 }
 
+
+locals {
+        eip_index = { for i, v in aws_instance.master2 : tonumber(i)  => v.id if var.create_eip}      
+}
+
+resource "aws_eip" "stop-eip-master" {
```
fmoral2, line 72: take a look on the stop, because its stopping even before install rke2

modules/install/rke2_stop_start_instace.sh
fmoral2, line 1: run a shell check in this file, it has some important improvements

fmoral2:
please add the qase link for the test or the qa-task , so we can know which test exactly is being created

rancher-max:
> please add the qase link for the test or the qa-task on the PR description , so we can know which test exactly is being created

I added to the description! 👍

modules/rke2/master/instances_server.tf
```diff
@@ -64,6 +64,97 @@ resource "aws_instance" "master" {
   }
 }
 
+
```
fmoral2, line 67: on here also i would ask help/opinion to @matttrach to see if its a better solution instead creating resources separately for elastic ip = true or = false.

like , if we can work on a single "resource creation" checking this var, if its true we create the elastic, and if it false we create the "normal ones"

modules/install/rke2_stop_start_instace.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash  
+#set -x
+echo "$@"
+type=$1
```
ShylajaDevadiga, line 4: Let's add a comment here on what this variable is for. Could we call it instanceState?

modules/rke2/master/instances_server.tf
ShylajaDevadiga, line 72: Let's call it something like "master_with_eip"

modules/rke2/master/instances_server.tf
```diff
@@ -64,6 +64,97 @@ resource "aws_instance" "master" {
...
+resource "aws_eip" "stop-eip-master" { 
+        count = var.create_eip ? 1 : 0
+        vpc = true
+        tags = {
+           Name ="${var.resource_name}-server"
+        }     
+}
+
+resource "aws_eip" "stop-eip-master2" {
```
ShylajaDevadiga, line 80: same

modules/rke2/master/instances_server.tf
```diff
@@ -64,6 +64,97 @@ resource "aws_instance" "master" {
...
+        for_each = local.eip_index
+        vpc = true
+          tags = {
+                  Name ="${var.resource_name}-servers-${each.key}"
+        }
+      depends_on = [aws_eip.stop-eip-master]
+}
+
+resource "aws_eip_association" "master-stop-association" {
```
ShylajaDevadiga, line 89: Rename to "master_eip_association" as what its functionality does

modules/rke2/worker/instances_worker.tf
```diff
@@ -38,6 +38,58 @@ resource "aws_instance" "worker" {
   }
 }
 
+locals {
+        eip_index = { for i, v in aws_instance.worker : tonumber(i)  => v.id if var.create_eip}      
+}
+
+resource "aws_eip" "worker-eip" {
```
ShylajaDevadiga, line 45: Let's keep name consistent with instances_server

modules/rke2/worker/instances_worker.tf
```diff
@@ -38,6 +38,58 @@ resource "aws_instance" "worker" {
...
+  depends_on = [var.dependency]
+  for_each = local.eip_index
+  vpc = true
+  tags = {
+      Name ="${var.resource_name}-worker-${each.key}"
+    }
+}
+
+resource "aws_eip_association" "worker-association" {
```
ShylajaDevadiga, line 54: Same

modules/rke2/worker/instances_worker.tf
```diff
@@ -54,4 +106,37 @@ data "local_file" "token" {
 
 locals {
   node_token = trimspace("${data.local_file.token.content}")
+}
+
+resource "null_resource" "stop_resource" {
```
ShylajaDevadiga, line 111: stop_all_nodes

modules/rke2/worker/instances_worker.tf
```diff
@@ -54,4 +106,37 @@ data "local_file" "token" {
...
+  count = var.create_eip ? 1 : 0
+  create_duration = "400s"
+  depends_on = [null_resource.stop_resource]
+}
+
+resource "null_resource" "start_server1_server2" {
+  count = var.create_eip ? 1 : 0
+  depends_on = [time_sleep.wait_for_stop]
+  provisioner "local-exec" {
+    command = "../install/rke2_stop_start_instace.sh start_s1_s2 ${var.resource_name}"
```
ShylajaDevadiga, line 132: We should start all server nodes at once and then start agent nodes.

modules/install/rke2_stop_start_instace.sh
ShylajaDevadiga: Since they are elastic ips, i would suggest let's limit the count to 3 server and 1 agent.

modules/rke2/master/instances_server.tf
est-suse: Can you elaborate more on this?
If I understand correctly, this is how it is right now.
if the elastic ip = true we are going to proceed to create the elasticip and associate them to the cluster.
if the var elastic ip = false, the process is not going to execute any of the steps relate to the creation / association of the elasticip.

If I run without 

image
 
No elastic_ip are created:
image

Setting Var elastic ip = true: 

image
image

modules/install/install_rke2_master_eip.sh
est-suse: thanks done!

modules/rke2/master/instances_server.tf
est-suse: Thanks

modules/install/rke2_stop_start_instace.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash  
+#set -x
+echo "$@"
```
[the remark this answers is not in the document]
est-suse: done!

modules/install/rke2_stop_start_instace.sh
est-suse: done!

mdrahman-suse:
I was wondering if there is anyway to reboot the instances using terraform

modules/install/rke2_master.sh
```diff
@@ -1,121 +1,151 @@
...
-
-hostname=$(hostname -f)
+
+create_directories() {
 mkdir -p /etc/rancher/rke2
-cat << EOF >/etc/rancher/rke2/config.yaml
+}
+
+create_config() {
+  local fake_fqdn="${1}"
```
fmoral2, line 10: if using locals, give a break line to rest of the code, to make it more readable

modules/install/rke2_master.sh
```diff
@@ -1,121 +1,151 @@
...
-   echo "$server_flags"
-   echo -e "$server_flags" >> /etc/rancher/rke2/config.yaml
-   if [[ "$server_flags" != *"cloud-provider-name"* ]]
-   then
-     echo -e "node-external-ip: $public_ip" >> /etc/rancher/rke2/config.yaml
-   fi
-   cat /etc/rancher/rke2/config.yaml
-else
+add_config() {
+  local server_flags="${1}"
```
fmoral2, line 21: https://github.com/rancher/distros-test-framework/pull/23/files#r1343092273

modules/install/rke2_master.sh
```diff
@@ -1,121 +1,151 @@
...
-   subscription-manager repos --enable=rhel-7-server-extras-rpms
-fi
+network_manager() {
+  local node_os="${1}"
+  local rhel_username="${2}"
+  local rhel_password="${3}"
 
-if [[ "$node_os" = *"centos"* ]] || [[ "$node_os" = *"rhel"* ]] || [[ "$node_os" = *"oracle"* ]]
-then
-  NM_CLOUD_SETUP_SERVICE_ENABLED=$(systemctl status nm-cloud-setup.service | grep -i enabled)
```
fmoral2, line 49: why did you take this out ?

modules/install/rke2_stop_start_instance.sh
```diff
@@ -0,0 +1,60 @@
...
+        aws ec2 describe-instances --filters "Name=tag:Name,Values=${resource_name}-worker" \
+         "Name=instance-state-name,Values=running" \
+        --output text --query 'Reservations[*].Instances[*].InstanceId' > /tmp/ids_master_worker.txt    
+
+        cat /tmp/ids_server_1_2.txt /tmp/ids_master_worker.txt > /tmp/ids_all.txt
+    fi
+}
+
+assign_file(){
+    local instanceState="${1}"
```
fmoral2, line 29: https://github.com/rancher/distros-test-framework/pull/23/files#r1343092273

modules/install/rke2_stop_start_instance.sh
```diff
@@ -0,0 +1,60 @@
...
+        file="ids_all.txt" 
+    elif [ "${1}" == "start_s1_s2" ]; then 
+        file="ids_server_1_2.txt" 
+    elif [ "${1}" == "start_master_worker" ]; then
+        file="ids_master_worker.txt" 
+    fi
+}
+
+stop_start_nodes(){
+    local instanceState="${1}"
```
fmoral2, line 40: https://github.com/rancher/distros-test-framework/pull/23/files#r1343092273

fmoral2:
> I was wondering if there is anyway to reboot the instances using terraform

i was thinking on that too, but since this is kind of part of a " execution or test flow " and not directly tied to infra configuration, think thats better have as a sh script or even a go aux function.

mdrahman-suse:
> > I was wondering if there is anyway to reboot the instances using terraform
> 
> i was thinking on that too, but since this is kind of part of a " execution or test flow " and not directly tied to infra configuration, think thats better have as a sh script or even a go aux function.
> 
> what u guys think ?

hmm, that sounds reasonable... I am leaning more towards having a functionality in code rather than shell scripts, may be utilize the aws-sdk for golang? That way we can have the flexibility of performing instance actions more seamlessly... have a separate module for that as part of the framework?

fmoral2:
> > > I was wondering if there is anyway to reboot the instances using terraform
> > 
> > 
> > i was thinking on that too, but since this is kind of part of a " execution or test flow " and not directly tied to infra configuration, think thats better have as a sh script or even a go aux function.
> > what u guys think ?
> 
> hmm, that sounds reasonable... I am leaning more towards having a functionality in code rather than shell scripts, may be utilize the aws-sdk for golang? That way we can have the flexibility of performing instance actions more seamlessly... have a separate module for that as part of the framework?

yes, in code we should use aws sdk, its pretty straightforward.

fmoral2:
> > > > I was wondering if there is anyway to reboot the instances using terraform
> > > 
> > > 
> > > i was thinking on that too, but since this is kind of part of a " execution or test flow " and not directly tied to infra configuration, think thats better have as a sh script or even a go aux function.
> > > what u guys think ?
> > 
> > 
> > hmm, that sounds reasonable... I am leaning more towards having a functionality in code rather than shell scripts, may be utilize the aws-sdk for golang? That way we can have the flexibility of performing instance actions more seamlessly... have a separate module for that as part of the framework?
> 
> yes, in code we should use aws sdk, its pretty straightforward.

what u think @est-suse ?

modules/install/rke2_master.sh
est-suse: add it back, thanks was a mistake

modules/install/rke2_master.sh
```diff
@@ -45,6 +47,19 @@ network_manager() {
       subscription-manager repos --enable=rhel-7-server-extras-rpms
   fi
 
+  if [[ "$node_os" = *"centos"* ]] || [[ "$node_os" = *"rhel"* ]] || [[ "$node_os" = *"oracle"* ]]
```
fmoral2, line 50: this should be a different function-  `disable_cloud_setup`

modules/rke2/worker/instances_worker.tf
```diff
@@ -38,6 +38,53 @@ resource "aws_instance" "worker" {
...
+      "sudo sed -i s/${local.master_ip_prev}/${local.master_ip}/g /etc/rancher/rke2/config.yaml",
+      "sudo sed -i s/-ip:.*/\"-ip: ${aws_eip.worker_with_eip[each.key].public_ip}\"/g /etc/rancher/rke2/config.yaml",
+      "sudo systemctl restart rke2-agent"
+    ]
+  }
+
+provisioner "local-exec" {
+    command    = "sed s/127.0.0.1/\"${local.master_ip}\"/g /tmp/${var.resource_name}_config >/tmp/${var.resource_name}_kubeconfig"
+    on_failure = continue
+  }
```
rancher-max, line 82: Does this work on the worker nodes? I don't think they should have any kubeconfig.

modules/rke2/worker/instances_worker.tf
```diff
@@ -47,11 +94,53 @@ locals {
...
+    command = "../install/rke2_stop_start_instance.sh start_s1_s2 ${var.resource_name}"
+  }
+}
+
+resource "null_resource" "start_master_worker" {
+  count = var.create_eip ? 1 : 0
+  depends_on = [null_resource.start_server1_server2]
+  provisioner "local-exec" {
+    command = "../install/rke2_stop_start_instance.sh start_master_worker ${var.resource_name}"
+  }
```
rancher-max, line 145: I believe this is what you're changing to use the aws sdk, but just in case it's not, I think we should create a testcase specifically that does this via the aws sdk. That way we can do the steps like:
1. Bring up the instances with the EIP
2. Deploy a few workloads and validate they are functional
3. Reboot the instances via stop and start like this (but via the aws sdk)
4. Confirm nodes and workloads still functional
5. Deploy a few more workloads and validate they are functional

Issue #25, opened 2023-09-23
delete resources script without path validation


fmoral2:
- Steps: 

1. Run ./scripts sh delete_resources.sh
2. When the script is not able to find the path to collect the resource it ended up loading just "Name=tag:Name,Values=${NAME_PREFIX}*" so resulting in getting and deleting all instances available 

Expected behavior:
- if the path is wrong, does not exist or wrong product it will break 

Actual behavior:
- It passes and allow a big security concern of security

fmoral2:
https://github.com/rancher/distros-test-framework/pull/26

PR #26, opened 2023-09-23
fix: delete resources plus general sec improvements

fmoral2:
#### Proposed Changes ####

**I am not adding new feat or functionality just fixing and improving.**

- fix: delete resources plus general sec improvements 
- fix: shell checks

- for now to a better Explicitness and compliance i have changed the `config.yaml file name to product.yaml`

- added an extra security validation for deletion, that the resource name used must have your host username in it

#### Types of Changes ####

- fix: delete resources plus general sec improvements 

 

#### Verification ####

- n/a
 

#### Linked Issues ####

https://github.com/rancher/distros-test-framework/issues/25


pkg/customflag/model.go
```diff
@@ -9,73 +9,73 @@ import (
...
+var TestCaseNameFlag stringSlice
 
 // FlagConfig is a type that wraps all the flags that can be used
 type FlagConfig struct {
-	InstallMode       InstallTypeValueFlag
-	TestConfig        TestConfigFlag
-	ClusterConfig     ClusterConfigFlag
-	SUCUpgradeVersion SUCUpgradeVersion
-	Channel           ChannelFlag
+	InstallMode       installTypeValueFlag
```
mdrahman-suse, line 16: Can we rename it to be consistent? `installModeFlag` instead of `installTypeValueFlag`?

factory/config.go
```diff
@@ -27,10 +27,10 @@ type Cluster struct {
 	NumWinAgents int
 	NumServers   int
 	NumAgents    int
-	Config       ClusterConfig
+	Config       clusterConfig
```
mdrahman-suse, line 30: is there any reason for

scripts/delete_resources.sh
```diff
@@ -1,8 +1,19 @@
...
+PRODUCT_NAME=$(grep ENV_PRODUCT <../config/product.yaml | cut -d: -f2 | tr -d ' "')
 RESOURCE_NAME=$(grep resource_name <../config/"$PRODUCT_NAME".tfvars | cut -d= -f2 | tr -d ' "')
+
+if [[ -z "$RESOURCE_NAME" ]]; then
+  echo "No file or directory found"
+  exit 1
+fi
+
+if  [[ "$RESOURCE_NAME" != "$PRODUCT_NAME".tfvars ]]; then
+  echo "Wrong tf.vars file found"
```
mdrahman-suse, line 13: nit: Fix typo - `Wrong tfvars file found` Also suggest to add the RESOURCE_NAME variable on failures/errors

mdrahman-suse (changes requested):
Few minor comments, otherwise LGTM

pkg/customflag/model.go
fmoral2: totally

rancher-max (commented):
I think if changing config.yaml name, it will also need to change in the Jenkinsfile

modules/install/optional_write_files.sh
```diff
@@ -7,7 +7,7 @@ files=$1
 
 if [ -n "$files" ]
 then
-  file_array=($(echo "$files" | tr ' ' '\n'))
+  read -r -a file_array <<< "$files"
```
rancher-max, line 10: I remember writing this a while back and getting shellchecker being upset with me for not doing it this way, but in certain OSes this didn't work. Have you confirmed this works at least in a critical subset of our supported OSes? (SLE Micro, SLES, RHEL, Ubuntu). Don't need to run the full automation, just the script itself should be fine.

Same comment with other shell file modifications, but I assume if it works in one then it'll work in the others.

scripts/delete_resources.sh
```diff
@@ -1,8 +1,24 @@
...
+  echo "No resource name found for: $PRODUCT_NAME.tfvars file"
+  exit 1
+fi
+
+if [[ "$RESOURCE_NAME" != *"$USER"* ]]; then
+  printf "Wrong tfvars file found for: %s\nThe resource name should have your username in it\nUser found: %s \
+  \nRESOURCE_NAME found: %s" "$PRODUCT_NAME" "$USER" "$RESOURCE_NAME"
+  exit 1
+fi
+
```
rancher-max, line 21: I'm a big fan of these changes overall! Thank you 🙌 

Just one question --  do we want `RESOURCE_NAME` to be so tied to `USER`? For example, my `USER=maxross`, but my `RESOURCE_NAME` is often just `max` or `maxr`

rancher-max (commented):
Good changes overall -- changing public variables to private when they don't need to be used outside of their defined scopes. Couple of questions and comments left on specific portions

modules/install/optional_write_files.sh
fmoral2: checking

scripts/delete_resources.sh
fmoral2: if the person plan to use this script somewhere or somehow in the future, i think we should "force" that yes
if u just use jenkins jobs for instance, no need to. 

and if so i can add that on the README task as well , to include this info

fmoral2: as per discussion now in standup , this is not a mandatory validaton, its just an extra layer of it .

we can delete that and all the validations necessary it will still be good and enough

mdrahman-suse: As I have mentioned, I am hesitant to add this check to restrict username to be used in the resource name as I am aware that no one really uses the entire username on resources.. also concerned on the security aspect of using username in resource allocation

fmoral2: @rancher-max  any more thoughts here? what u think ?

modules/install/optional_write_files.sh
VestigeJ: Do we want to have shellcheck results run on the repo itself? I know in my own scripts I have needed to add some ignore sections, because the shellcheck rule was breaking my working across all OS under test functionality.

scripts/delete_resources.sh
mdrahman-suse: Also I think we should create a separate job in Jenkins to delete resources as we have currently in rancher-ha folder and utilize this script to do similar task

rancher-max: I agree that it would be nice to have a separate job in Jenkins to delete resources as MD mentioned.

I do think that this extra layer of validation for USER should be deleted as it really limits us on naming, and people with longer names might even be prevented entirely from create loadbalancers since they have a 32 character limit IIRC. 

I like the idea of having another sanity check though. I think asking for user input could even be good here, since delete can be catastrophic. Something like:
```
printf "This is going to delete all AWS resources with the prefix %s. Continue (y/n)?" "$RESOURCE_NAME"
read  REPLY
if [[ $REPLY =~ ^[Yy]$ ]]
then
  # do delete logic
else
  # print something about not getting confirmation and exiting.
fi
```
Then to run in Jenkins, maybe have it run this script and automatically continue with `echo y | script.sh` or using `yes`.

modules/install/optional_write_files.sh
fmoral2: i think we need to evaluate each case/warning, sometimes really makes sense to add ignore , in other cases really makes sense to adapt or to comply

fmoral2: but regarding to this specific one, i am rolling back the changes, since it actually does not work well on some Os.s

scripts/delete_resources.sh
fmoral2: i have deleted this extra validation
and added this

rancher-max: LGTM on this - thank you!

VestigeJ (approved):

scripts/delete_resources.sh
```diff
@@ -1,110 +1,140 @@
...
+    STATUS_ID=$(echo "$CHANGE_STATUS" | jq -r '.ChangeInfo.Id')
+    #Get status from the change
+    aws route53 wait resource-record-sets-changed --id "$STATUS_ID"
+    echo "Successfully deleted Route53 record ${R53_RECORD}: status: ${STATUS_ID}"
+  else
+    echo "No Route53 record found"
+  fi
 else
-  echo "No Route53 record found"
+  echo "Exiting: User did not confirm"
```
mdrahman-suse, line 138: Nit: Actually user did confirm and selected No in this case. Not to look confusing but would be good if we mention that resources were not deleted and user has to delete them at their own discretion

modules/install/optional_write_files.sh
mdrahman-suse: Is the change reverted?

scripts/delete_resources.sh
fmoral2: totally, good cacth

mdrahman-suse (approved):
LGTM

modules/install/optional_write_files.sh
rancher-max: I think still need to rollback these changes

rancher-max (changes requested):
Just rolling back changes to optional_write_files.sh array and then it LGTM

rancher-max (approved):

mdrahman-suse (approved):
LGTM, Don't forget to squash and merge

PR #32, opened 2023-09-28
 feat: add test runner and makefile adjustments

fmoral2:
#### Proposed Changes ####

-feat: add test runner and makefile adjustments

The proposal is to add a test-runner to run make file local and in docker

Bellow you find the related tasks

#### Types of Changes ####
- Yes README and examples updated

#### Testing ####

Checklist:

- YES 

   UpgradeSUC - https://jenkins.int.rancher.io/job/rke2-tests/view/rke2-patch-validation/job/rke2_validate_cluster/383/console

   Windows -https://jenkins.int.rancher.io/job/rke2-tests/view/rke2-patch-validation/job/rke2_validate_cluster/385/console

- ok 
Screenshot 2023-09-28 at 18 20 51

- ok

- ok 

- ok 
- ok 

#### Linked Issues ####
- https://github.com/rancher/qa-tasks/issues/802
- https://github.com/rancher/qa-tasks/issues/854

#### Further Comments ####

- Variables precedence order 

 First: code
 Second: CLI
 Third: .env


Makefile
```diff
@@ -1,83 +1,183 @@
...
 	@docker logs -f acceptance-test${IMGNAME}
 
 .PHONY: test-env-down
 test-env-down:
-	@echo "Removing containers and images"
-	@docker stop $$(docker ps -a -q --filter="name=acceptance-test*")
-	@docker rm $$(docker ps -a -q --filter="name=acceptance-test*") ; \
-	 docker rmi --force $$(docker images -q --filter="reference=acceptance-test*")
+	@echo "Removing containers"
+	@docker ps -a -q --filter="name=acceptance-test*" | xargs -r docker rm -f || true
```
mdrahman-suse, line 54: Can we add a check to not error out if the container does not exists?

scripts/build.sh
```diff
@@ -14,7 +14,7 @@ fi
 
 count=0
 while [[ 3 -gt $count ]]; do
-    docker build . -f scripts/Dockerfile.build -t acceptance-tests-"${TRIM_JOB_NAME}""${BUILD_NUMBER}"
+    docker build . -f scripts/Dockerfile.jenkins -t acceptance-tests-"${TRIM_JOB_NAME}""${BUILD_NUMBER}"
```
mdrahman-suse, line 17: Is there any reason for changing the extension?

Makefile
fmoral2: sure

scripts/build.sh
fmoral2: actually i have added a new dockerfile , once its actually a different build environment , instead of playing with conditionals and env vars, i created a dockerfike per env, one for "local "builds and another one for jenkins

Makefile
```diff
@@ -1,83 +1,190 @@
...
-            -expectedValueUpgrade "${VALUEUPGRADED}" -installVersionOrCommit "${INSTALLTYPE}" -channel "${CHANNEL}" -testCase "${TESTCASE}" \
-            -deployWorkload "${DEPLOYWORKLOAD}" -workloadName "${WORKLOADNAME}" -description "${DESCRIPTION}"; \
-		elif [ "${TESTDIR}" = "mixedoscluster" ]; then \
-            go test -timeout=45m -v -tags=mixedos ./mixedoscluster/...; \
-        fi; \
-    elif [ -z "${TESTDIR}" ]; then \
-        go test -timeout=45m -v ./createcluster/...; \
-    fi;'
+	@if [ -z "$${IMGNAME}" ]; then IMGNAME=${IMGNAME}; fi; \
+	if [ -z "$${TAGNAME}" ]; then TAGNAME=${TAGNAME}; fi; \
```
ShylajaDevadiga, line 11: Is this duplicate line?

Makefile
```diff
@@ -1,83 +1,190 @@
...
+	  docker run -dt --name acceptance-test-$${IMGNAME} \
```
ShylajaDevadiga, line 12: Is this type or do we need literal $ here?

Makefile
```diff
@@ -1,83 +1,190 @@
...
+	  -e AWS_ACCESS_KEY_ID=$${AWS_ACCESS_KEY_ID} \
+	  -e AWS_SECRET_ACCESS_KEY=$${AWS_SECRET_ACCESS_KEY} \
+	  --env-file ./config/.env \
+	  -v ${ACCESS_KEY_LOCAL}:/go/src/github.com/rancher/distros-test-framework/config/.ssh/aws_key.pem \
+	  -v ./scripts/test-runner.sh:/go/src/github.com/rancher/distros-test-framework/scripts/test-runner.sh \
+	  acceptance-test-${TAGNAME} && \
+	  make test-logs USE=IMGNAME acceptance-test-${IMGNAME}
```
ShylajaDevadiga, line 19: Default value to be added

Makefile
```diff
@@ -1,83 +1,190 @@
...
+	CONTAINER_ID=$(shell docker ps -a -q --filter ancestor=acceptance-test-${TAGNAME} | head -n 1); \
+    	if [ -z "$$CONTAINER_ID" ]; then \
+    		echo "No matching container found."; \
+    		exit 1; \
+    	else \
+    		docker commit $$CONTAINER_ID teststate:latest; \
+    		if [ $$DOCKERCOMMIT -eq 0 ]; then \
+    			docker run -dt --name acceptance-test-${TESTSTATE} --env-file ./config/.env \
+    			-e AWS_ACCESS_KEY_ID=$${AWS_ACCESS_KEY_ID} \
+    			-e AWS_SECRET_ACCESS_KEY=$${AWS_SECRET_ACCESS_KEY} \
```
ShylajaDevadiga, line 32: Can we standardize the way the variables are accessed? Example $$var ${Var} $Var Based on the right thing to use

Makefile
fmoral2: typo, fixed

Makefile
fmoral2: added tks

Makefile
fmoral2: sure

Makefile
```diff
@@ -1,83 +1,190 @@
...
+    			-v $${ACCESS_KEY_LOCAL}:/go/src/github.com/rancher/distros-test-framework/config/.ssh/aws_key.pem \
+    			-v ./scripts/test-runner.sh:/go/src/github.com/rancher/distros-test-framework/scripts/test-runner.sh \
+    			teststate:latest && \
+    			 make test-logs USE=TESTSTATE acceptance-test-${TESTSTATE} \
```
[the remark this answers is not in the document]
fmoral2: those are not VAR, they are string values being checkd here--
```
test-logs:
	@if [ "${USE}" = "IMGNAME" ]; then \
		docker logs -f acceptance-test-${IMGNAME}; \
	elif [ "${USE}" = "TESTSTATE" ]; then \
		docker logs -f acceptance-test-${TESTSTATE}; \
	fi;

```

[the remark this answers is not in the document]
fmoral2: i am checking the string to them be able to use the correct logs then with the var ${IMGNAME} or  ${TESTSTATE}

entrypoint/createcluster/createcluster_suite_test.go
```diff
@@ -20,7 +20,7 @@ func TestMain(m *testing.M) {
 	flag.Var(&customflag.ServiceFlag.ClusterConfig.Destroy, "destroy", "Destroy cluster after test")
 	flag.Parse()
 
-	cfg, err = config.AddConfigEnv("../../config")
+	cfg, err = config.AddConfigEnv("../../config/.env")
```
mdrahman-suse, line 23: Question: Is there a way to get rid of this `../../` and have relative or absolute path derived? I remember there was a BasePath but not sure if that can be used here...

mdrahman-suse: Also in some places I see we are using `./config/.env` as well. So if we can have the `.env` path saved in shared, I think that will be great

pkg/template/helper.go
```diff
@@ -60,21 +60,23 @@ func executeTestCombination(v VersionTestTemplate) error {
...
 
-	testCase := map[string]testCase{
-		"TestDaemonset":                   testcase.TestDaemonset,
-		"TestIngress":                     testcase.TestIngress,
-		"TestDnsAccess":                   testcase.TestDnsAccess,
-		"TestServiceClusterIP":            testcase.TestServiceClusterIp,
-		"TestServiceNodePort":             testcase.TestServiceNodePort,
-		"TestLocalPathProvisionerStorage": testcase.TestLocalPathProvisionerStorage,
-		"TestServiceLoadBalancer":         testcase.TestServiceLoadBalancer,
+	tcs := map[string]testCase{
```
mdrahman-suse, line 63: Nit: Instead of `tcs` may be rename to `testCaseMap`?

scripts/Dockerfile.build
```diff
@@ -27,3 +28,11 @@ RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/i
 WORKDIR $GOPATH/src/github.com/rancher/distros-test-framework
 
 COPY . .
+
+RUN go mod download
```
mdrahman-suse, line 32: Thanks! 👍

scripts/Dockerfile.jenkins
mdrahman-suse, line 1: Not sure I understand the need to having separate dockerfiles

scripts/docker_stats.sh
mdrahman-suse, line 1: Where will this file be used?

entrypoint/createcluster/createcluster_suite_test.go
fmoral2: sure i can create some function do handle.
base path will not be enough

pkg/template/helper.go
fmoral2: since this var is not used out of the current scope and u are creating a map, no need to add "map " in the name, in this case we can keep short and concise

scripts/docker_stats.sh
fmoral2: its going to be used on the makefile to monitor container usage of resources.
its saving on a /tmp file.

config/reader.go
```diff
@@ -31,28 +30,54 @@ func AddConfigEnv(path string) (*ProductConfig, error) {
...
+	if err = setEnv(fullPath); err != nil {
 		return nil, err
+	}
+
+	config = &ProductConfig{}
+	config.TFVars = os.Getenv("ENV_TFVARS")
+	config.Product = os.Getenv("ENV_PRODUCT")
 
+	if config.TFVars == "" || (config.TFVars != "k3s.tfvars" && config.TFVars != "rke2.tfvars") {
+		fmt.Printf("unknown tfvars: %s\n", config.TFVars)
```
mdrahman-suse, line 43: Do we want to use logger instead of fmt here?

docs/examples/.env.example
```diff
@@ -0,0 +1,51 @@
+# Framework config variables
+ENV_PRODUCT="{{PRODUCT}}"
+ENV_TFVARS="{{PRODUCT}}".tfvars
+# Note: PRODUCT can be k3s or rke2
+
+
+#ACCESS_KEY_LOCAL="˜/jenkins-rke-validation.pem"
+ACCESS_KEY_LOCAL="~/jenkins-elliptic-validation.pem"
```
mdrahman-suse, line 8: Redact key name from here?

entrypoint/createcluster/createcluster_suite_test.go
```diff
@@ -20,7 +22,12 @@ func TestMain(m *testing.M) {
 	flag.Var(&customflag.ServiceFlag.ClusterConfig.Destroy, "destroy", "Destroy cluster after test")
 	flag.Parse()
 
-	cfg, err = config.AddConfigEnv("../../config")
+	configPath, err := shared.EnvDir("entrypoint")
+	if err != nil {
+		return
+	}
+	fmt.Println("configPath: ", configPath)
```
mdrahman-suse, line 29: https://github.com/rancher/distros-test-framework/pull/32/files#r1352755686

pkg/testcase/localpathstorage.go
```diff
@@ -30,11 +30,12 @@ func TestLocalPathProvisionerStorage(deleteWorkload bool) {
 	Eventually(func(g Gomega) {
 		var res string
 		fmt.Println("Writing and reading data from pod")
```
mdrahman-suse, line 32: Do we need this Print statement? If yes can we update with the proper actions? It looks like it is just reading data from pod...

scripts/Dockerfile.build
```diff
@@ -1,11 +1,12 @@
 FROM golang:alpine
 
-ARG TF_VERSION=1.4.5
+ARG TF_VERSION=1.5.7
```
mdrahman-suse, line 3: If we are upgrading TF version, I think we should run full regression on existing tests

mdrahman-suse:
Also I see some lint warnings from GH Actions, do we need to address those?

scripts/Dockerfile.build
fmoral2: already done, and this is only for local.

config/reader.go
fmoral2: to use logger here we will need a refactor on the calls, which is not the case of this PR

config/reader.go
```diff
@@ -31,28 +30,54 @@ func AddConfigEnv(path string) (*ProductConfig, error) {
...
-	if config.TFVars == "" || (config.TFVars != "k3s.tfvars" && config.TFVars != "rke2.tfvars") {
-		fmt.Printf("unknown tfvars: %s\n", config.TFVars)
-		os.Exit(1)
+	return config, nil
+}
+
+func setEnv(fullPath string) error {
+	file, err := os.Open(fullPath)
+	if err != nil {
+		fmt.Println("failed to open file:", err)
```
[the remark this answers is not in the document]
fmoral2: to use logger here we will need a refactor on the calls, which is not the case of this PR

entrypoint/createcluster/createcluster_suite_test.go
fmoral2: this should not be here
tks

fmoral2:
> Also I see some lint warnings from GH Actions, do we need to address those?

warnings no, only errors. 
 warnings its not a must to. Its only if its feasible and makes real sense

mdrahman-suse (dismissed):
LGTM

modules/install/k3s_master.sh
```diff
@@ -124,7 +124,7 @@ wait_nodes() {
   export PATH="$PATH:/usr/local/bin"
   timeElapsed=0
 
-  while (( timeElapsed < 300 )); do
+  while (( timeElapsed < 1200 )); do
```
rancher-max, line 127: Which use case does increasing the time here by 4x resolve?

shared/cluster.go
```diff
@@ -184,14 +188,14 @@ func kubectlCmdOnNode(cmd string) (string, error) {
 }
 
 // FetchClusterIP returns the cluster IP and port of the service.
-func FetchClusterIP(namespace string, serviceName string) (string, string, error) {
-	ip, err := RunCommandHost("kubectl get svc " + serviceName + " -n " + namespace +
+func FetchClusterIP(namespace string, serviceName string) (ip, port string, err error) {
```
rancher-max, line 191: Should be consistent here and define the type for `ip` also:
`func FetchClusterIP(namespace string, serviceName string) (ip string, port string, err error) {`

rancher-max (commented):
This looks overall fine. Slight concern that we're adding a maintenance burden in that now we can run this via:
- `go test`
- `make`
- `docker build...` (local)
- `docker build...` (jenkins) 
where there can be different files that go into each of them. So now each time a new test is added, there need to be four separate run configurations to attempt just for one test, let alone any regression checks.

We should standardize on a common way to run the code. It requires docker for Jenkins, so that should definitely be one way, but if we want to move forward using environment vars/files then we should also do that in Jenkins so that we have to maintain less files and change complexity overall.

modules/install/k3s_master.sh
fmoral2: A lot of times  in docker and also in Jenkins we have latency that is not allowing the wait and timing out.
We can decrease a bit if you find needed # but less than 600 no.

shared/cluster.go
fmoral2: This is a common standard and good practice in golang. If you 2 params is the same type you declare it once

fmoral2:
> This looks overall fine. Slight concern that we're adding a maintenance burden in that now we can run this via:
> 
>     * `go test`
> 
>     * `make`
> 
>     * `docker build...` (local)
> 
>     * `docker build...` (jenkins)
>       where there can be different files that go into each of them. So now each time a new test is added, there need to be four separate run configurations to attempt just for one test, let alone any regression checks.
> 
> 
> We should standardize on a common way to run the code. It requires docker for Jenkins, so that should definitely be one way, but if we want to move forward using environment vars/files then we should also do that in Jenkins so that we have to maintain less files and change complexity overall.


> * So now each time a new test is added, there need to be four separate run configurations to attempt just for one test, let alone any regression checks.








i totally get your point but you are kind of mixing up some different subjects and also making not too correct  assumptions 

1- Related to how to run the code , we changed nothing. We only 3 ways of running as before , via local , via docker or via Jenkins

2- The changes was basic 2 : 
  a) simplifying the env config 
  b) adding a new docker file to avoid start having a docker file with a lot of conditionals. 

   b) this is a normal approach when you have different build targets , think as this as something like you have an application , that u need to build your image with different params and configs ( which is totally normal and necessary ) so its a common way to better organize, having docker and config files separated like: 
build dev env - docker compose with different params due to differences of needs and configurations environment 
build production - same stuff 

summarizing , this is not bringing additional burden in a way thats bad, this is doing the opposite , its starting to organize better the framework.

We need to start to understand that the idea here is to have a framework, not a scripting automation. 
Saying that the framework will act as a app or service or lib. ( whatever path is better ) , so in that way we need to know that this "app" needs features, functionalities , to start to grow and do lots of nice and handy stuff, so its impossible to an app grow and add features, not adding any burden, not adding any complexity. The case here is to be aware of try to always maintain things simple ( as possible ) , sometimes we cant run away of having some complexity or some burden or also some duplication even ( things that we are tired to listen that is bad and we should avoid) yes of course we should avoid, avoid when possible and not adding extra complexity or over engineering stuff , but this is different than add new features or functionalities.

Again we need to change the perspective here, this is not a scripting automation, its a framework being built. 

So now each time a new test is added, there need to be four separate run configurations to attempt just for one test, let alone any regression checks.

i cant see this happening , because when u add a new test case u should just do whatever you were already doing, adding the make command for that and creating the job on jenkins. No need for any of this.  Maybe u are misunderstanding what make file is.


We should standardize on a common way to run the code

thats exactly what we are doing here, standardising, separating what needs to be separate, and create a common way to run , via go test , make or jenkins.



but if we want to move forward using environment vars/files then we should also do that in Jenkins so that we have to maintain less files and change complexity overall.

this for instance its a over engineering decision , we dont need that complexity in jenkins for now, and probably never.
you should handle this in the framework config, jenkins dont need to know those internal details

shared/cluster.go
rancher-max: Ah good to know! Learn something new every day 😄

modules/install/k3s_master.sh
rancher-max: No need to decrease! In some cases it might cause for longer delay when things don't come up properly, but I think that's okay to account for the positive scenario

rancher-max:
> 1- Related to how to run the code , we changed nothing. We only 3 ways of running as before , via local , via docker or via Jenkins

So `via local` here can mean running either `go test`, `make` or `docker build... && docker run...`. Jenkins just runs via the docker flow. I wasn't super clear in my point there though. I meant to say just that now there are two Dockerfiles to maintain -- one for when we run `docker build... && docker run...` locally and the other for when that is run in Jenkins. 

You did touch on that a bit later on:
> b) this is a normal approach when you have different build targets , think as this as something like you have an application , that u need to build your image with different params and configs ( which is totally normal and necessary ) so its a common way to better organize, having docker and config files separated like:
> - build dev env - docker compose with different params due to differences of needs and configurations environment 
> - build production - same stuff diff args and also sometimes different strategies ( which is also our case on docker files  one docker file has different approach to use entrypoint bc locally its needed. ) 

> - summarizing , this is not bringing additional burden in a way thats bad, this is doing the opposite , its starting to organize better the framework.

I'd argue here that we shouldn't have different build targets other than different operating systems where this might run, which should be accounted for by having the docker flow already. 

It seems to me the biggest point of this change is to simplify how the tests are run, which I do approve of. I think we can allow that simplicity to be present in our local docker runs as well as the jenkins docker runs. Because technically here by having two dockerfiles, I can actually run it locally via docker using today's flow, just building from the jenkins dockerfile.


> but if we want to move forward using environment vars/files then we should also do that in Jenkins so that we have to maintain less files and change complexity overall.
>  > this for instance its a over engineering decision , we dont need that complexity in jenkins for now, and probably never.
you should handle this in the framework config, jenkins dont need to know those internal details

Jenkins is simply a CI tool that runs things, so if it doesn't need additional complexity, then we really shouldn't need that locally either. As I see it though, this PR actually makes things less complex, so I'd argue it should be present in Jenkins the same as locally.

scripts/test_runner.sh
```diff
@@ -0,0 +1,69 @@
...
+EXPECTEDVALUE="${EXPECTEDVALUE}"
+VALUEUPGRADED="${VALUEUPGRADED}"
+CHANNEL="${CHANNEL}"
+INSTALLVERSIONORCOMMIT="${INSTALLVERSIONORCOMMIT}"
+SUCUPGRADEVERSION="${SUCUPGRADEVERSION}"
+TESTCASE="${TESTCASE}"
+WORKLOADNAME="${WORKLOADNAME}"
+DESCRIPTION="${DESCRIPTION}"
+DEPLOYWORKLOAD="${DEPLOYWORKLOAD}"
+SONOBUOYVERSION="${SONOBUOYVERSION}"
```
rancher-max, line 17: Do we need to redefine all of these variables?

scripts/test_runner.sh
```diff
@@ -0,0 +1,69 @@
...
+   exit 1
+fi
+
+
+case "$TESTDIR" in
+     upgradecluster|versionbump|mixedoscluster|dualstack|createcluster)
+      printf "\n\nRunning tests for %s\n\n" "${TESTDIR} on ${ENV_PRODUCT}"
+        ;;
+    *)
+        printf "\n\nTESTDIR is not a go test package\n\n"
```
rancher-max, line 35: Here should print the value of the TESTDIR variable instead of just the string "TESTDIR" itself

fmoral2: Tk you. Yeah

fmoral2:
> > 1- Related to how to run the code , we changed nothing. We only 3 ways of running as before , via local , via docker or via Jenkins
> 
> So `via local` here can mean running either `go test`, `make` or `docker build... && docker run...`. Jenkins just runs via the docker flow. I wasn't super clear in my point there though. I meant to say just that now there are two Dockerfiles to maintain -- one for when we run `docker build... && docker run...` locally and the other for when that is run in Jenkins.
> 
> You did touch on that a bit later on:
> 
> > b) this is a normal approach when you have different build targets , think as this as something like you have an application , that u need to build your image with different params and configs ( which is totally normal and necessary ) so its a common way to better organize, having docker and config files separated like:
> > 
> > * build dev env - docker compose with different params due to differences of needs and configurations environment
> > * build production - same stuff diff args and also sometimes different strategies ( which is also our case on docker files  one docker file has different approach to use entrypoint bc locally its needed. )
> 
> > * summarizing , this is not bringing additional burden in a way thats bad, this is doing the opposite , its starting to organize better the framework.
> 
> I'd argue here that we shouldn't have different build targets other than different operating systems where this might run, which should be accounted for by having the docker flow already.
> 
> It seems to me the biggest point of this change is to simplify how the tests are run, which I do approve of. I think we can allow that simplicity to be present in our local docker runs as well as the jenkins docker runs. Because technically here by having two dockerfiles, I can actually run it locally via docker using today's flow, just building from the jenkins dockerfile.
> 
> > but if we want to move forward using environment vars/files then we should also do that in Jenkins so that we have to maintain less files and change complexity overall.
> > > this for instance its a over engineering decision , we dont need that complexity in jenkins for now, and probably never.
> > > you should handle this in the framework config, jenkins dont need to know those internal details
> 
> Jenkins is simply a CI tool that runs things, so if it doesn't need additional complexity, then we really shouldn't need that locally either. As I see it though, this PR actually makes things less complex, so I'd argue it should be present in Jenkins the same as locally.

Gotta you yeah.


But the point is that to run locally now we have an extra feature that I don't see a need in Jenkins. The override stuff.

Besides that feature, all of the changes were applied on Jenkins file as well.

scripts/test_runner.sh
fmoral2: Not mandatory. I was just checking if has value in it and "documenting" in the beginning.  Take it out?

scripts/test_runner.sh
fmoral2: Actually no 

Since I was reassign, it is correct

fmoral2: But If I will remove the reasgin I will refactor everything

rancher-max: Wait shouldn't it just be: `printf "\n\n%s is not a go test package\n\n" "${TESTDIR}"` kind of like how it is above?

fmoral2: yes. sorry, forget what i said . was to tired lol

Makefile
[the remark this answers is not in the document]
VestigeJ: We should standardize multi-word variables akin to AWS_SECRET_ACCESS_KEY so IMGNAME would be IMG_NAME and TEST_STATE this improves readability and maintainability

Makefile
```diff
@@ -1,83 +1,190 @@
...
 .PHONY: test-upgrade-suc
 test-upgrade-suc:
-	@go test -timeout=45m -v -tags=upgradesuc  ./entrypoint/upgradecluster/... -upgradeVersion ${UPGRADEVERSION}
+	@go test -timeout=45m -v -tags=upgradesuc -count=1 ./entrypoint/upgradecluster/... -sucUpgradeVersion ${SUCUPGRADEVERSION}
+
 
 .PHONY: test-upgrade-manual
 test-upgrade-manual:
-	@go test -timeout=45m -v -tags=upgrademanual ./entrypoint/upgradecluster/... -installVersionOrCommit ${INSTALLTYPE}
+	@go test -timeout=45m -v -tags=upgrademanual -count=1 ./entrypoint/upgradecluster/... -installVersionOrCommit ${INSTALLVERSIONORCOMMIT} -channel ${CHANNEL}
```
VestigeJ, line 93: INSTALL_VERSION_OR_COMMIT

Makefile
```diff
@@ -1,83 +1,190 @@
...
-	  -cmd "${CMD}" \
-	  -expectedValue ${VALUE} \
-	  -expectedValueUpgrade ${VALUEUPGRADED} \
-	  -installVersionOrCommit ${INSTALLTYPE} -channel ${CHANNEL} \
-	  -testCase "${TESTCASE}" -deployWorkload ${DEPLOYWORKLOAD} -workloadName ${WORKLOADNAME} -description "${DESCRIPTION}"
 
 .PHONY: test-create-mixedos
 test-create-mixedos:
-	@go test -timeout=45m -v ./entrypoint/mixedoscluster/...
+	@go test -timeout=45m -v -count=1 ./entrypoint/mixedoscluster/... $(if ${SONOBUOYVERSION},-sonobuoyVersion ${SONOBUOYVERSION})
```
VestigeJ, line 98: SONOBUOY_VERSION

Makefile
```diff
@@ -1,83 +1,190 @@
...
+.PHONY: test-create-dualstack
+test-create-dualstack:
+	@go test -timeout=45m -v -count=1 ./entrypoint/dualstack/...
 
+.PHONY: test-version-bump
+test-version-bump:
+	@go test -timeout=45m -v -count=1 ./entrypoint/versionbump/... -tags=versionbump \
+	-cmd "${CMD}" \
+    -expectedValue ${EXPECTEDVALUE} \
+    $(if ${VALUEUPGRADED},-expectedValueUpgrade ${VALUEUPGRADED}) \
```
VestigeJ, line 109: EXPECTED_VALUE
VALUE_UPGRADED
INSTALL_VERSION_OR_COMMIT
TEST_CASE
WORKLOAD_NAME
DEPLOY_WORKLOAD
et al

docs/examples/.env.example
```diff
@@ -0,0 +1,51 @@
...
+# Note: PRODUCT can be k3s or rke2
+
+
+#ACCESS_KEY_LOCAL="˜/aws-key.pem"
+ACCESS_KEY_LOCAL="~/aws-key.pem"
+
+
+
+######  Test runner config variables  ########
+IMGNAME=rke2ewedd3
```
VestigeJ, line 13: IMG_NAME=
TAG_NAME=
TEST_DIR=
TEST_STATE=
TEST_TAG=

EXPECTED_VALUE=
VALUE_UPGRADED=
INSTALL_VERSION_OR_COMMIT=
SUC_UPGRADE_VERSION=
TEST_CASE=
WORKLOAD_NAME=
DEPLOY_WORKLOAD=

modules/install/rke2_master.sh
```diff
@@ -96,7 +96,7 @@ sudo systemctl enable rke2-server
 sudo systemctl start rke2-server
```
VestigeJ, line 97: We could potentially get more status output in the log and better test efficiency by using something more like 

```
while ! systemctl is-active rke2-server; do sleep 2; done
```

modules/install/k3s_master.sh
VestigeJ: We could potentially get more status output in the log and better test efficiency by using something more like 

```
while ! systemctl is-active k3s; do sleep 2; done
```

scripts/test_runner.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+
+ENV_PRODUCT="${ENV_PRODUCT}"
+ENV_TFVARS="${ENV_TFVARS}"
+TESTDIR="${TESTDIR}"
```
VestigeJ, line 5: Again I vote we expand all multi-word variables 
TEST_DIR="${TEST_DIR}"
 and so on

Makefile
fmoral2: prefer that too

done

please do a overhaul with me on those vars, bc its easy to mess up and let some pass

modules/install/rke2_master.sh
fmoral2: this lets wait a PR that is related to that to change , since its going to need more testing.

Makefile
fmoral2: @VestigeJ

fmoral2:
> Yeah I'm not too concerned with how it is, it just needs to be taken into consideration that there are subtle differences, so post-development when testing new changes we just need to ensure that it builds properly in the separate targets is all. We don't have to make additional changes


Hey, actually the difference between dockerfiles is not on the building , its just on the running itself.
So your worry will not need to be take into account, because all tests created have no reason to change the way it needs to be created based on the docker file changes.

So no need to have a double check on that, and either way we are having as PR requirement that the tests should be ran ok locally and on jenkins.

VestigeJ (dismissed):
Nice ✔️

rancher-max (approved):

PR #33, opened 2023-10-02
refactor: improve validate functions to be dynamic 

fmoral2:
#### Proposed Changes ####

- improve validate function to enable consumer decide which value to use on timeout and tickers

#### Types of Changes ####

- Refactor on validate function

#### Testing ####

Checklist:
- n/a

- ok 

  -n/a

  -n/a

- ok

- n/a 
#### Linked Issues ####

- https://github.com/rancher/qa-tasks/issues/964


pkg/assert/validate.go
```diff
@@ -80,18 +86,35 @@ func runAssertion(
 // ValidateOnHost runs an exec function on RunCommandHost and assert given is fulfilled.
 // The last argument should be the assertion.
 // Need to send kubeconfig file.
-func ValidateOnHost(args ...string) error {
+func ValidateOnHost(opt AsyncOpt, args ...string) error {
```
mdrahman-suse, line 89: So now every time we call this func, we need to specify the AsyncOpt struct? Can we have a default AsyncOpt so that we don't have to always set it? Also how is it different from Eventually func in Gomega?

pkg/assert/validate.go
```diff
@@ -80,18 +86,35 @@ func runAssertion(
...
 	exec := func(cmd string) (string, error) {
 		return shared.RunCommandOnNode(cmd, ip)
 	}
-	return validate(exec, args...)
+	return validate(exec, opt, args...)
+}
+
+// setAsyncTimer sets the timeout and ticker for the async validation
+func setAsyncTimer(opt AsyncOpt) (timeout <-chan time.Time, ticker *time.Ticker) {
+	if opt.Timeout == nil {
```
fmoral2, line 107: .

pkg/assert/validate.go
fmoral2: hey tks for commenting 
its already set a default value -
https://github.com/rancher/distros-test-framework/pull/33/files#r1344533170 

so you can pass a nil reference to it or customize to your own need 


> Also how is it different from Eventually func in Gomega?

- from gomega 
```
func (g *Gomega) makeAsyncAssertion(asyncAssertionType AsyncAssertionType, offset int, actualOrCtx interface{}, args ...interface{}) types.AsyncAssertion {
	baseOffset := 3
	timeoutInterval := -time.Duration(1)
	pollingInterval := -time.Duration(1)
	intervals := []interface{}{}
	var ctx context.Context

	actual := actualOrCtx
	startingIndex := 0
	if _, isCtx := actualOrCtx.(context.Context); isCtx && len(args) > 0 {
		// the first argument is a context, we should accept it as the context _only if_ it is **not** the only argumnent **and** the second argument is not a parseable duration
		// this is due to an unfortunate ambiguity in early version of Gomega in which multi-type durations are allowed after the actual
		if _, err := toDuration(args[0]); err != nil {
			ctx = actualOrCtx.(context.Context)
			actual = args[0]
			startingIndex = 1
		}
	}

	for _, arg := range args[startingIndex:] {
		switch v := arg.(type) {
		case context.Context:
			ctx = v
		default:
			intervals = append(intervals, arg)
		}
	}
	var err error
	if len(intervals) > 0 {
		timeoutInterval, err = toDuration(intervals[0])
		if err != nil {
			g.Fail(err.Error(), offset+baseOffset)
		}
	}
	if len(intervals) > 1 {
		pollingInterval, err = toDuration(intervals[1])
		if err != nil {
			g.Fail(err.Error(), offset+baseOffset)
		}
	}

	return NewAsyncAssertion(asyncAssertionType, actual, g, timeoutInterval, pollingInterval, 1, ctx, offset)
}
```

- so the difference is only on the implementation itself ( above using interface types and doing type assertions )  our is not using interfaces its using concrete types - 
```
type AsyncOpt struct {
	Timeout *time.Duration
	Ticker  *time.Ticker
} 
```
and our functions has a lot of less calls to perform.

- This validate functions we already had, its not new.  I am just improving it.

mdrahman-suse: Well, you are changing the method signature and now there is this additional param we have pass in even if we may not need to pass it all the time. I can understand you are improving the existing feature but is it really needed is my question.

fmoral2: its an improvement totally needed, because not everybody who calls this functions has the same need of wait a fixed amount of time.

fmoral2: > we have pass in even if we may not need to pass it all the time. 
@mdrahman-suse 
if you call this fucntion it means you need.
the only point here is that the function was doing it for you, not allowing you to customize it.

mdrahman-suse: Understood, although I am still hesitant on this change and we should try and make things simpler... If I understand it correctly, the right way would be to pass in a nil reference on the method signature
- like this?
```
assert.ValidateOnHost(nil, args...)
```
- OR I have to use the method like this?
```
assert.ValidateOnHost(assert.AsyncOpt{Timeout: nil, Ticker: nil}, args....)
```

mdrahman-suse: I would like to go with the former approach

pkg/assert/validate.go
```diff
@@ -50,20 +55,21 @@ func runAssertion(
 	errorsChan chan<- error,
 ) error {
 	for {
+		res, err := exec(cmd)
+		if err != nil {
+			errorsChan <- err
+			return fmt.Errorf("error from runCmd: %s\n %s", res, err)
```
mdrahman-suse, line 61: Can we clarify and log which cmd causes this error?

pkg/template/processor.go
```diff
@@ -110,6 +111,10 @@ func processOnNode(resultChan chan error, ip, cmd, expectedValue string) {
 	cmds := strings.Split(cmd, ",")
 	for _, c := range cmds {
 		err = assert.ValidateOnNode(
+			assert.AsyncOpt{
```
mdrahman-suse, line 114: also this implementation is now calling objects from various dependencies, which can cause confusion. For ex:
- AsyncOpt is called from assert
- Timeout is a value from shared.PointerDuration
- Ticker is from time
Previous implementation was simpler, just using time library

pkg/assert/validate.go
fmoral2: u should use like this 
```
 	err = assert.ValidateOnHost(
		assert.AsyncOpt{
			Timeout: shared.PointerDuration(420 * time.Second),
			Ticker:  time.NewTicker(5 * time.Second),
		},

```

OR
```
	err = assert.ValidateOnHost(
		assert.AsyncOpt{},
              .....
```

So the point here , you are using the same stuff when u use Eventually.
This is the actual signature for Eventually
`func (g *Gomega) makeAsyncAssertion(asyncAssertionType AsyncAssertionType, offset int, actualOrCtx interface{}, args ...interface{}) types.AsyncAssertion {`


but instead of passing a concrete type as we are passing ( a struct ) in the eventually u pass  as a string
```
........

	}, "1200s", "20s").Should(Succeed())


```


both ways has this pros and cons, TBH if we go to the approach above , THEN its going to be much more complex, because you will need to treat the received interfaces doing type assertions.

fmoral2: also on the implementation using concrete type is much more explicity and clear the usage
```
assert.AsyncOpt{
			Timeout: shared.PointerDuration(420 * time.Second),
			Ticker:  time.NewTicker(5 * time.Second),
		},

```

because u are actually seeing what u are passing  and what type is what, which u cant on the interfaces

pkg/assert/validate.go
fmoral2: sure

pkg/template/processor.go
fmoral2: yeah, but previous implementation was hard coded, and we are using pkgs, not a single source, so we need to call stuff from their respectively place

pkg/assert/validate.go
ShylajaDevadiga, line 89: Do we need this?

ShylajaDevadiga (commented):
I would like to keep it simple unless there is a need to add functions to keep time within limits

pkg/assert/validate.go
fmoral2: We need a way when call this function to pass a timeout customized because not every caller needs the same amount of time. For instance we have functions on the processor code that needs really small times, and other places that needs more ( like 420 s that today is hard-coded)

PR #35, opened 2023-10-04
improving functions name

fmoral2:
#### Proposed Changes ####

- The idea is to standardize some names across framework with more idiomatic golang patterns ( shorter and concise naming , also avoiding at most stuff like (gets and set ) have only left get when its totally appropriated like one the get nodes and get pods and adding  more meaningful names 

#### Testing ####

Checklist:


entrypoint/upgradecluster/upgrademanual_test.go
```diff
@@ -70,7 +70,7 @@ var _ = Describe("Test:", func() {
 	It("Checks Node Status pos upgrade and validate version", func() {
```
mdrahman-suse, line 70: typo: post

factory/cluster.go
```diff
@@ -11,13 +11,13 @@ import (
...
-func AddCluster(g GinkgoTInterface) *Cluster {
+// BuildCluster returns a singleton cluster with all terraform config and vars
+func BuildCluster(g GinkgoTInterface) *Cluster {
 	once.Do(func() {
 		var err error
-		cluster, err = newCluster(g)
+		_, err = newCluster(g)
 		if err != nil {
-			err = shared.ReturnLogError("error getting cluster: %w\n", err)
+			err = shared.ReturnLogError("error getting c: %w\n", err)
```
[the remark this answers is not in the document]
fmoral2: typo, tks

pkg/assert/host.go
```diff
@@ -9,17 +9,17 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-// CheckComponentCmdHost runs a command on the host and asserts that the value received contains the specified substring
+// CheckCmdHost runs a command on the host and asserts that the value received contains the specified substring
```
ShylajaDevadiga, line 12: ExecCmdOnHost or RunCmdOnHost

pkg/assert/node.go
```diff
@@ -38,10 +38,10 @@ func assertVersion(c customflag.FlagConfig) NodeAssertFunc {
 
 // assertCommit returns the NodeAssertFunc for asserting commit
 func assertCommit(c customflag.FlagConfig) NodeAssertFunc {
-	product, err := shared.GetProduct()
+	product, err := shared.FetchProduct()
 	Expect(err).NotTo(HaveOccurred(), "error getting product: %v", err)
 
-	commit, err := shared.GetProductVersion(product)
+	commit, err := shared.ProductVersion(product)
```
ShylajaDevadiga, line 44: FetchProductVersion to align with FetchProduct

pkg/assert/node.go
```diff
@@ -72,15 +72,15 @@ func NodeAssertReadyStatus() NodeAssertFunc {
 	}
 }
 
-// CheckComponentCmdNode runs a command on a node and asserts that the value received
+// CheckCmdNode runs a command on a node and asserts that the value received
 // contains the specified substring.
-func CheckComponentCmdNode(cmd, assert, ip string) error {
+func CheckCmdNode(cmd, assert, ip string) error {
```
ShylajaDevadiga, line 77: Same as above Run or Exec

pkg/assert/node.go
```diff
@@ -72,15 +72,15 @@ func NodeAssertReadyStatus() NodeAssertFunc {
...
 	if cmd == "" || assert == "" {
 		return shared.ReturnLogError("cmd and/or assert should not be sent empty")
 	}
 
 	Eventually(func(g Gomega) {
-		res, err := shared.RunCommandOnNode(cmd, ip)
+		res, err := shared.RunCmdNode(cmd, ip)
```
ShylajaDevadiga, line 83: Suggesting RunCmdOnNode to be more clear.

pkg/template/helper.go
```diff
@@ -56,11 +56,11 @@ func executeTestCombination(v VersionTestTemplate) error {
 	return nil
 }
 
-// AddTestCases returns the test case based on the name to be used as customflag.
-func AddTestCases(names []string) ([]testCase, error) {
-	var testCases []testCase
+// TestCases returns the test case based on the name to be used as customflag.
+func TestCases(names []string) ([]testCase, error) {
+	var tcs []testCase
```
ShylajaDevadiga, line 61: Let's retain testCases

pkg/template/helper.go
```diff
@@ -56,11 +56,11 @@ func executeTestCombination(v VersionTestTemplate) error {
...
 
-	testCase := map[string]testCase{
+	tc := map[string]testCase{
```
ShylajaDevadiga, line 63: same testCase/testcase

factory/cluster.go
```diff
@@ -27,7 +27,7 @@ func AddCluster(g GinkgoTInterface) *Cluster {
 
 // newCluster creates a new cluster and returns his values from terraform config and vars
 func newCluster(g GinkgoTInterface) (*Cluster, error) {
-	terraformOptions, varDir, err := addTerraformOptions()
+	opts, varDir, err := terraformOptions()
```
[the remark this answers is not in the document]
fmoral2: sure

modules/install/k3s_master.sh
```diff
@@ -51,7 +51,7 @@ policy_files() {
   fi
 }
 
-rhel() {
+network_manager() {
```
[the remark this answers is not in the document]
fmoral2: but what we are doing is stuff with net manager . OS config is too generic. could be any stuff.

pkg/assert/host.go
fmoral2: this func is calling the RunCmd.....to check some result.

pkg/template/helper.go
fmoral2: its too repetitive....in the func name we alreadu have TestCase, the name of the map is testCase...so no need to a var in use only in this scope, to be named like this.

fmoral2: it will be a shadow, the map already has this name

pkg/template/helper.go
fmoral2: it will be a shadow, the map already has this name

PR #40, opened 2023-11-07
Add test for k3s Traefik IngressRoute

rancher-max:
#### Proposed Changes ####

This adds a test for Traefik CRDs. It should be able to catch regressions such as https://github.com/k3s-io/k3s/issues/8755 in the future.

#### Types of Changes ####

Issue validation and regression enhancement

#### Testing ####

Checklist:
Jenkins jobs available to review upon request

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/8755


pkg/testcase/ingressdns.go
```diff
@@ -59,3 +63,68 @@ func TestDnsAccess(deleteWorkload bool) {
...
+	getIngressRoutePodsRunning := fmt.Sprintf("kubectl get pods -n test-ingressroute -l app=whoami" +
+		" --field-selector=status.phase=Running --kubeconfig=%s", shared.KubeConfigFile)
+	err = assert.ValidateOnHost(getIngressRoutePodsRunning, statusRunning)
+	Expect(err).NotTo(HaveOccurred(), err)
+
+	// Query the IngressRoute Host
+	pods, err := shared.GetPodsByNamespaceAndLabel("test-ingressroute", "app=whoami", false)
+	Expect(err).NotTo(HaveOccurred(), err)
+	
+	for _, pod := range pods {
```
fmoral2, line 98: i think here we have a opportunity to improve the code + the framework

what u think about creating a func on assert pkg that uses your Getpods.... and receives one args variadic something like that 
```
func  (args ...string) 
```
or receive 2 lists for simplicite , one list for cmds and another for asserts, 

```
func  ([]cmds,[]asserts) error {

here u call RunCommandHost(cmd) that accepts list of params and validate your stuff
}
```

then here on the test case it self you will abstract a lot just calling your new function that will do all of that 
```
pods, err := shared.GetPodsByNamespaceAndLabel("test-ingressroute", "app=whoami", false)

	for _, pod := range pods {
Comment
		Eventually(func(g Gomega) {
			// Positive test cases
			err = assert.CheckComponentCmdHost("curl -sk http://"+publicIp+"/notls",
				fmt.Sprintf("Hostname: %s", pod.Name),
				fmt.Sprintf("IP: %s", pod.NodeIP),
			)

			err = assert.CheckComponentCmdHost("curl -sk https://"+publicIp+"/tls",
				fmt.Sprintf("Hostname: %s", pod.Name),
				fmt.Sprintf("IP: %s", pod.NodeIP),
			)

			// Negative test cases
			err = assert.CheckComponentCmdHost("curl -sk http://"+publicIp+"/tls",
				"404 page not found",
			)

			err = assert.CheckComponentCmdHost("curl -sk https://"+publicIp+"/notls",
				"404 page not found",
			)
```

+ no repetion calling CheckComponentCmdHost a lot of times sequentially 



}

entrypoint/validatecluster/validatecluster_test.go
```diff
@@ -58,6 +58,10 @@ var _ = Describe("Test:", func() {
 		It("Verifies LoadBalancer Service", func() {
 			testcase.TestServiceLoadBalancer(true)
 		})
+
+		It("Verifies Traefik IngressRoute", func() {
```
fmoral2, line 62: this test will not be needed the upgrade flows ?

go.sum
```diff
@@ -191,8 +191,6 @@ github.com/OneOfOne/xxhash v1.2.2/go.mod h1:HSdplMjZKSmBqAxg5vPj2TmRDmfkzw+cTzAE
 github.com/agext/levenshtein v1.2.3 h1:YB2fHEn0UJagG8T1rrWknE3ZQzWM06O8AMAatNn7lmo=
 github.com/agext/levenshtein v1.2.3/go.mod h1:JEDfjyjHDjOF/1e4FlBE/PkbqA9OfWu2ki2W0IB5558=
 github.com/antihax/optional v1.0.0/go.mod h1:uupD/76wgC+ih3iEmQUL+0Ugr19nfwCT1kdvxnR2qWY=
-github.com/apparentlymart/go-dump v0.0.0-20180507223929-23540a00eaa3/go.mod h1:oL81AME2rN47vu18xqj1S1jPIPuN7afo62yKTNn3XMM=
-github.com/apparentlymart/go-textseg/v13 v13.0.0/go.mod h1:ZK2fH7c4NqDTLtiYLvIkEghdlcqw7yxLeM89kiTRPUo=
```
fmoral2, line 195: this file is updated on purpose? why ?

shared/cluster.go
```diff
@@ -336,6 +336,22 @@ func GetPods(print bool) ([]Pod, error) {
 	return pods, nil
 }
 
+// GetPodsByNamespaceAndLabel returns pods parsed from kubectl get pods in a specific namespace with a specific label
+func GetPodsByNamespaceAndLabel(namespace, label string, print bool) ([]Pod, error) {
```
fmoral2, line 340: do we have a need to print this ?

shared/cluster.go
```diff
@@ -347,21 +363,32 @@ func parsePods(res string) []Pod {
 			continue
 		}
 
-		if len(fields) < 9 {
+		if len(fields) == 9 {
```
fmoral2, line 366: can we improve this bunch of elseif , else and if ? 

idk maybe creating an offset with the minimun fields allowed and adding if it is a namespace or not

fmoral2: also sorry, why not counting with namespace sometimes?

pkg/testcase/ingressdns.go
```diff
@@ -59,3 +63,68 @@ func TestDnsAccess(deleteWorkload bool) {
 		Expect(err).NotTo(HaveOccurred(), "dnsutils manifest not deleted")
 	}
 }
+
+func TestIngressRoute(deleteWorkload bool) {
```
fmoral2, line 67: we need to add that on the helper.AddTestCases

entrypoint/validatecluster/validatecluster_test.go
rancher-max: Actually it's better to have in the upgrade too yeah. I'll add it

go.sum
rancher-max: This just happened from a `go mod tidy` and figured it wasn't harmful

shared/cluster.go
rancher-max: Nah but was keeping it similar to the `GetPods` function that already exists and has print capability. I actually don't love all the prints so I am more than happy to remove it 😄

shared/cluster.go
rancher-max: > can we improve this bunch of elseif , else and if ?
Probably! I didn't want to break what was already there. I'm not sure why there was the check in the first place and didn't really want to dig into it too hard as part of these changes.

> why not counting with namespace sometimes?
In the case of GetPodsByNamespace, doing like a `kubectl get pods -n somenamespace ...` the output doesn't have the NAMESPACE column at all, so this accounts for that. And in the case that function is being used, the namespace should already be known anyway so it's okay if it's not in the output.

shared/cluster.go
fmoral2: lol

but the getPods its pretty useful despite not liking as well

thats why i was wondering maybe this one is useful also

shared/cluster.go
fmoral2: got it tk for the explation

but tbh in this case we should totally improve that, sorry! you added a lot of more conditionals and code duplication

fmoral2: > Probably! I didn't want to break what was already there. I'm not sure why there was the check in the first place and didn't really want to dig into it too hard as part of these changes.

that check was to validate the range index, to never break if its less then 9 ( total of fields that we were expecting to deal with  +.1 )  
its a way to ensure that parsing above would not break ( which already happen due to the nature of parsing outputs from cli )

fmoral2: in those cases a good approach would working with offsets 
that way u dont need else even more elsif and avoid duplication.

tbh i dont have something done on my mind but something like: 
```
offset == 0

if len == {exact number when expecting NS} {
NamesPace = [0]   because its the first field right 
offset= 1
}
        Name    offset +1 
 
and then follow through



```

pkg/testcase/ingressdns.go
rancher-max: I agree this makes sense. Check out the latest commit for the implementation and let me know if that satisfies what you're thinking.

fmoral2: i liked but since we do already have 2 functions that take care of that as  assert.CheckComponentCmdHost and assert.ValidateOnHost it will unecessary creating new one on assert. 
i misinterpreted your need probably sorry.

also it will be too duplication of responsabilities have this CheckMultipleCmdHost, just calling the CheckComponentCmdHost both already range over asserts

i will leave u with a more lean and idiomatic approach.
 

```
	// Query the IngressRoute Host
	pods, err := shared.GetPodsByNamespaceAndLabel("test-ingressroute", "app=whoami", false)

	for _, pod := range pods {
		helper := struct {
			cmds []string
			asserts
		}{
			cmds: []string{
				"curl -sk http://" + publicIp + "/notls",
				"curl -sk https://" + publicIp + "/tls",
			},
			asserts: asserts{
				positive: []string{
					fmt.Sprintf("Hostname: %s", pod.Name),
					fmt.Sprintf("IP: %s", pod.NodeIP),
				},
				negative: "404 page not found",
			},
		}

		for _, cmd := range helper.cmds {
			err = assert.CheckComponentCmdHost(cmd, helper.asserts.positive...)
			err = assert.CheckComponentCmdHost(cmd, helper.asserts.negative)
			Expect(err).To(HaveOccurred(), err)
		}
	}


```

rancher-max: hmmm I see what you wanted now and the solution you provided is really nice! But in that case I think we should probably revert it to _almost how_ I had it originally as there isn't a direct need for including a new type on this. I will modify to use a string slice for the asserts though.

rancher-max: See latest commit and let me know what you think -- there are actually four separate commands here (even though they're very similar). This actually has the least amount of nested loops as well so is probably the most ideal!

fmoral2: for me its totally fine though, however i really prefer 2  for loops  rather rather than call 4 for times ` assert.CheckComponentCmdHost(` a function , plus imagine that you have a 15 positive and 10 negative asserts to do, it would go crazy this , thats why the usage of the struct here, to create a table driven strategy.
even though having 2 cmds only .
but again if this suits your actual need ! go for it

rancher-max: Yeah let's go with this right now. We can refactor in the future if need arises!

workloads/amd64/ingressroute.yaml
```diff
@@ -0,0 +1,105 @@
...
+        app: whoami
+    spec:
+      containers:
+        - name: whoami
+          image: traefik/whoami
+          ports:
+            - name: web
+              containerPort: 80
+---
+apiVersion: traefik.containo.us/v1alpha1
```
ShylajaDevadiga, line 75: We will need to test new apiGroup `traefik.containo.us/v1alpha1`

mdrahman-suse (commented):
LGTM, will approve once the comments are addressed

shared/cluster.go
rancher-max: Addressed with https://github.com/rancher/distros-test-framework/pull/40/commits/780acf3dc2bd5387e112dffc3ba3dd2ae96cec5e but I kind of like the if/else better myself. Let me know what you (and others) think here and we can use this approach if desired!

workloads/amd64/ingressroute.yaml
rancher-max: Added a test for this but only in validatecluster for now. At some point we'll want to completely remove the older one when it's officially removed

shared/cluster.go
fmoral2: let others give their opinion, for me makes no sense adding else, if, else if unnecessary. 
also duplicating code.

shared/cluster.go
```diff
@@ -357,30 +357,26 @@ func parsePods(res string) []Pod {
...
-		p = &Pod{
-			Name:      fields[offset],
-			Ready:     fields[offset + 1],
-			Status:    fields[offset + 2],
-			Restarts:  fields[offset + 3],
-			Age:       fields[offset + 4],
-			NodeIP:    fields[offset + 5],
-			Node:      fields[offset + 6],
+		var p Pod
+		if len(fields) != 9 {
```
fmoral2, line 367: if len(fields) == 8 {
	      pod.NameSpace = fields[0]
	   	offset = 1
 }

rancher-max: I saw this, made the change, and didn't commit it 🤦

shared/cluster.go
```diff
@@ -336,36 +336,49 @@ func GetPods(print bool) ([]Pod, error) {
...
 		fields := strings.Fields(rec)
-		if strings.TrimSpace(rec) == "" {
+		if strings.TrimSpace(rec) == "" || len(fields) < 9 {
 			continue
 		}
-
-		if len(fields) < 9 {
-			continue
+		var p Pod
+		if len(fields) == 10 {
```
fmoral2, line 367: is this working fine ?

rancher-max: Yep! 
image


Length needs to be 10 with the Namespace because it's one additional field

fmoral2: but its not 7 the pod structure ?

fmoral2 (approved):

shared/cluster.go
rancher-max: Thanks, didn't see this message! What do you mean by "7 the pod structure"? This works in both cases the function is called: either with namespaces (kubectl get pods -A) or without (kubectl get pods -n default)

fmoral2: no worries, sorry for dont send complete context , why are you doing `		if len(fields) == 10 {`  if the size is 8 ? with the namespace
 
```

type Pod struct {
	NameSpace string
	Name      string
	Ready     string
	Status    string
	Restarts  string
	Age       string
	NodeIP    string
	Node      string
}
```

rancher-max: Ah the length is 10! There are two more fields that we don't have accounted for right now:
`NOMINATED NODE` and `READINESS GATES`. But those values are always <none> for us so we don't necessarily need to account for them right now.
```
NAMESPACE        NAME                                         READY   STATUS      RESTARTS   AGE     IP           NODE              NOMINATED NODE   READINESS GATES
kube-system      pod/coredns-77ccd57875-6mhdm                 1/1     Running     0          8m59s   10.42.0.6    ip-172-31-5-154   <none>           <none>
kube-system      pod/helm-install-traefik-crd-9kjss           0/1     Completed   0          8m59s   10.42.0.5    ip-172-31-5-154   <none>           <none>
kube-system      pod/helm-install-traefik-j6gws               0/1     Completed   1          8m59s   10.42.0.3    ip-172-31-5-154   <none>           <none>
kube-system      pod/local-path-provisioner-957fdf8bc-gnn45   1/1     Running     0          8m59s   10.42.0.4    ip-172-31-5-154   <none>           <none>
kube-system      pod/metrics-server-648b5df564-vjdjl          1/1     Running     0          8m59s   10.42.0.2    ip-172-31-5-154   <none>           <none>
kube-system      pod/svclb-traefik-6186304f-fpxmn             2/2     Running     0          8m24s   10.42.0.7    ip-172-31-5-154   <none>           <none>
kube-system      pod/traefik-768bdcdcdd-r2d7v                 1/1     Running     0          8m24s   10.42.0.8    ip-172-31-5-154   <none>           <none>
```

fmoral2: yep , but we mapped this to 
```
type Pod struct {
	NameSpace string
	Name      string
	Ready     string
	Status    string
	Restarts  string
	Age       string
	NodeIP    string
	Node      string
}
```

- so our map ( not the real fields ) is 8, so lets count on the implementation details.

fmoral2: if we need to grow to the real fields , then we validate against that.  We should always validate against whats implemented at the moment.

fmoral2: either way its fine!

rancher-max: Alright let's just leave as is then and we can add some enhancement in the future if needed to include the two additional fields in the map?

fmoral2: leave as it is you mean ? 
```
if len(fields) == 8 {
```

if so yes, if its == 10 no. 

as i stated above we should validate , count and deal with what we have implemented.

and right now we dont have 10 implemented.

rancher-max: `if len(fields) == 8 {`isn't allowed as that isn't a valid scenario and it will never hit that block, since the length of the fields is either 9 or 10. This checks the length of the _actual_ fields on the result, not the way we've mapped the fields.

fmoral2: if thats the case, then yes. lets add as it is

rancher-max: Updated!

rancher-max:
Rebased and made a small update to pods map to reflect the actual state of the fields.
image

Also when doing this, I noticed that sometimes the initial pods check would show passed with 0 pods. This was a race condition where the Node would be in Ready state but the pods hadn't started up yet. Therefore, I also added a small check to that testcase to ensure that the pods aren't empty.

fmoral2 (dismissed):

fmoral2 (approved):

VestigeJ (approved):
LGTM

aganesh-suse (approved):
LGTM

ShylajaDevadiga (approved):

PR #38, opened 2023-10-30
fix: external db config 

fmoral2:
#### Proposed Changes ####

- Fix external db issue see -> https://github.com/rancher/distros-test-framework/issues/37 
- General improvements related
- Adding make command to create new containers without any manual intervation
- fix manage workload signature 

#### Testing ####

Checklist:
1. If your PR changes anything on or related to Jenkins, run it pointing to your branch to make sure it's okay.
 -n/a

2. Verify code lint; we should not have errors.

- ok
3. Update the documentation if needed.

- ok
4. Update makefile and docker run if adding new tests.
- ok

5. Run your tests at least 4 times with all configurations needed and possible.
- ok

6. If needed test with different os types.
-ok

PR #41, opened 2023-11-08
fix: selinux tests

fmoral2:
#### Proposed Changes ####

- fix selinux tests that was breaking

#### Testing ####

Checklist:
- no, but will create a new one 

 - ok fixed and the error on the lint i am fixing in the  https://github.com/rancher/distros-test-framework/pull/38 

- n/a 
- updated 

- yes both products 

- yes, centos7 , centos8

#### Further Comments ####

tests passing without upgrade and with


pkg/testcase/selinux.go
```diff
@@ -12,56 +11,36 @@ import (
...
 
 type configuration struct {
 	distroName string
 	cmdCtx
 }
 
-// TestSelinuxEnabled Validates that containerd is running with selinux enabled in the config
-func TestSelinuxEnabled() {
+// setHelper returns the cluster and product
+func setHelper() (*factory.Cluster, string, error) {
```
mdrahman-suse, line 22: This func name is confusing. Can we rename to reflect its purpose like `getClusterProduct`

fmoral2: sure

pkg/testcase/selinux.go
```diff
@@ -209,42 +153,35 @@ func selectSelinuxPolicy(product, osType string) cmdCtx {
...
-		log.Println(err)
-	}
-	cluster := factory.AddCluster(GinkgoT())
-	var serverUninstallCmd string
-	var agentUninstallCmd string
-	var serverCmd string
+	cluster, product, err := setHelper()
+	Expect(err).NotTo(HaveOccurred())
+
+	serverUninstallCmd := "sudo rke2-uninstall.sh"
```
mdrahman-suse, line 177: can we extract these into shared func? also this will fail if the path to `*-uninstall.sh` is different... as in some OS's it is different: `/usr/local/bin` vs `/opt/local/bin`

fmoral2: why shared if its specific to selinux ?

mdrahman-suse (commented):
Few comments

pkg/testcase/selinux.go
mdrahman-suse: the uninstall can be re-purposed for other testing, like validating the uninstallation itself, hence I asked if it can be moved...

mdrahman-suse (dismissed):

fmoral2:
` Red Hat Enterprise Linux Server 7.9 (Maipo);`
Screenshot 2023-11-10 at 12 59 37



` CentOS Linux 7 (Core)":`

Screenshot 2023-11-10 at 13 03 40




`  Oracle Linux Server 8.7 : `

Screenshot 2023-11-10 at 13 04 54

.golangci.yaml
```diff
@@ -5,27 +5,26 @@ linters:
...
+  gofmt:
+    simplify: true
   govet:
     check-shadowing: true
   gci:
     local-prefixes: github.com/rancher/distros-test-framework
-  gosec:
-    issues:
-      - G306
-      - G106
```
VestigeJ, line 20: Are these removed because the framework is hitting edge cases and failing to adhere to these rules?

fmoral2: it was removed not on purpose , going back

mdrahman-suse (approved):
LGTM

shared/aux.go
VestigeJ, line 1: Just a few linting fixes it seems there's a rule directly against ending the error return with a newline - not sure why the rule is there but if we're using it we should adhere to it.

VestigeJ (changes requested):
See aux.go for linting fixes - changing the error returns to not send newlines. or we edit the linting rules to ignore this if the output isn't working the way we want.

shared/aux.go
fmoral2: this is not an error, just a warn. 

 plus on that place we want to break line to have a better output debugging

fmoral2:
@VestigeJ @mdrahman-suse  , just needed to solve conflicts !

VestigeJ (dismissed):

VestigeJ (approved):
✅

mdrahman-suse (approved):
LTGM

PR #42, opened 2023-11-10
Cert rotate test

aganesh-suse:
#### Proposed Changes ####

Automating the certificate rotate testcase
Added the following: 
1) Helper methods for stop/start services - rke2 or k3s and certificate rotate operation in shared/cluster.go
2) testcase for running the cert rotate test in pkg/testcase/certificates.go
3) Separate Entrypoint/Test Suite for cert rotate: entrypoint/certrotate
4) changes to Makefile and test_runner.sh to call the certrotate tests. 
5) Added docs/test_related.md to convey information regarding setup assumptions for this testcase to be successful

#### Types of Changes ####

Patch Validation testcase

#### Testing ####

Checklist:
N/A.

Done

Added a test_related.md section for any particular test related information. 

Done

Done

Not OS related. Tested with sles

#### Linked Issues ####

https://github.com/rancher/qa-tasks/issues/597
https://github.com/rancher/qa-tasks/issues/598


docs/test_related.md
```diff
@@ -0,0 +1,28 @@
+### Setup related Information
```
fmoral2, line 1: love this!  :) we should have added from the very beginning

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/rancher/distros-test-framework/factory"
+	"github.com/rancher/distros-test-framework/shared"
+)
+
+func errCheck(error error, message string, ignore bool) {
+	if error != nil {
+		shared.LogLevel("error", message)
```
fmoral2, line 16: we dont need that!   also if you want we already have a shared func that logs and return err ! 
https://github.com/rancher/distros-test-framework/blob/dbc0224ca28dba1bd9567c33a0a89112ed2904bf/shared/aux.go#L290

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+		if !ignore {
+			Expect(error).NotTo(HaveOccurred())
+		}
+
+	}
+}
+
+// Rotate certificate for etcd only and cp only nodes
+func certRotate(product string, ip string) {
+	// Stop service on server
```
fmoral2, line 26: your code should talk by itself, comments are needed in tryck blocks or very specific stuff.
and the name funcs are already good and clear

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+	// Retry Stop start of service. (Inconsistent failure to start service above,
+	// can be fixed hopefully with a restart)
+	if startError != nil {
+		_, retryError := shared.StopStartService(product, ip, "server")
+		errCheck(retryError, fmt.Sprintf("ERROR: Retrying Stop/Start %s service", product), false)
+	}
+}
+
+// Compare TLS Directories before and after cert rotation to display identical files
+func compareTLSDir(product string, ip string) (string, error) {
```
fmoral2, line 47: func compareTLSDir(product , ip string)   dont need 2 strings!

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+	dataDir := fmt.Sprintf("/var/lib/rancher/%s", product)
+	serverDir := fmt.Sprintf("%s/server", dataDir)
+	origTLSDir := fmt.Sprintf("%s/tls", serverDir)
+	cmd := fmt.Sprintf("sudo ls -lt %s/ | grep tls | awk {'print $9'} | sed -n '2 p'", serverDir)
+	tlsDir, error := shared.RunCommandOnNode(cmd, ip)
+	if error != nil {
+		shared.LogLevel("warn", "Unable to get new TLS Directory name")
```
fmoral2, line 54: not getting the dir name , its not supposed to break execution right ?

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	shared.LogLevel("info", fmt.Sprintf("TLS Directory name: %s", tlsDir))
+	newTLSDir := fmt.Sprintf("%s/%s", serverDir, tlsDir)
+	shared.LogLevel("info", "Comparing Directories: %s and %s", origTLSDir, newTLSDir)
+	cmd2 := fmt.Sprintf("sudo diff -sr %s/ %s/ | grep -i identical | "+
+		"awk '{print $2}' | xargs basename -a | "+
+		"awk 'BEGIN{print \"Identical Files:  \"}; {print $1}'", origTLSDir, newTLSDir)
+	return shared.RunCommandOnNode(cmd2, ip)
```
fmoral2, line 62: missing break like  before last code to return 
```

		"awk 'BEGIN{print \"Identical Files:  \"}; {print $1}'", origTLSDir, newTLSDir)

	return shared.RunCommandOnNode(cmd2, ip)
```

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func TestCertRotate() {
+	// Node Types needed for this test:
```
fmoral2, line 66: since u added a doc file for that lets remove the comments

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+	time.Sleep(time.Duration(10) * time.Second) // Adding 10 second sleep between steps
+	// Retry Stop start of service. (Inconsistent failure to start service above,
+	// can be fixed hopefully with a restart)
+	if startError != nil {
+		_, retryError := shared.StopStartService(product, ip, "server")
+		errCheck(retryError, fmt.Sprintf("ERROR: Retrying Stop/Start %s service", product), false)
+	}
+}
+
+// Compare TLS Directories before and after cert rotation to display identical files
```
fmoral2, line 46: naming functions doc should be exact the same name and should start with it 
```

// compareTLSDir Compare TLS Directories before and after cert rotation to display identical files
func compareTLSDir(product string, ip string) (string, error) {

``

File: pkg/testcase/certificates.go [hunk 9]
```diff
@@ -0,0 +1,122 @@
...
+func compareTLSDir(product string, ip string) (string, error) {
+	dataDir := fmt.Sprintf("/var/lib/rancher/%s", product)
+	serverDir := fmt.Sprintf("%s/server", dataDir)
+	origTLSDir := fmt.Sprintf("%s/tls", serverDir)
+	cmd := fmt.Sprintf("sudo ls -lt %s/ | grep tls | awk {'print $9'} | sed -n '2 p'", serverDir)
+	tlsDir, error := shared.RunCommandOnNode(cmd, ip)
```
fmoral2, line 52: you shouldnt be naming error var as error, its already the interface name

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+	shared.LogLevel("info", fmt.Sprintf("TLS Directory name: %s", tlsDir))
+	newTLSDir := fmt.Sprintf("%s/%s", serverDir, tlsDir)
+	shared.LogLevel("info", "Comparing Directories: %s and %s", origTLSDir, newTLSDir)
+	cmd2 := fmt.Sprintf("sudo diff -sr %s/ %s/ | grep -i identical | "+
+		"awk '{print $2}' | xargs basename -a | "+
+		"awk 'BEGIN{print \"Identical Files:  \"}; {print $1}'", origTLSDir, newTLSDir)
+	return shared.RunCommandOnNode(cmd2, ip)
+}
+
+func TestCertRotate() {
```
fmoral2, line 65: your first function of the file should be this one. the exported

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+	time.Sleep(time.Duration(10) * time.Second) // Adding 10 second sleep between steps
+	// start service on server
+	_, startError := shared.StartService(product, ip, "server")
+	errCheck(startError, fmt.Sprintf("Error starting %s service", product), true)
+	time.Sleep(time.Duration(10) * time.Second) // Adding 10 second sleep between steps
+	// Retry Stop start of service. (Inconsistent failure to start service above,
+	// can be fixed hopefully with a restart)
+	if startError != nil {
+		_, retryError := shared.StopStartService(product, ip, "server")
+		errCheck(retryError, fmt.Sprintf("ERROR: Retrying Stop/Start %s service", product), false)
```
fmoral2, line 42: error messages dont start with capital should be  "error...."

shared/cluster.go
```diff
@@ -409,3 +409,43 @@ func WriteDataPod(namespace string) (string, error) {
 
 	return RunCommandHost(cmd)
 }
+
+// Get service name. Used to work with stop/start k3s/rke2 services
```
fmoral2, line 413: https://github.com/rancher/distros-test-framework/pull/42/files#r1389310242

shared/cluster.go
```diff
@@ -409,3 +409,43 @@ func WriteDataPod(namespace string) (string, error) {
...
+}
+
+// stop_start k3s or rke2 service on the node
+func StopStartService(product string, ip string, serviceType string) (error, error) {
+	_, stopError := StopService(product, ip, serviceType)
+	_, startError := StartService(product, ip, serviceType)
+	return stopError, startError
+}
+
+// certificate rotate for k3s or rke2
```
fmoral2, line 447:
```
// CertRotate  rotate certs  for k3s or rke2
func CertRotate(product, ip string) (string, error) {
	cmd := fmt.Sprintf("sudo %s certificate rotate", product)

	res, err := RunCommandOnNode(cmd, ip)
	if err != nil {
		return "", err
	}

	return res, nil
}

```

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+
+	}
+}
+
+// Rotate certificate for etcd only and cp only nodes
+func certRotate(product string, ip string) {
+	// Stop service on server
+	_, stopError := shared.StopService(product, ip, "server")
+	errCheck(stopError, fmt.Sprintf("Error stopping %s service", product), false)
+	time.Sleep(time.Duration(10) * time.Second) // Adding 10 second sleep between steps
```
fmoral2, line 29: lets remove please all the sleeps.
we should deal with it in other layer(shared ) and probably another way, like u did here
```
	cmd := fmt.Sprintf("timeout 2m sudo systemctl start %s", getServiceName(product, serviceType))

```

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+
+func TestCertRotate() {
+	// Node Types needed for this test:
+	// Server1 - Etcd only node
+	// Server2 - Control Plane only node
+	// Agent1 - Worker node
+	// Initialize the cluster
+	cluster := factory.AddCluster(GinkgoT())
+	// Get server1, server2, product info from cluster
+	server1 := cluster.ServerIPs[0]
```
fmoral2, line 73: this test will always be run with: 
- 2 servers 
- 1 agent 
 ??

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Rotate certificate for etcd only and cp only nodes
+func certRotate(product string, ip string) {
+	// Stop service on server
+	_, stopError := shared.StopService(product, ip, "server")
+	errCheck(stopError, fmt.Sprintf("Error stopping %s service", product), false)
+	time.Sleep(time.Duration(10) * time.Second) // Adding 10 second sleep between steps
+	// Rotate certificate
+	_, rotateError := shared.CertRotate(product, ip)
+	errCheck(rotateError, fmt.Sprintf("Error running certificate rotate for %s service", product), false)
```
fmoral2, line 32: give space in your code to better represent context blocks!!

```
// Compare TLS Directories before and after cert rotation to display identical files
func compareTLSDir(product string, ip string) (string, error) {
	dataDir := fmt.Sprintf("/var/lib/rancher/%s", product)
	serverDir := fmt.Sprintf("%s/server", dataDir)
	origTLSDir := fmt.Sprintf("%s/tls", serverDir)

	cmd := fmt.Sprintf("sudo ls -lt %s/ | grep tls | awk {'print $9'} | sed -n '2 p'", serverDir)
	tlsDir, error := shared.RunCommandOnNode(cmd, ip)
	if error != nil {
		shared.LogLevel("warn", "Unable to get new TLS Directory name")
	}

	shared.LogLevel("info", fmt.Sprintf("TLS Directory name: %s", tlsDir))

	newTLSDir := fmt.Sprintf("%s/%s", serverDir, tlsDir)
	shared.LogLevel("info", "Comparing Directories: %s and %s", origTLSDir, newTLSDir)

	diff := fmt.Sprintf("sudo diff -sr %s/ %s/ | grep -i identical | "+
		"awk '{print $2}' | xargs basename -a | "+
		"awk 'BEGIN{print \"Identical Files:  \"}; {print $1}'", origTLSDir, newTLSDir)
	return shared.RunCommandOnNode(diff, ip)
}

also when possible give meaning on the vars like   cmd2 => diff 

```

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+func TestCertRotate() {
+	// Node Types needed for this test:
+	// Server1 - Etcd only node
+	// Server2 - Control Plane only node
+	// Agent1 - Worker node
+	// Initialize the cluster
+	cluster := factory.AddCluster(GinkgoT())
+	// Get server1, server2, product info from cluster
+	server1 := cluster.ServerIPs[0]
+	server2 := cluster.ServerIPs[1]
```
fmoral2, line 74: product, err := shared.GetProduct()
	ips := shared.FetchNodeExternalIP()
	ipagent := ips[1:]
	ipservers := ips[:1]
	// some logic to split agent and servers ips because here it will be better get from the current cluster state,
	// //not from the immutable data from the addcluster

shared/aux.go
```diff
@@ -45,7 +46,7 @@ func RunCommandOnNode(cmd, ip string) (string, error) {
 	if cmd == "" {
 		return "", ReturnLogError("cmd should not be empty")
 	}
-
+	LogLevel("debug", fmt.Sprintf("EXECUTE: %s on %s", cmd, ip))
```
fmoral2, line 49: why this here ?

shared/aux.go
```diff
@@ -72,9 +73,9 @@ func RunCommandOnNode(cmd, ip string) (string, error) {
 		!strings.Contains(cleanedStderr, "2")) {
 		return cleanedStderr, nil
 	} else if cleanedStderr != "" {
-		return "", fmt.Errorf("command: %s failed with error: %v\n", cmd, stderr)
+		return "", fmt.Errorf("command: %s failed with error: %v", cmd, stderr)
```
fmoral2, line 76: please roolback the \n, makes huge difference on console

shared/cluster.go
```diff
@@ -409,3 +409,43 @@ func WriteDataPod(namespace string) (string, error) {
 
 	return RunCommandHost(cmd)
 }
+
+// Get service name. Used to work with stop/start k3s/rke2 services
+func getServiceName(product string, serviceType string) string {
```
fmoral2, line 414: no need for this func!!

this u could simplify a lot to this: using * 
```
		cmd = fmt.Sprintf("timeout 2m sudo systemctl start %s*", product)

```

shared/cluster.go
```diff
@@ -409,3 +409,43 @@ func WriteDataPod(namespace string) (string, error) {
...
+		serviceName = product // k3s
+	} else { // k3s-agent, rke2-server, rke2-agent
+		serviceName = fmt.Sprintf("%s-%s", product, serviceType)
+	}
+	LogLevel("debug", fmt.Sprintf("Service Name: %s", serviceName))
+	return serviceName
+}
+
+// stop k3s or rke2 service on the node
+func StopService(product string, ip string, serviceType string) (string, error) {
```
fmoral2, line 426: these 3 func u can also improve your code and the framework as a whole creating something like a ManagerService . maybe something like that 
```
func ManageClusterService(action, product string, agents, servers []string) error {
	var err error

	for _, ip := range agents {
		if err = serviceCmds(action, product, ip); err != nil {
			return err
		}
	}

	for _, ip := range servers {
		if err = serviceCmds(action, product, ip); err != nil {
			return err
		}
	}

	return nil
}

func serviceCmds(action, product, ip string) error {
	var cmd string
	var err error

	switch action {
	case "start":
		cmd = fmt.Sprintf("timeout 2m sudo systemctl start %s*", product)
		_, err = RunCommandOnNode(cmd, ip)
		if err != nil {
			return err
		}
	case "stop":
		cmd = fmt.Sprintf("sudo systemctl stop %s*", product)
		_, err = RunCommandOnNode(cmd, ip)
		if err != nil {
			return err
		}
	default:
		return ReturnLogError("invalid action: %s. Must be 'start' or 'stop'", action)
	}

	return nil
}
```

fmoral2: it will also remove some burden from your test cases

pkg/testcase/cluster.go
```diff
@@ -4,12 +4,12 @@ import (
 	"fmt"
 	"strings"
 
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
```
fmoral2, line 9: lets add this on the last imports
order:

- internal go
- internal pkg
- internal project
- external libs

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
+package testcase
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
```
fmoral2, line 7: https://github.com/rancher/distros-test-framework/pull/42/files#r1389389670

pkg/testcase/certificates.go
```diff
@@ -0,0 +1,122 @@
...
+		"server-ca.crt",
+		"server-ca.key",
+		"request-header-ca.crt",
+		"request-header-ca.key",
+		"server-ca.crt",
+		"server-ca.key",
+		"server-ca.nochain.crt",
+		"service.current.key",
+		"service.key"}
+	for i := 0; i < len(fileNames); i++ {
```
fmoral2, line 116: same comment for the context blocks spaces !

fmoral2:
Very good job! just left comments more related to golang and some patterns !

Makefile
```diff
@@ -82,6 +82,10 @@ remove-tf-state:
 test-create:
 	@go test -timeout=45m -v -count=1 ./entrypoint/createcluster/...
 
+.PHONY: test-cert-rotate
+test-create:
```
fmoral2, line 86: wrong name for make target

pkg/testcase/certificates.go
fmoral2: just a tip to try not use the ones came from  cluster := factory.AddCluster(GinkgoT())

shared/aux.go
```diff
@@ -54,7 +55,7 @@ func RunCommandOnNode(cmd, ip string) (string, error) {
 	stdout, stderr, err := runsshCommand(cmd, conn)
 	if err != nil && !strings.Contains(stderr, "restart") {
 		return "", fmt.Errorf(
-			"command: %s failed on run ssh: %s with error: %w\n",
+			"command: %s failed on run ssh: %s with error: %w",
```
fmoral2, line 58: https://github.com/rancher/distros-test-framework/pull/42/files#r1389382879

docs/test_related.md
mdrahman-suse: +1

pkg/testcase/certificates.go
mdrahman-suse, line 1: Suggest renaming the file to certrotate.go

Makefile
aganesh-suse: fixed.

shared/cluster.go
fmoral2: + u can also add more stuff like restart and etc

pkg/testcase/certificates.go
fmoral2: + 1

docs/test_related.md
mdrahman-suse: Suggestion: Rename file to `testing.md`

entrypoint/certrotate/certrotate_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	It("Validate Pods", func() {
+		testcase.TestPodStatus(
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus(),
+		)
+	})
+
+	It("Certificate Rotate", func() {
```
mdrahman-suse, line 33: Suggestion:  Rename test description ->`Validate Certificate Rotation`

mdrahman-suse (commented):
LGTM other than changes suggested

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	ip, rotateError := product.CertRotate(ips)
+	if rotateError != nil {
+		shared.ReturnLogError(fmt.Sprintf("Error running certificate rotate for %s service on %s", product, ip))
+	}
+
+	startError := product.ManageClusterService(shared.Start, shared.Server, ips)
+	if startError != nil {
+		shared.ReturnLogError(fmt.Sprintf("Error starting %s service", product))
+	}
+	// Preventative restart - sometimes, start hangs after a cert rotate which gets fixed on a second restart
```
mdrahman-suse, line 52: So instead of doing start and then restart, have you tried to see if only restarting services works or not?

docs/test_related.md
aganesh-suse: fixed

pkg/testcase/certificates.go
aganesh-suse: removed and using ReturnLogError instead

pkg/testcase/certificates.go
aganesh-suse: removed most comments

pkg/testcase/certificates.go
aganesh-suse: updated this method.

pkg/testcase/certificates.go
aganesh-suse: updated to ReturnLogError

pkg/testcase/certrotate.go
aganesh-suse: it was already in stopped state - so did not try restart directly. going to try now. will keep you posted.

pkg/testcase/certificates.go
aganesh-suse: done

pkg/testcase/certificates.go
aganesh-suse: done

pkg/testcase/certificates.go
aganesh-suse: fixed

pkg/testcase/certificates.go
aganesh-suse: fixed

pkg/testcase/certificates.go
aganesh-suse: fixed

shared/cluster.go
aganesh-suse: fixed

shared/cluster.go
aganesh-suse: fixed

pkg/testcase/certificates.go
aganesh-suse: fixed

pkg/testcase/certificates.go
aganesh-suse: fixed

pkg/testcase/certificates.go
aganesh-suse: fixed

pkg/testcase/certificates.go
aganesh-suse: this did not work. the ips returned are never in the same order - hard to tell which is server/agent. 
going back to the previous way of getting ips for now.

shared/aux.go
aganesh-suse: to log the commands run - even if there was no error. this helps with debugging and get the flow and figure out which node the command is being run on etc. it is better to add this here instead of adding particular logs outside everytime a run command method is called. this method knows the cmd and where it runs the same automatically.

shared/aux.go
aganesh-suse: done

shared/cluster.go
aganesh-suse: need this method. tried without it and get this error message: 

```
EXECUTE: timeout 2m sudo systemctl start rke2* on 18.225.55.124 
WARN[2023-11-16T05:14:47Z]/go/src/github.com/rancher/distros-test-framework/shared/aux.go:318 github.com/rancher/distros-test-framework/shared.LogLevel Warning: systemctl start called with a glob pattern.
Hint: unit globs expand to loaded units, so start will usually have no effect.
   Passing --all will also load units which are pulled in by other units.
   See systemctl(1) for more details.
Job for rke2-server.service failed because the control process exited with error code.
See "systemctl status rke2-server.service" and "journalctl -xeu rke2-server.service" for details.
 
ERRO[2023-11-16T05:14:47Z]/go/src/github.com/rancher/distros-test-framework/shared/aux.go:298 github.com/rancher/distros-test-framework/shared.ReturnLogError Error starting rke2 service
Last call: github.com/rancher/distros-test-framework/pkg/testcase.certRotate in /go/src/github.com/rancher/distros-test-framework/pkg/testcase/certrotate.go:57
```

shared/cluster.go
aganesh-suse: fixed

pkg/testcase/cluster.go
aganesh-suse: fixed

pkg/testcase/certificates.go
aganesh-suse: fixed

shared/cluster.go
```diff
@@ -409,3 +409,139 @@ func WriteDataPod(namespace string) (string, error) {
 
 	return RunCommandHost(cmd)
 }
+
+type ServiceType string
+
+const (
+	Server ServiceType = "server"
```
mdrahman-suse, line 416: Suggestion: Using `nodeType` instead of `serviceType`?

pkg/testcase/certificates.go
aganesh-suse: fixed

shared/aux.go
aganesh-suse: fixed

pkg/testcase/certificates.go
aganesh-suse: fixed

entrypoint/certrotate/certrotate_test.go
aganesh-suse: fixed

shared/cluster.go
```diff
@@ -409,3 +409,139 @@ func WriteDataPod(namespace string) (string, error) {
...
+		}
+	}
+
+	return false
+}
+
+type ServiceAction string
+
+const (
+	Stop    ServiceAction = "stop"
```
mdrahman-suse, line 434: Are these part of systemctl commands?

pkg/testcase/certrotate.go
aganesh-suse: yup - you were right 👍 restart seems to work. ran 2 times so far without issues - so changed to this. will run some more to double check. but going this route for now.

shared/cluster.go
aganesh-suse: yes

shared/cluster.go
```diff
@@ -409,3 +409,139 @@ func WriteDataPod(namespace string) (string, error) {
...
+		}
+	}
+
+	return false
+}
+
+type Product string
+
+const (
+	K3S  Product = "k3s"
```
mdrahman-suse, line 465: We should already have the product from .env file, isnt it?

aganesh-suse: just trying to create an enum kind of class - and have all methods related to k3s/rke2 here. 
we do get the product from .env => based on which we set the product := shared.K3S or product := shared.RKE2

and we run commands like 
product.ManageClusterService -> for start/stop k3s or rke2 etc. 

similarly product.CertRotate

since all these are really commands we are running on our k3s/rke2 service/product.

aganesh-suse: will talk about this tomm in the standup.

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	var product shared.Product
+	if p == "k3s" {
+		product = shared.K3S
+	} else {
+		product = shared.RKE2
+	}
+
+	certRotate(product, serverIPs)
+
+	product.ManageClusterService(shared.Restart, shared.Agent, agentIPs)
```
fmoral2, line 29: you need to handle error callback here since this returns err

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	cmd := fmt.Sprintf("sudo ls -lt %s/ | grep tls | awk {'print $9'} | sed -n '2 p'", serverDir)
+
+	for _, ip := range ips {
+
+		shared.LogLevel("info", fmt.Sprintf("Working on node with ip: %s", ip))
+
+		// Get new tls directory name
+		tlsDir, tlsError := shared.RunCommandOnNode(cmd, ip)
+		if tlsError != nil {
+			shared.ReturnLogError(fmt.Sprintf("Unable to get new TLS Directory name for %s", ip))
```
fmoral2, line 91: this func does not return error, so you shouldnt use a "Return...err" func

or change the behavior (signature) or just simple "return" and Log

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	compareAndVerifyTLSDirContent(product, serverIPs)
+
+}
+
+// certRotate Rotate certificate for etcd only and cp only nodes
+func certRotate(product shared.Product, ips []string) {
+
+	stopError := product.ManageClusterService(shared.Stop, shared.Server, ips)
+	if stopError != nil {
+		shared.ReturnLogError(fmt.Sprintf("Error stopping %s service", product))
```
fmoral2, line 40: https://github.com/rancher/distros-test-framework/pull/42/files#r1397247590 
same here and for all returns

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/rancher/distros-test-framework/factory"
+	"github.com/rancher/distros-test-framework/shared"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCertRotate() {
+	// Please refer to docs/testing.md for instructions on setup for this test
+
```
fmoral2, line 15: no need for this first empty line + no need to ref to docs
this ref should be documented in some doc, readme , doesn matter, but not in code .

pkg/testcase/certificates.go
fmoral2: not fixed 

```
		shared.ReturnLogError(fmt.Sprintf("Error running certificate rotate for %s service on %s", product, ip))
	}

```

fmoral2: same for all error messages on this method

shared/cluster.go
```diff
@@ -409,3 +409,139 @@ func WriteDataPod(namespace string) (string, error) {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// CertRotate certificate rotate for k3s or rke2
+func (product Product) CertRotate(ips []string) (string, error) {
+	if !product.IsValidProduct() {
```
fmoral2, line 535: also fix this please - 


receiver-naming: receiver name product should be consistent with previous receiver name p for Product (revive)

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	product.ManageClusterService(shared.Restart, shared.Agent, agentIPs)
+
+	compareAndVerifyTLSDirContent(product, serverIPs)
+
+}
+
+// certRotate Rotate certificate for etcd only and cp only nodes
+func certRotate(product shared.Product, ips []string) {
+
```
fmoral2, line 37: first line shouldnt be empty

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	expectedFileList := getExpectedFileNames()
+	for i := 0; i < len(expectedFileList); i++ {
+		Expect(identicalFileList).To(ContainSubstring(expectedFileList[i]))
+		shared.LogLevel("info", fmt.Sprintf("PASS: Looking for: %s", expectedFileList[i]))
+	}
+}
+
+// compareAndVerifyTLSDirContent Compare TLS Directories before and after cert rotation to display identical files
+func compareAndVerifyTLSDirContent(product shared.Product, ips []string) {
+
```
fmoral2, line 77: first line shouldnt be empty

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	dataDir := fmt.Sprintf("/var/lib/rancher/%s", product)
+	serverDir := fmt.Sprintf("%s/server", dataDir)
+	origTLSDir := fmt.Sprintf("%s/tls", serverDir)
+
+	cmd := fmt.Sprintf("sudo ls -lt %s/ | grep tls | awk {'print $9'} | sed -n '2 p'", serverDir)
+
+	for _, ip := range ips {
+
```
fmoral2, line 85: first line inside a for shouldnt be empty

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+		shared.LogLevel("info", fmt.Sprintf("Working on node with ip: %s", ip))
+
+		// Get new tls directory name
+		tlsDir, tlsError := shared.RunCommandOnNode(cmd, ip)
```
fmoral2, line 89: your variable is pretty clear about what u are doing, no need to comment

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+		if tlsError != nil {
+			shared.ReturnLogError(fmt.Sprintf("Unable to get new TLS Directory name for %s", ip))
+		}
+		shared.LogLevel("info", fmt.Sprintf("TLS Directory name: %s", tlsDir))
+		newTLSDir := fmt.Sprintf("%s/%s", serverDir, tlsDir)
+
+		// Compare original and new tls directories
+		shared.LogLevel("info", "Comparing Directories: %s and %s", origTLSDir, newTLSDir)
```
fmoral2, line 97: https://github.com/rancher/distros-test-framework/pull/42/files#r1397266615
same here
you are already logging "Comparing  stuff...."   no need to comments

pkg/testcase/certificates.go
fmoral2: my example was just an example actually , not an implementation for you ! lol
 thats why i added the comment 
// some logic to split agent and servers ips because here it will be better get from the current cluster state,

if you want a tip, u can create 2 preivous lists with 	 cluster.ServerIP  cluster.AgentIPs , then get one more list from  FetchNodeExternalIP , then compare ips,  then split into 2 newlists  ( finalAgents , finalServers ) 

but tbh, since u are not rebooting nor even deleting,  its not that bad use those that u are using, its just to avoid use globals

shared/aux.go
fmoral2: nice. 
do you have a example output to show us here?

shared/aux.go
```diff
@@ -45,7 +46,7 @@ func RunCommandOnNode(cmd, ip string) (string, error) {
 	if cmd == "" {
 		return "", ReturnLogError("cmd should not be empty")
 	}
-
```
fmoral2, line 48: add a new empty line here please and after the log statement

shared/cluster.go
fmoral2: cmd = fmt.Sprintf("timeout 2m sudo systemctl start --all %s*", product)

but only if you want, if you prefer stick with the function, thats nice also

shared/cluster.go
```diff
@@ -409,3 +409,139 @@ func WriteDataPod(namespace string) (string, error) {
...
+	for _, i := range valid {
+		if p == i {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (p Product) getString() string {
```
fmoral2, line 480: you are creating enums, to bellow have the need to create a method to return it to string again

same for the types

normally we use enums and his respective validations in specific situations to operate in sensitive data, or have a place like source of truth of your growing size of data model, like having 15 products  and actions that will potentially grow or modelling data to use as json payload, dbs , message systems and etc 

So my suggestion would be to not over-engineer on this that is very straightforward stuff , like just pass the desired action with your product to actually use in only one place "ManageClusterService"  
a simple switch or even map statement would be enough for all of those new methods ,enums and validations

shared/cluster.go
fmoral2: https://github.com/rancher/distros-test-framework/pull/42/files#r1397327139

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+func verifyIdenticalFiles(identicalFileList string) {
+	expectedFileList := getExpectedFileNames()
+	for i := 0; i < len(expectedFileList); i++ {
+		Expect(identicalFileList).To(ContainSubstring(expectedFileList[i]))
+		shared.LogLevel("info", fmt.Sprintf("PASS: Looking for: %s", expectedFileList[i]))
+	}
+}
+
+// compareAndVerifyTLSDirContent Compare TLS Directories before and after cert rotation to display identical files
+func compareAndVerifyTLSDirContent(product shared.Product, ips []string) {
```
fmoral2, line 76: lets try shorter and more concise naming, if you are comparing something is implied that u are doing for a areason ( verify whatever ) 

so maybe  "func compareTLSDir(product shared.Product, ips []string) { "
or

"func verifyTLSDir(product shared.Product, ips []string) {"

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+
+// getExpectedFileNames Used for verification of file list
+func getExpectedFileNames() []string {
+	return []string{"client-ca.crt", "client-ca.key", "client-ca.nochain.crt", "client-supervisor.crt",
+		"client-supervisor.key", "peer-ca.crt", "peer-ca.key", "server-ca.crt", "server-ca.key", "request-header-ca.crt",
+		"request-header-ca.key", "server-ca.crt", "server-ca.key", "server-ca.nochain.crt", "service.current.key", "service.key"}
+}
+
+// verifyIdenticalFiles Verify the actual and expected identical file lists match
+func verifyIdenticalFiles(identicalFileList string) {
```
fmoral2, line 67: this func and its usage on "func compareAndVerifyTLSDirContent(product shared.Product, ips []string) {
"
you. should abstract it to compare any file, and receive a "filename" as parameter 

and them both should live on aux.go, to be used anywhere else later

shared/cluster.go
fmoral2: > just trying to create an enum - ensuring unique values and also to have all methods related to k3s/rke2 here.  

also this validation is already done on the runtime   https://github.com/rancher/distros-test-framework/blob/dbc0224ca28dba1bd9567c33a0a89112ed2904bf/config/reader.go#L33C2-L54C1

shared/cluster.go
```diff
@@ -409,3 +409,139 @@ func WriteDataPod(namespace string) (string, error) {
...
+	if action.isStartAction() {
+		return fmt.Sprintf("timeout 2m sudo systemctl --no-block %s %s; sleep 60", action, product.getServiceName(serviceType))
+	}
+	return fmt.Sprintf("timeout 2m sudo systemctl --no-block %s %s", action, product.getServiceName(serviceType))
+}
+
+// ManageClusterService action:stop/start/restart/status product:rke2/k3s ips:ips array for serviceType:agent/server
+func (product Product) ManageClusterService(action ServiceAction, serviceType ServiceType, ips []string) error {
+	if !product.IsValidProduct() {
+		ReturnLogError("product needs to be one of: k3s | rke2")
```
fmoral2, line 513: https://github.com/rancher/distros-test-framework/pull/42/files#r1397405156

no need for this , since in the beginning of your call stack u are loading the product from here , which is loading from  https://github.com/rancher/distros-test-framework/blob/dbc0224ca28dba1bd9567c33a0a89112ed2904bf/config/reader.go#L33C2-L54C1 

Screenshot 2023-11-17 at 11 33 47

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+
+func TestCertRotate() {
+	// Please refer to docs/testing.md for instructions on setup for this test
+
+	cluster := factory.AddCluster(GinkgoT())
+	serverIPs := cluster.ServerIPs
+	agentIPs := cluster.AgentIPs
+	p, _ := shared.GetProduct()
+	var product shared.Product
+	if p == "k3s" {
```
fmoral2, line 21: https://github.com/rancher/distros-test-framework/pull/42/files#r1397409126

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCertRotate() {
+	// Please refer to docs/testing.md for instructions on setup for this test
+
+	cluster := factory.AddCluster(GinkgoT())
+	serverIPs := cluster.ServerIPs
+	agentIPs := cluster.AgentIPs
```
fmoral2, line 18: you shoulnd never omit the err var, just in very rare occasions

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+	p, _ := shared.GetProduct()
```
fmoral2, line 19: you shoulnd never omit the err var, just in very rare occasions

shared/aux.go
aganesh-suse:
```
DEBU[2023-11-17T00:22:24Z]/go/src/github.com/rancher/distros-test-framework/shared/aux.go:314 github.com/rancher/distros-test-framework/shared.LogLevel EXECUTE: sudo ls -lt /var/lib/rancher/rke2/server/ | grep tls | awk {'print $9'} | sed -n '2 p' on 3.149.246.214 
DEBU[2023-11-17T00:22:25Z]/go/src/github.com/rancher/distros-test-framework/shared/aux.go:314 github.com/rancher/distros-test-framework/shared.LogLevel StdOut: tls-1700180360 StdErr:  
```

shared/cluster.go
mdrahman-suse: Then suggest naming them to be in sync with the cli command, like `SystemctlAction`

mdrahman-suse: Or may be even better to have something similar to what we have for `KubectlCommand` func here: https://github.com/rancher/distros-test-framework/blob/main/shared/cluster.go#L139

mdrahman-suse: Similary we can have a `ServiceCommand` func for the k3s/rke2 services

shared/cluster.go
mdrahman-suse: Yeah if we want to create Enums, then lets do it at the start of the cluster creation cause during test run time, it wont change

pkg/testcase/certrotate.go
aganesh-suse: fixed the name of the functions and moved a genric provide filenames actual vs expected given a path to the aux.go. the comparison of tls dir content is very specific to cert rotate - so retaining that portion in the testcase file.

pkg/testcase/certrotate.go
aganesh-suse: removed comments

pkg/testcase/certrotate.go
aganesh-suse: removed comments

shared/aux.go
aganesh-suse: added 👍

pkg/testcase/certrotate.go
aganesh-suse: changed back to start - and added a verify node/pod status test at the end as well. by adding "--no-block" in systemctl cmd, it seems to not hang anymore waiting for the service to come up. @rancher-max - fyi

shared/cluster.go
aganesh-suse: fixed and moved to product.go

shared/cluster.go
aganesh-suse: fixed and moved to product.go

shared/cluster.go
aganesh-suse: fixed and moved to product.go

shared/cluster.go
```diff
@@ -409,3 +409,139 @@ func WriteDataPod(namespace string) (string, error) {
...
+		}
+	}
+
+	return nil
+}
+
+// CertRotate certificate rotate for k3s or rke2
+func (product Product) CertRotate(ips []string) (string, error) {
+	if !product.IsValidProduct() {
+		ReturnLogError("Product needs to be one of: k3s | rke2")
```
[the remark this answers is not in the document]
aganesh-suse: fixed and moved to product.go

pkg/testcase/certrotate.go
aganesh-suse: fixed

pkg/testcase/certrotate.go
aganesh-suse: fixed

pkg/testcase/certrotate.go
aganesh-suse: fixed

pkg/testcase/certrotate.go
aganesh-suse: changed to verifyTLSDirContent

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,109 @@
...
+		shared.ReturnLogError(fmt.Sprintf("Error starting %s service", product))
+	}
+	// Preventative restart - sometimes, start hangs after a cert rotate which gets fixed on a second restart
+	restartError := product.ManageClusterService(shared.Restart, shared.Server, ips)
+	if restartError != nil {
+		shared.ReturnLogError(fmt.Sprintf("Error restarting %s service", product))
+	}
+}
+
+// getExpectedFileNames Used for verification of file list
```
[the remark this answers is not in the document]
aganesh-suse: removed

shared/cluster.go
aganesh-suse: removed all product validations.

pkg/testcase/certrotate.go
aganesh-suse: fixed

pkg/testcase/certrotate.go
aganesh-suse: changed to gomega expects

pkg/testcase/certrotate.go
aganesh-suse: changed to gomega expects

pkg/testcase/certrotate.go
aganesh-suse: fixed

pkg/testcase/certrotate.go
aganesh-suse: fixed

pkg/testcase/certrotate.go
aganesh-suse: fixed

shared/cluster.go
aganesh-suse: i hv removed the enums for the rest and used maps as MD and you suggested in another comment. After our conversation in the standup, moved the product stuff to product.go. this is expected to grow significantly with more actions - for example secret encryption, install, uninstall, and other ipv6 related stuff that MD was talking about in the standup as well.

shared/product.go
```diff
@@ -0,0 +1,103 @@
+package shared
+
+import (
+	"fmt"
+
+	"github.com/rancher/distros-test-framework/config"
+)
+
+// GetProductObject returns the distro "Product" object based on the config file
+func GetProductObject() (Product, error) {
```
fmoral2, line 10: Why do u need a get product if we already have in aux.go?

aganesh-suse: this one returns the Product object with which you can perform operations product.CertRotate etc. the other returns a string.

shared/aux.go
```diff
@@ -352,3 +357,27 @@ func fileExists(files []os.DirEntry, workload string) bool {
 
 	return false
 }
+
+// VerifyFileMatchWithPath verify actual file list matches the expected file list
+func VerifyFileMatchWithPath(actualFileList []string, expectedFileList []string, path string) {
```
fmoral2, line 362: this func should return err , since you have a conditional 
```
equalLength := Expect(len(actualFileList)).To(Equal(len(expectedFileList)))
	if !equalLength {
		LogLevel("error", fmt.Sprintf("actual file and expected file lists size are NOT equal: "+
			"%d vs %d", len(actualFileList), len(expectedFileList)))
	}
```

and the caller deals with it how they want

fmoral2, line 362: func VerifyFileMatchWithPath(actualFileList , expectedFileList []string, path string) {

fmoral2: no need to repeat []string

shared/aux.go
```diff
@@ -352,3 +357,27 @@ func fileExists(files []os.DirEntry, workload string) bool {
...
+	equalLength := Expect(len(actualFileList)).To(Equal(len(expectedFileList)))
```
fmoral2, line 363: on this pkg we do not use gomega or ginkgo as per architectural standars

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,87 @@
...
+
+		shared.LogLevel("info", fmt.Sprintf("TLS Directory name: %s", tlsDir))
+		newTLSDir := fmt.Sprintf("%s/%s", serverDir, tlsDir)
+
+		shared.LogLevel("info", "Comparing Directories: %s and %s", origTLSDir, newTLSDir)
+		cmd2 := fmt.Sprintf("sudo diff -sr %s/ %s/ | grep -i identical | "+
+			"awk '{print $2}' | xargs basename -a | "+
+			"awk 'BEGIN{print \"Identical Files:  \"}; {print $1}'", origTLSDir, newTLSDir)
+
+		idFiles, err := shared.RunCommandOnNode(cmd2, ip)
```
fmoral2, line 80: u should use the same var as the parameter,  or vice verse
idFiles x  func verifyIdenticalFiles(identicalFileList string) {

pkg/testcase/certrotate.go
```diff
@@ -0,0 +1,87 @@
...
+		"client-ca.crt", "client-ca.key", "client-ca.nochain.crt",
+		"client-supervisor.crt", "client-supervisor.key",
+		"dynamic-cert.json",
+		"peer-ca.crt", "peer-ca.key",
+		"server-ca.crt", "server-ca.key",
+		"request-header-ca.crt", "request-header-ca.key",
+		"server-ca.crt", "server-ca.key", "server-ca.nochain.crt",
+		"service.current.key", "service.key"}
+
+	identicalFileList = strings.TrimSpace(identicalFileList)
```
fmoral2, line 53: maybe something like , to give a better name rather then too generic "slice" 
```
	newFileList := strings.Split(strings.TrimSpace(identicalFileList), "\n")

	shared.VerifyFileMatchWithPath(newFileList[1:], expectedFileList, "")
```

shared/product.go
fmoral2: ok , so we should use one or another 

i will add some important points here also for this file:

- if this is kind of a implementation of something ( which is , this is not some aux or helper  functions ) and some methods ( for product ) it should live on /PKG , not here. on pkg should live our internal or custom implementations


- i would not tie all this up to a type string ( product ) this adds no value + you are  coupling more your implementation with this type needing now to pass around all functions as methods (  func ()  name () {}  ) 
and also whenever we are going to use the implementation we would need to import this type as dependencie


- GetSystemCtlCmd and ManageService also on the same subject , in a design perpespecive do not make much sense live in a product implementation , they dont even need.  a product to work , even though needed a product they are generic enough to not make part on the product implementation 

- i really like the ideia of adding a product implementation itself but in here we are adding another function that do the same of the existing one ( getproduct) just return a type instead of string , also we are adding functions not related


summary :

- if u are going to create a implementation should live under /pkg
- u should add those there and refactor and improve if u want 

```

// GetProduct returns the distro product based on the config file
func GetProduct() (string, error) {
	cfgPath, err := EnvDir(".")
	if err != nil {
		return "", ReturnLogError("failed to get config path: %v\n", err)
	}

	cfg, err := config.AddConfigEnv(cfgPath)
	if err != nil {
		return "", ReturnLogError("failed to get config: %v\n", err)
	}
	if cfg.Product != "k3s" && cfg.Product != "rke2" {
		return "", ReturnLogError("unknown product")
	}

	return cfg.Product, nil
}

// GetProductVersion return the version for a specific distro product
func GetProductVersion(product string) (string, error) {
	if product != "rke2" && product != "k3s" {
		return "", ReturnLogError("unsupported product: %s\n", product)
	}
	version, err := getVersion(product + " -v")
	if err != nil {
		return "", ReturnLogError("failed to get version for product: %s, error: %v\n", product, err)
	}

	return version, nil
}

```

OR

keep under aux funcs as you were doing before 

in both cases i would suggest not tie this up with a simple type as string, u are going to create dependencies not needed or not worthy.

shared/product.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			return ip, err
+		}
+	}
+
+	return "", nil
+}
+
+// CertRotate certificate rotate for k3s or rke2
+func (p Product) CertRotate(ips []string) (string, error) {
```
fmoral2, line 93: this should not live on the same level of a product implementantion

shared/product.go
```diff
@@ -0,0 +1,103 @@
...
+	serviceName, err := p.getServiceName(nodeType)
+	if err != nil {
+		return "", ReturnLogError("error getting service name")
+	}
+
+	return fmt.Sprintf("%s %s", sysctlPrefix, serviceName), nil
+}
+
+// ManageService action:stop/start/restart/status product:rke2/k3s ips:ips array for nodeType:agent/server
+func (p Product) ManageService(action, nodeType string, ips []string) (string, error) {
```
fmoral2, line 77: https://github.com/rancher/distros-test-framework/pull/42/files#r1403534257

aganesh-suse: this is an action you perform on the k3s/rke2 product - why should it not live on the same level as a product implementation? 
where do u suggest this be? 
if we dont have a product.actions -> then the whole implementation of this is moot and we can simply keep the static methods way of implementation.

shared/product.go
aganesh-suse: this is an action you perform on the k3s/rke2 product - why should it not live on the same level as a product implementation? 
where do u suggest this be? 
if we dont have a product.actions -> then the whole implementation of this is moot and we can simply keep the static methods way of implementation. is this something you would prefer instead?

shared/product.go
aganesh-suse: per our conversation, moved the GetProduct and GetProductVersion methods to product.go as well
changed all of these to static methods for now. - passing the product var as a string. 
if we want to make this an implementation in pkg, we will address this in the future in another PR and keep this short and simple.

shared/product.go
aganesh-suse: per our conversation, moved the GetProduct and GetProductVersion methods to product.go as well
changed all of these to static methods for now. - passing the product var as a string.
if we want to make this an implementation in pkg, we will address this in the future in another PR and keep this short and simple.

shared/product.go
aganesh-suse: per our conversation, moved the GetProduct and GetProductVersion methods to product.go as well
changed all of these to static methods for now. - passing the product var as a string.
if we want to make this an implementation in pkg, we will address this in the future in another PR and keep this short and simple.

pkg/testcase/certrotate.go
aganesh-suse: fixed

pkg/testcase/certrotate.go
aganesh-suse: fixed

shared/aux.go
aganesh-suse: fixed

shared/aux.go
aganesh-suse: fixed

aganesh-suse: fixed

shared/product.go
fmoral2: nice, tk you so much for your effort

shared/product.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return cfg.Product, nil
+}
+
+// GetProductVersion return the version for a specific distro product
+func GetProductVersion(product string) (string, error) {
+	if product != "rke2" && product != "k3s" {
+		return "", ReturnLogError("unsupported product: %s\n", product)
+	}
+	version, err := getVersion(product + " -v")
```
fmoral2, line 32: please only missed this getVersion method to move in here also

aganesh-suse: moved 🙇

fmoral2:
only missing now the jenkins job PR!

fmoral2 (approved):

aganesh-suse:
> only missing now the jenkins job PR!

and the k3s split role changes

mdrahman-suse:
> > only missing now the jenkins job PR!
> 
> and the k3s split role changes

This should be a separate PR

VestigeJ (approved):
Looks great

shared/product.go
```diff
@@ -0,0 +1,119 @@
...
+
+import (
+	"fmt"
+
+	"github.com/rancher/distros-test-framework/config"
+)
+
+// GetProduct returns the distro product based on the config file
+func GetProduct() (string, error) {
+	cfgPath, err := EnvDir(".")
```
fmoral2, line 11: .  to "shared"

fmoral2 (approved):

rancher-max (approved):

PR #46, opened 2023-11-24
feat: add cluster restart tests

fmoral2:
#### Testing ####

Checklist:
- https://github.com/rancherlabs/jenkins-job-builder/pull/337/files 

- ok 

- ok 

- ok 
 - ok 

- ok

### DIFF
--- a/modules/install/join_k3s_agent.sh
+++ b/modules/install/join_k3s_agent.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 ## Uncomment the following lines to enable debug mode
-set -x
-PS4='+(${LINENO}): '
-set -e
-trap 'echo "Error on line $LINENO: $BASH_COMMAND"' ERR
+# set -x
+# PS4='+(${LINENO}): '
+# set -e
+# trap 'echo "Error on line $LINENO: $BASH_COMMAND"' ERR
 
 create_config() {
   mkdir -p /etc/rancher/k3s
--- a/modules/install/join_k3s_master.sh
+++ b/modules/install/join_k3s_master.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 ## Uncomment the following lines to enable debug mode
-set -x
- PS4='+(${LINENO}): '
-set -e
-trap 'echo "Error on line $LINENO: $BASH_COMMAND"' ERR
+# set -x
+#  PS4='+(${LINENO}): '
+# set -e
+# trap 'echo "Error on line $LINENO: $BASH_COMMAND"' ERR
 
 
 create_directories() {
--- a/modules/install/join_rke2_agent.sh
+++ b/modules/install/join_rke2_agent.sh
@@ -2,10 +2,10 @@
 # This script is used to join one or more nodes as agents
 echo "$@"
 
-set -x
-PS4='+(${LINENO}): '
-set -e
-trap 'echo "Error on line $LINENO: $BASH_COMMAND"' ERR
+# set -x
+# PS4='+(${LINENO}): '
+# set -e
+# trap 'echo "Error on line $LINENO: $BASH_COMMAND"' ERR
 
 node_os=$1
 server_ip=$2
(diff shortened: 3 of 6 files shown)


pkg/testcase/restartservice.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func TestRestartService() {
+	delay := time.After(30 * time.Second)
+	product, err := shared.GetProduct()
+	Expect(err).NotTo(HaveOccurred(), "failed to get product")
+	c := factory.AddCluster(GinkgoT())
+
+	for _, ip := range c.ServerIPs {
+		cmd, err := GetSystemCtlCmd(product, "restart", "server")
```
aganesh-suse, line 20: we could directly do the shared.ManageService  call - which does the getsytemctl+run restart cmd, for you - it ll get the systemctl cmd and run the restart service as well. any errors will be caught and logged internally as well. 
```
// ManageService action:stop/start/restart/status product:rke2/k3s ips:ips array for nodeType:agent/server
func ManageService(product, action, nodeType string, ips []string) (string, error) {
	for _, ip := range ips {
		cmd, getError := GetSystemCtlCmd(product, action, nodeType)
		if getError != nil {
			return ip, getError
		}
		_, err := RunCommandOnNode(cmd, ip)
		if err != nil {
			return ip, err
		}
	}

	return "", nil
}
```

entrypoint/restartservice/restartservice_test.go
```diff
@@ -0,0 +1,75 @@
...
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus(),
+		)
+	})
+
+	It("Restart server and agent service", func() {
+		testcase.TestRestartService()
+	})
+
```
aganesh-suse, line 36: Re verifying the node and pod status here -after the restart, should be enough to ensure the cluster is actually up - before testing the cluster ip and other stuff -> this could be added here instead of adding a simple delay at the end of the restart test - without doing any verification.

pkg/testcase/restartservice.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	for _, ip := range c.AgentIPs {
+		cmd, err := GetSystemCtlCmd(product, "restart", "agent")
+		Expect(err).NotTo(HaveOccurred(), "failed to get systemctl command")
+		_, err = shared.RunCommandOnNode(cmd, ip)
+		Expect(err).NotTo(HaveOccurred(), "failed to restart agent service")
+	}
+	// TODO: add validation for delays(maybe implement k8s api client)
+	<-delay
```
aganesh-suse, line 33: https://github.com/rancher/distros-test-framework/pull/46/files#r1404563463

entrypoint/restartservice/restartservice_test.go
fmoral2: idk if i totally got your point, but if i do , just adding the tests after fails, because need some sec to cluster come back.

pkg/testcase/restartservice.go
fmoral2: truee that ! tk you

entrypoint/restartservice/restartservice_test.go
aganesh-suse: after restart step, add these 2 tests again:
```
It("Validate Nodes", func() {
		testcase.TestNodeStatus(
			assert.NodeAssertReadyStatus(),
			nil,
		)
	})

	It("Validate Pods", func() {
		testcase.TestPodStatus(
			assert.PodAssertRestart(),
			assert.PodAssertReady(),
			assert.PodAssertStatus(),
		)
	})
```

the validation test for this, is in a loop, waiting for the node and pod status to come up.

fmoral2: got youu!!!   tk you

pkg/testcase/node.go
```diff
@@ -36,7 +36,17 @@ func TestNodeStatus(
 				nodeAssertVersion(g, node)
 			}
 		}
-	}, "1500s", "20s").Should(Succeed())
+		return true
```
fmoral2, line 39: ,

PR #47, opened 2023-11-27
K3s split role

aganesh-suse:
#### Proposed Changes ####

Adding support to be able to deploy k3s with split role. 
Changes to terraform files.
Added k3s_node_role.sh script. 
minor update to k3s_master.sh for the etcd only node as master node scenario. 
updates to files in modules/k3s:
instances_server.tf file
main.tf
variables.tf

#### Testing ####

Checklist:

done

docs examples already have k3s examples. nothing additional needed

n/a

done

n/a


modules/install/k3s_node_role.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
```
mdrahman-suse, line 1: It seems that this file is quite/almost similar to `rke2_node_role.sh`. Can we explore if that script can be repurposed for k3s as well and have a common `node_roles.sh` file?

mdrahman-suse (commented):
Just that one comment, else LGTM

modules/k3s/master/instances_server.tf
```diff
@@ -337,4 +357,28 @@ resource "aws_route53_record" "aws_route53" {
...
+  }
+  provisioner "local-exec" {
+    command    = "test -f /tmp/${var.resource_name}_control_plane_${count.index} && grep '6444' /tmp/${var.resource_name}_config && sed s/127.0.0.1:6444/\"${count.index == 0 ? local.serverIp : aws_instance.master2-ha[count.index - 1].public_ip}:6443\"/g /tmp/${var.resource_name}_config >/tmp/${var.resource_name}_kubeconfig"
+    on_failure = continue
+  }
+
+  provisioner "local-exec" {
+    command    = "test -f /tmp/${var.resource_name}_control_plane_${count.index} && grep '6443' /tmp/${var.resource_name}_config && sed s/127.0.0.1:6443/\"${count.index == 0 ? local.serverIp : aws_instance.master2-ha[count.index - 1].public_ip}:6443\"/g /tmp/${var.resource_name}_config >/tmp/${var.resource_name}_kubeconfig"
+    on_failure = continue
+  }
```
rancher-max, line 382: Is it required to replace both the 6444 and the 6443 here? The 6444 is the etcd node right? So technically that part should never get hit because it will fail `test -f /tmp/${var.resource_name}_control_plane_${count.index}` I think?

rancher-max: The answer is that yes it is required 😄 Thank you for explaining this in Slack huddle! To try to summarize:
- The config file is written ONCE to `/tmp/#{var.resource_name}_config`. This happens on only the first node. In the case that this first node has cp components, then it will also have port 6443. In the case that it does NOT, it will have port 6444.
- Using the Kubeconfig externally to the cluster requires pointing to a cp node with port 6443.
- Therefore, we are always updating the kubeconfig to use the cp node IP and port 6443, whether it started with an non-cp node first or not.

modules/install/k3s_node_role.sh
aganesh-suse: fixed

VestigeJ (approved):

endawkins (approved):

fmoral2 (approved):

PR #48, opened 2023-11-30
fix for rhel error

aganesh-suse:
#### Proposed Changes ####

fix for the rhel error

changing *"rhel"* to "rhel" - so in other cases, the disable cloud setup service etc dont execute.
 

#### Types of Changes ####

minor script changes for bug fix

#### Testing ####

Checklist:
will do soon

done

n/a

n/a

will do

will do

### DIFF
--- a/modules/install/join_k3s_agent.sh
+++ b/modules/install/join_k3s_agent.sh
@@ -50,7 +50,7 @@ rhel() {
 disable_cloud_setup() {
    local node_os="${1}"
 
-if  [[ "$node_os" = *"rhel"* ]] || [[ "$node_os" = "centos8" ]]
+if  [[ "$node_os" = "rhel" ]] || [[ "$node_os" = "centos8" ]]
   then
     NM_CLOUD_SETUP_SERVICE_ENABLED=$(systemctl status nm-cloud-setup.service | grep -i enabled)
     NM_CLOUD_SETUP_TIMER_ENABLED=$(systemctl status nm-cloud-setup.timer | grep -i enabled)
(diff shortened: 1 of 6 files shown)


aganesh-suse:
> So in this case we have to just pass `rhel` or `centos8` directly now? We can't do `rhel8` any longer?

no we can - it just wont run the disable cloud section if we give rhel8 i believe. for ex: rke2 rocky - rhel8 - wont need this section run. @fmoral2 please feel free to correct me if am wrong.

fmoral2:
> So in this case we have to just pass `rhel` or `centos8` directly now? We can't do `rhel8` any longer?

please look at the jenkins jobs on slack!


we are just disabling cloud setup only when its needed (. OSs.  node_os == rhel and centos8)  all other node_os names dont need to disable it.

rancher-max:
I believe those jobs used amis that already had it disabled. It still needs to be disabled for all rhel setups. The issue here is that passing just `rhel` will also then force this to call the rhel7 specific stuff:
```
  if [ "$node_os" = "rhel" ]
    then
      subscription-manager register --auto-attach --username="$username" --password="$password"
      subscription-manager repos --enable=rhel-7-server-extras-rpms
  fi
```

which is not what we want for rhel8+

fmoral2:
> I believe those jobs used amis that already had it disabled. It still needs to be disabled for all rhel setups. The issue here is that passing just `rhel` will also then force this to call the rhel7 specific stuff:
> 
> ```
>   if [ "$node_os" = "rhel" ]
>     then
>       subscription-manager register --auto-attach --username="$username" --password="$password"
>       subscription-manager repos --enable=rhel-7-server-extras-rpms
>   fi
> ```
> 
> which is not what we want for rhel8+

i have tested with other images, the images from job it was me adding
i also tested locally with other ones
but should we gather more amis and test it  ? 
it would be nice

fmoral2 (approved):

rancher-max (approved):

PR #51, opened 2023-12-07
Use regex to parse pods on multiple spaces

rancher-max:
#### Proposed Changes ####

Instead of using `strings.Fields` to parse the pods, which parses based off a single whitespace, this is using a regex that parses based off 2 or more spaces. I'm not 100% sure if there's ever less than 2 spaces, but in my tests I found this to work consistently.

#### Types of Changes ####

bugfix

#### Testing ####

I tested locally a few times using the certrotate job in rke2, where this was consistently reproducible.

### DIFF
--- a/shared/cluster.go
+++ b/shared/cluster.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


fmoral2 (approved):

ShylajaDevadiga (approved):
LGTM

PR #56, opened 2023-12-08
Update ingressroute and servicelb tests to use new GetNodesByRoles function

rancher-max:
#### Proposed Changes ####

This adds a new function to GetNodesByRoles. Essentially, you can pass in `etcd`, `control-plane`, and/or `worker` to return a slice of Node objects that satisfy those conditions. There is a new label passed on the nodes called `role-<role>=true`, where <role> is any of the aforementioned roles. This has to be passed in this way because the kubelet rejects slashes and other values that fit with a standard "node role". 

This new function is useful in both IngressRoute tests and ServiceLB tests. In IngressRoute, we're retrieving worker nodes specifically. In ServiceLB, it only works on control-plane and worker nodes, so we that's where we should run the curl commands on (instead of all nodes. There's a bug that this fixes where this would fail with etcd-only nodes). 

#### Types of Changes ####

bugfix

#### Testing ####

Validated locally using split role setup and standard setup


VestigeJ (approved):
Great pairing time for the team :)

pkg/testcase/ingressdns.go
```diff
@@ -71,7 +71,11 @@ func TestDnsAccess(applyWorkload, deleteWorkload bool) {
 }
 
 func TestIngressRoute(applyWorkload, deleteWorkload bool, apiVersion string) {
-	publicIp := fmt.Sprintf("%s.nip.io", shared.FetchNodeExternalIP()[0])
+	workerNodes, err := shared.GetNodesByRoles("worker")
+	Expect(workerNodes).NotTo(BeEmpty())
+	Expect(err).NotTo(HaveOccurred())
+	publicIp := fmt.Sprintf("%s.nip.io", workerNodes[0].ExternalIP)
+
 	if applyWorkload {
```
fmoral2, line 79: can we maintain the pattern and have the apply workload check first? because if the apply fail first its better

shared/cluster.go
```diff
@@ -305,6 +305,30 @@ func GetNodes(print bool) ([]Node, error) {
 	return nodes, nil
 }
 
+// GetNodesByRoles takes in one or multiple node roles and returns the slice of nodes that have those roles
+// Valid values for roles are: etcd, control-plane, worker
+func GetNodesByRoles(roles ...string) ([]Node, error) {
```
fmoral2, line 310: this is awesome!

shared/cluster.go
```diff
@@ -305,6 +305,30 @@ func GetNodes(print bool) ([]Node, error) {
...
+			" -l role-" + role
+		res, err := RunCommandHost(cmd)
+		if err != nil {
+			return nil, err
+		}
+		matchedNodes = append(matchedNodes, parseNodes(res)...)
+	}
+
+	for _, matchedNode := range matchedNodes {
+		nodes = AppendIfMissing(nodes, matchedNode)
```
fmoral2, line 326: have u created this function only to avoid a nested loop?

shared/aux.go
```diff
@@ -123,6 +123,16 @@ func PrintBase64Encoded(path string) error {
 	return nil
 }
 
+// AppendIfMissing appends a value to a slice if that value does not already exist in the slice.
+func AppendIfMissing(slice []Node, i Node) []Node {
```
fmoral2, line 127: if your slice is type of []Node , lets give a more meaningful name to it ... maybe nodes ? idk

pkg/testcase/ingressdns.go
```diff
@@ -99,7 +102,7 @@ func TestIngressRoute(applyWorkload, deleteWorkload bool, apiVersion string) {
 
 	getIngressRoutePodsRunning := fmt.Sprintf("kubectl get pods -n test-ingressroute -l app=whoami"+
 		" --field-selector=status.phase=Running --kubeconfig=%s", shared.KubeConfigFile)
-	err := assert.ValidateOnHost(getIngressRoutePodsRunning, statusRunning)
+	err = assert.ValidateOnHost(getIngressRoutePodsRunning, statusRunning)
 	Expect(err).NotTo(HaveOccurred(), err)
 
 	// Query the IngressRoute Host
```
fmoral2, line 108: since we are touching here again and i forgot this on the original PR, can we extract here unto line 133 into a separated func?

shared/aux.go
rancher-max: Ah this is a great point... and I think also touches on this comment too:
> do u intend to export this func outside the pkg ?

because initially I was thinking "hey we can use this for any slice we want to append to!" forgetting entirely that the func signature is for Node. Soooo yeah I'll change the name and remove the export actually in this case!

pkg/testcase/ingressdns.go
rancher-max: ah for this one I needed to modify for this first because the publicIp is required during the apply, since there's a lot of file rewrite logic in there to move it from the template to the actual applied file with the IPs

fmoral2: got you
tks!

shared/cluster.go
rancher-max: Not really since the function itself is a loop, this still ends up being a nested loop which sucks. I played around for way too much time with people trying to:
1. Change the kubectl command itself to return everything in one (fails)
2. Use a map (still ends up needing a nested loop)
3. Brainstorm/search for some other solution

This ended up being just the most simple and not too time complex still we're dealing with such small datasets (less than 10 in pretty much all cases)

fmoral2: nice, tk you for the explanation

fmoral2 (approved):

fmoral2 (approved):

aganesh-suse (approved):
LGTM. This was a great peer programming session 👍

PR #60, opened 2023-12-13
Update install scripts and modules for dual stack support

mdrahman-suse:
#### Proposed Changes ####

- Update install scripts and terraform files in modules to support dual stack
- Replace local vars with global vars in install scripts
- Organize and Add consistency across install scripts for the products 

#### Types of Changes ####

- Refactor for dual stack tests

#### Testing ####

- Run through existing jobs on the branch to ensure no regression is observed

Checklist:
- Ok

- Ok

- N/A

- N/A

- Ok

- Ok

#### Linked Issues ####

https://github.com/rancher/qa-tasks/issues/882

#### Further Comments ####

GH Actions for linting: https://github.com/mdrahman-suse/distros-test-framework/actions

### DIFF
--- a/modules/install/join_k3s_agent.sh
+++ b/modules/install/join_k3s_agent.sh
@@ -6,6 +6,7 @@ PS4='+(${LINENO}): '
 set -e
 trap 'echo "Error on line $LINENO: $BASH_COMMAND"' ERR
 
+# Script args
 node_os=${1}
 server_ip=${2}
 token=${3}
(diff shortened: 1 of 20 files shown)


modules/install/join_k3s_agent.sh
```diff
@@ -65,36 +84,26 @@ if [[ "$node_os" = *"rhel"* ]] || [[ "$node_os" = "centos8" ]] || [[ "$node_os"
    fi
 }
```
fmoral2, line 86: to much empty lines here

rancher-max (dismissed):
I know it was a bit of a pain, but I appreciate the Terraform stuff separated out like this. It makes the review a lot easier. As long as there are no regressions with validate and upgrade then I'm totally fine with this, pending others' reviews. Thank you! LGTM

fmoral2:
can we add a function to install helm on nodes? it would be good already have that installed to be able to send cmds throught the template cli, that needs it installed.

mdrahman-suse:
> can we add: ?
> 
> * a function to install helm on nodes? it would be good already have that installed to be able to send cmds throught the template cli, that needs it installed.
> * also since we are refactoring we need to add an exit if install fails and it would be nice to print the journalctl logs !
>   something like
> 
> ```
> 
> 
> if [[ "$version" == *"v1.18"* ]] || [[ "$version" == *"v1.17"* ]] && [[ -n "$worker_flags" ]]
>   then
>    if !  curl -sfL https://get.k3s.io | sh -s - "$install_mode" --node-external-ip="$ip" --server https://"$server_ip":6443 --token "$token"
> echo "fail"
> journalctl --no pager
> 
> exit 1 
> ```

- I like the idea of adding helm on the nodes, I actually thought about using a rancher installation script on the cluster, based on some flag... but lets keep this PR minimal and we can work on the going through helm/rancher installation as a separate PR
- I like the idea to fail the test on failed install, can raise a separate PR for that too as it would require some additional testing

fmoral2:
> > can we add: ?
> > 
> > * a function to install helm on nodes? it would be good already have that installed to be able to send cmds throught the template cli, that needs it installed.
> > * also since we are refactoring we need to add an exit if install fails and it would be nice to print the journalctl logs !
> >   something like
> > 
> > ```
> > 
> > 
> > if [[ "$version" == *"v1.18"* ]] || [[ "$version" == *"v1.17"* ]] && [[ -n "$worker_flags" ]]
> >   then
> >    if !  curl -sfL https://get.k3s.io | sh -s - "$install_mode" --node-external-ip="$ip" --server https://"$server_ip":6443 --token "$token"
> > echo "fail"
> > journalctl --no pager
> > 
> > exit 1 
> > ```
> 
>     * I like the idea of adding helm on the nodes, I actually thought about using a rancher installation script on the cluster, based on some flag... but lets keep this PR minimal and we can work on the going through helm/rancher installation as a separate PR
> 
>     * I like the idea to fail the test on failed install, can raise a separate PR for that too as it would require some additional testing

ok sir!

fmoral2 (dismissed):

fmoral2 (approved):

aganesh-suse (approved):
LGTM. This is some great refactoring :) Appreciate you doing this. Thanks.

PR #63, opened 2023-12-14
fix logger call

fmoral2:
#### Proposed Changes ####

- change logger to only log call stack when level is not INFO or DEBUG
Screenshot 2023-12-14 at 10 23 55
Screenshot 2023-12-14 at 10 43 10

- fix return log error to add link to error it self

#### Testing ####

Checklist:

#### Further Comments ####

Screenshot 2023-12-14 at 10 43 14

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -2,7 +2,6 @@ package logger
 
 import (
 	"fmt"
-	"os"
 	"runtime"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


mdrahman-suse (approved):
LGTM 👍

VestigeJ (approved):

pkg/logger/logger.go
```diff
@@ -14,41 +13,45 @@ var (
...
 
 	return le
 }
 
-func newLogger(json bool) *log.Logger {
+func newLogger() *log.Logger {
 	logger := log.New()
+	logger.SetFormatter(customFormatter())
+	logger.SetReportCaller(true)
+	logger.SetLevel(log.DebugLevel)
```
aganesh-suse, line 29: Curious thought: and by no means this is blocking merging of this PR. 
is there a way, we can set the LogLevel var to 'debug' or 'info' in the tfvars... and then modify the level here - based on the same. 
that way, we can add more debug logging - and not have it logged if it was info level during test run.

fmoral2: like have a var on tf.vars ?

level_log = {desired level}  ? 

if yes, yes!

lol

but i would not recommend tie this up with any config env, especially tf

PR #64, opened 2023-12-14
Add fix for race condition during rpm installs

rancher-max:
#### Proposed Changes ####

There is currently a race condition that prevents the GPG key in an rke2 install from successfully being signed in some cases.

#### Types of Changes ####

bugfix

#### Testing ####

Ran a jenkins job that was always failing previously

Checklist:

### DIFF
--- a/modules/install/join_rke2_agent.sh
+++ b/modules/install/join_rke2_agent.sh
@@ -91,6 +91,8 @@ then
     fi
     systemctl restart systemd-sysctl
 fi
-yum install tar iptables -y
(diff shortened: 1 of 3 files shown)


VestigeJ (approved):

aganesh-suse (approved):
LGTM

mdrahman-suse:
Do we need to add this k3s as well? @rancher-max

rancher-max:
> Do we need to add this k3s as well, to be consistent? @rancher-max

Nope, k3s doesn't have any of this anyway so we should be fine leaving it out!

rancher-max:
image
Success!

ShylajaDevadiga (approved):

PR #67, opened 2023-12-19
Make kubeconfig explicit during k3s initial server check

rancher-max:
#### Proposed Changes ####

bugfix

#### Testing ####

Checklist:

### DIFF
--- a/modules/install/k3s_master.sh
+++ b/modules/install/k3s_master.sh
@@ -116,7 +116,7 @@ wait_nodes() {
   timeElapsed=0
 
   while (( timeElapsed < 1200 )); do
-    if kubectl get nodes >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


mdrahman-suse:
Do we need to add the same for rke2?

aganesh-suse (approved):

endawkins (approved):

mdrahman-suse (commented):
Also can we define a global var like `kubeconfigpath=/etc/rancher/k3s/k3s/yaml` for this so that its not repetitive?

aganesh-suse:
> Do we need to add the same for rke2?



> Do we need to add the same for rke2?

no other shell script seems to use the kubectl command.

rancher-max:
@mdrahman-suse RKE2 doesn't have this check here so we should be fine on that. We can define a global var but it's just three spots in this one file so I thought maybe this would be fine. I'll do whichever you prefer 😇

mdrahman-suse:
LGTM then

PR #74, opened 2024-02-26
 Version bump updates

fmoral2:
#### Types of Changes ####
 
- debug mode added on template
- fixed 2 issues related to concurrency  
- Here we was unnecessaryly spinning up routines based on CMDS , which with a lot of commands and more than 1 ip, started to flood ssh and returning error for handshake 
* src/github.com/rancher/distros-test-framework/shared/aux.go:57 ERRO[2024-02-21T15:03:58Z] ssh: handshake failed: EOF Last call: github.com/rancher/distros-test-framework/shared.configureSSH in file:/go/src/github.com/rancher/distros-test-framework/shared/aux.go:182 ERRO[2024-02-21T15:03:58Z] ssh: handshake failed: EOF

https://github.com/rancher/distros-test-framework/blob/3311bc26a363a4418f3a05eeafebc032d2fb49e9/pkg/template/processor.go#L36 

 - fixed an issue here https://github.com/rancher/distros-test-framework/blob/3311bc26a363a4418f3a05eeafebc032d2fb49e9/pkg/template/processor.go#L67
which was sending a cmd that returns  the node info not parsed correctly , causing  to pass by this info to next function returning error related to not finding the host 

- fixed general error formatting wrapper  on the PKG 
- added a job that runs most of all general bumps tested
 
- fixed expectedValue result formatting 

#### Testing ####

Checklist:
yes - 

yes - 

yes- 

yes 

yes 

OS TESTED

Red Hat Enterprise Linux 8.9 (Ootpa)

Red Hat Enterprise Linux  9.2 

 Rocky Linux 8.9 (Green Obsidian)

 Oracle Linux Server 9.3 
ORACLE 8.9

SUSE Linux Enterprise Server 15 SP4

ubuntu


mdrahman-suse (dismissed):
LGTM, one suggestion though, not sure if it will be bit overhead but I wanted to suggest to have the commands in a file so that they can be maintained separately and can be reused in other tests as needed, thoughts? 
For ex: commands in this file https://github.com/rancher/distros-test-framework/blob/bac6b325f52ae7b162c70ee7090745aa1b7f45f6/entrypoint/versionbump/componentsbump_test.go

fmoral2:
> LGTM, one suggestion though, not sure if it will be bit overhead but I wanted to suggest to have the commands in a file so that they can be maintained separately and can be reused in other tests as needed, thoughts? For ex: commands in this file https://github.com/rancher/distros-test-framework/blob/bac6b325f52ae7b162c70ee7090745aa1b7f45f6/entrypoint/versionbump/componentsbump_test.go


Hey i have used in this way here since they are all used in the same file

But having a separated file for that we need to discuss a bit further , since normally in go we tend to use the consts and var very near to its usage, to increase readability and maintenance 

so depending for which commands are you saying , maybe the " cleaner " and maintainable way to use would be having a function that holds your used cmd options, and then u just call it and construct your command accordingly to each need ( in your consumer ) 
so in that way we are  "centralizing "  the cmd creation and maintenance and not creating a whole file for that 
https://github.com/rancher/qa-tasks/issues/967

https://github.com/rancher/distros-test-framework/blob/2ff8ae8ad52acf8d6fb3347f4b6bc57a59b1ada8/shared/cluster.go#L139C1-L139C5


BUT i also think that is not a bad idea , specially if the commands **are not exported**, which i dont think it will be the case , since we use cmds, through  out the whole project 

so looking at our use case and how things work on the framework x coding best practices on declaration , i would say the good approach would be using stuff like the function above on the file pasted 

BUT thats my POV, we should  definitely discuss that with team

shared/cluster.go
```diff
@@ -233,14 +234,14 @@ func FetchServiceNodePort(namespace, serviceName string) (string, error) {
...
+		return "", ReturnLogError("failed to fetch service node port: %w", err)
 	}
 
 	return nodeport, nil
 }
 
-// FetchNodeExternalIP returns the external IP of the nodes.
-func FetchNodeExternalIP() []string {
+// FetchNodeExternalIPs returns the external IP of the nodes.
+func FetchNodeExternalIPs() []string {
```
mdrahman-suse, line 244: Did we update this method in the tests where this is being used?

mdrahman-suse: Ok I see we did

mdrahman-suse (dismissed):
LGTM

VestigeJ (dismissed):

entrypoint/versionbump/componentsbump_test.go
```diff
@@ -0,0 +1,177 @@
...
+						},
+						{
+							Cmd: "sudo sed -i '/cni:/d' /etc/rancher/rke2/config.yaml && " +
+								" sudo sed -i '/multus/d' /etc/rancher/rke2/config.yaml && sudo sed -i '/canal/d' /etc/rancher/rke2/config.yaml " +
+								" && echo -e \"cni: cilium\" | sudo tee -a /etc/rancher/rke2/config.yaml " +
+								" && cat /etc/rancher/rke2/config.yaml",
+							ExpectedValue: "cilium",
+						},
+						{
+							Cmd:           "sudo sed -i '/^$/d' /etc/rancher/rke2/config.yaml && cat /etc/rancher/rke2/config.yaml && sleep 20",
```
ShylajaDevadiga, line 109: Do we need sleep 20 here for sed?

entrypoint/versionbump/componentsbump_test.go
```diff
@@ -0,0 +1,177 @@
...
+							ExpectedValue: "cilium",
+						},
+						{
+							Cmd:           "sudo systemctl restart rke2-server && if [ $? -eq 0 ]; then echo \"ok\"; fi",
+							ExpectedValue: "ok",
+						},
+						{
+							Cmd:           "sleep 10 && if [ $? -eq 0 ]; then echo \"ok\"; fi",
+							ExpectedValue: "ok",
```
ShylajaDevadiga, line 118: Could we add sleep with the restart cmd instead of having it as a separate cmd?

entrypoint/versionbump/versionbump_suite_test.go
```diff
@@ -49,19 +49,18 @@ func TestMain(m *testing.M) {
...
+	if customflag.ServiceFlag.TestConfig.DebugMode == true {
+		shared.LogLevel("info", "debug mode enabled on template\n\n")
 	}
 
 	os.Exit(m.Run())
-
 }
 
-func TestVersionTestSuite(t *testing.T) {
+func TestBumpVersionSuite(t *testing.T) {
```
ShylajaDevadiga, line 63: Would prefer TestVersionBumpSuite instead

modules/install/rke2_master.sh
```diff
@@ -4,8 +4,8 @@
 
 echo "$@"
 # the following lines are to enable debug mode
-set -x
-PS4='+(${LINENO}): '
+#set -x
```
ShylajaDevadiga, line 7: Let's remove it if it is not needed.

entrypoint/versionbump/componentsbump_test.go
fmoral2: probably we dont , but i will double check

entrypoint/versionbump/componentsbump_test.go
fmoral2: this one i would like to keep separate just to make more clear, like if we have any error from restart we will know thats from there , rather than go to logs to find out if the "  nok " answer is from restart or from some other network/connection issue that might happen when trying to sleep 

but its just to separate concerns nothing else .

modules/install/rke2_master.sh
fmoral2: fixed tks

entrypoint/versionbump/componentsbump_test.go
fmoral2: apparently i have added because after the sed, if you go try to restart right after it kind of stucks there
-- without sleep 
https://jenkins.int.rancher.io/job/rke2-tests/job/rke2-all-version-bump-validation/11/console

-with sleep 
https://jenkins.int.rancher.io/job/rke2-tests/job/rke2-all-version-bump-validation/12/console

fmoral2: @ShylajaDevadiga

VestigeJ (approved):

mdrahman-suse (dismissed):
LGTM, will wait for Shylaja's approval as she had few comments

entrypoint/versionbump/componentsbump_test.go
ShylajaDevadiga: It seems to have failed on systemctl restart rke2-server and the test times out.

entrypoint/versionbump/componentsbump_test.go
```diff
@@ -0,0 +1,177 @@
...
+								" && echo -e \"cni: cilium\" | sudo tee -a /etc/rancher/rke2/config.yaml " +
+								" && cat /etc/rancher/rke2/config.yaml",
+							ExpectedValue: "cilium",
+						},
+						{
+							Cmd:           "sudo sed -i '/^$/d' /etc/rancher/rke2/config.yaml && cat /etc/rancher/rke2/config.yaml && sleep 20",
+							ExpectedValue: "cilium",
+						},
+						{
+							Cmd:           "sudo systemctl restart rke2-server && if [ $? -eq 0 ]; then echo \"ok\"; fi",
```
ShylajaDevadiga, line 113: Could we add sleep here instead?

modules/install/join_rke2_master.sh
```diff
@@ -97,6 +97,53 @@ cis_setup() {
   fi
 }
 
+etcd_download() {
+    sudo curl -L https://github.com/etcd-io/etcd/releases/download/v3.5.0/etcd-v3.5.0-linux-amd64.tar.gz -o etcd-v3.5.0-linux-amd64.tar.gz
+    sudo tar xzvf etcd-v3.5.0-linux-amd64.tar.gz -C /usr/local/bin --strip-components=1 etcd-v3.5.0-linux-amd64/etcdctl
```
rancher-max, line 102: Can we have the etcd version in a variable so that it's easier to update? (in all the places where we're doing this)

modules/install/join_rke2_master.sh
```diff
@@ -97,6 +97,53 @@ cis_setup() {
...
+               sleep 20
+             done
+    else
+        echo "Certificate files not found, skipping etcdctl endpoint health check."
+    fi
+else
+    echo "Installation failed or etcdctl not found in PATH."
+    return 1
+fi
+}
```
rancher-max, line 145: Do we need this to run on every install? This feels like it'll be good in the etcd version test, but outside of that, I don't think we need to do these validations during the install here as part of this script. Just having the sleep alone will slow things down.

modules/install/join_rke2_master.sh
```diff
@@ -132,6 +179,7 @@ main() {
   subscription_manager
   disable_cloud_setup
   install
+  install_etcdctl
```
rancher-max, line 182: We should only install etcdctl on etcd nodes. This will currently run on cp-only nodes as well if that is passed in the configuration.

modules/install/k3s_master.sh
```diff
@@ -195,16 +242,17 @@ main() {
...
+    # etcd only node needs api server to come up fully, which is in control plane node.
+    # and hence we cannot wait for node/pod status in this case.
     wait_nodes
     wait_ready_nodes
     wait_pods
   else
     # add sleep to make sure install finished and the node token file is present on the node for a copy
     sleep 30
   fi
+  install_etcdctl
```
rancher-max, line 255: For k3s, we should also only install etcdctl when using an etcd DB. So when externaldb is passed, we should not run this at all (in addition to the other comments made on one of the rke2 files).

pkg/assert/validate.go
```diff
@@ -97,3 +118,43 @@ func ValidateOnNode(ip string, args ...string) error {
...
+func addResult(command, assertion, result string) {
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	results = append(results, TestResult{Command: "\nCommand:\n" + command + "\n", Assertion: "\nAssertion:\n" +
+		assertion + "\n", Result: "\nMatched with result:\n" + result + "\n"})
+}
+
+// PrintResults prints the results from the tests called
+func PrintResults() {
```
rancher-max, line 131: Can we update the tests themselves so that they print their own results? This feels like an arbitrary call that is prone to miss needed changes when changes are made to those tests.

shared/aux.go
```diff
@@ -480,6 +480,21 @@ func appendNodeIfMissing(slice []Node, i Node) []Node {
 	}
 	return append(slice, i)
 }
+
 func EncloseSqBraces(ip string) string {
 	return "[" + ip + "]"
 }
+
+// PrintGetAll prints the output of kubectl get all -A -o wide and kubectl get nodes -o wide
+func PrintGetAll() {
```
rancher-max, line 489: I think we have this already with `PrintClusterState`

rancher-max (commented):
I think we should leave the CNI version bumps separate. It's not really valid to just change the config.yaml to a different CNI as they affect the host as well and can conflict with one another. Small ref: https://www.suse.com/support/kb/doc/?id=000020904.

modules/install/join_rke2_master.sh
fmoral2: i just added here thinking that we might find good to have etcdctl available where needed , now or future 

but if you think there is no reason, i can add that on the test itself. let me know  @rancher-max

pkg/assert/validate.go
fmoral2: tbh i dont want to add anything like that in any test, i want to have the flexibilty to only call where i want when i want , if i understood correctly your request.

so in this way, we only call this gather and print of results, in where we want.

shared/aux.go
fmoral2:
```
	cmd := "kubectl get nodes,pods -A -o wide --kubeconfig=" + KubeConfigFile
```
PrintClusterState is only print that, not all other  resources

modules/install/join_rke2_master.sh
fmoral2: done

modules/install/k3s_master.sh
fmoral2: done

modules/install/k3s_master.sh
```diff
@@ -93,14 +93,37 @@ disable_cloud_setup() {
...
+    else
+      zypper update -y > /dev/null 2>&1
+      etcd_download
+  fi
+}
+
 install() {
   export "$install_mode"="$version"
 
   if [ "$datastore_type" = "etcd" ]; then
```
mdrahman-suse, line 120: We can add the `install_etcdctl` here before checking channel

modules/install/join_rke2_master.sh
```diff
@@ -97,6 +97,53 @@ cis_setup() {
   fi
 }
 
+#etcd_download() {
```
mdrahman-suse, line 100: Should remove this commented code?

modules/install/join_k3s_master.sh
```diff
@@ -94,6 +94,53 @@ policy_files() {
   fi
   sleep 20
 }
+#
```
mdrahman-suse, line 97: Should remove the commented code?

modules/install/join_rke2_master.sh
fmoral2: yes, tk you !

modules/install/k3s_master.sh
fmoral2: good call, fixed

docs/version_bump_template.md
```diff
@@ -37,50 +71,49 @@ Available arguments to create your command with examples:
...
 
-Example of an execution with multiple values on k3s:
+Example of an execution with multiple values on k3s using versionbump tag:
 ```bash
 go test -timeout=45m -v -tags=versionbump  ./entrypoint/versionbump/... \
 -cmd "/var/lib/rancher/k3s/data/current/bin/cni, kubectl get pod test-pod -o yaml : | grep -A2 annotations, k3s -v" \
 -expectedValue "v1.2.0-k3s1,1M, v1.26" \
 -expectedValueUpgrade "v1.2.0-k3s1,1M, v1.27" \
--installVersionOrCommit INSTALL_K3S_VERSION=v1.27.2+k3s1 \
+-installVersionOrCommit 1.27.2+k3s1 \
```
mdrahman-suse, line 80: should be `v1.27...`?

entrypoint/versionbump/componentsbump_test.go
```diff
@@ -0,0 +1,111 @@
...
+	etcdRke2          = "kubectl get node -o yaml : | grep 'hardened-etcd' -A1, "
+	containerd        = "kubectl get node -o yaml : | grep  containerd, "
+	cniPlugins        = "/var/lib/rancher/k3s/data/current/bin/cni, "
+	traefik           = "kubectl get node -o yaml : | grep traefik  -A1, "
+	localPath         = "kubectl get node -o yaml : | grep local-path -A1, "
+	etcdK3s           = `sudo journalctl -u k3s | grep etcd-version,`
+	klipperLB         = "kubectl get node -o yaml : | grep klipper -A5, "
+)
+
+var _ = Describe("Components Version Upgrade:", func() {
```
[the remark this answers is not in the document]
fmoral2: added dns, etcd i removed everything as it was getting too much out of scope due to changes required

fmoral2:
https://github.com/rancher/distros-test-framework/pull/74#discussion_r1556334143
@rancher-max  
i was reading your comment , looking at the function but thinking in another one. lol sorry 
yes totally, fixed.

entrypoint/versionbump/componentsbump_test.go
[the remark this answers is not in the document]
fmoral2: ADDING top nodes ,pods as well

rancher-max (dismissed):
LGTM

docs/version_bump_template.md
fmoral2: i had fixed somewhere else, doing now 
tks

rancher-max (approved):

mdrahman-suse (approved):
LGTM

PR #77, opened 2024-03-06
secrets encryption testcase

aganesh-suse:
#### Proposed Changes ####

Adding test for secrets encryption workflows:

    traditional workflows: prepare, rotate, reencrypt
    New workflow: rotate-keys
    Note: rotate-keys feature is not ready yet for rke2.

Added documentation help as well.

#### Types of Changes ####

Patch Validation test
#### Testing ####

Checklist:

No

Yes

Yes

Yes

Yes

N/A

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -29,6 +29,35 @@ agent1 ->  agent/worker node
 
 Note/TODO: k3s external db fails working with etcd only node. Refer: https://docs.k3s.io/datastore/ha
 
+### Secret-Encryption tests Setup Requirements/Assumptions
+We need a split role setup for this test:
+1 Etcd ONLY node
+2 Control Plane ONLY node
+1 Agent node
+
+To set this up, please use the following in the tfvars file: 
+
+```
+no_of_server_nodes = 0  # This is for all roles server - etcd + control plane
+no_of_worker_nodes = 1  # Agent node
+split_roles        = true
+etcd_only_nodes    = 1  # etcd only node count
+etcd_cp_nodes      = 0 
+etcd_worker_nodes  = 0
+cp_only_nodes      = 2  # control plane only node count
+cp_worker_nodes    = 0
+# Numbers 1-6 correspond to: all-roles (1), etcd-only (2), etcd-cp (3), etcd-worker (4), cp-only (5), cp-worker (6).
+role_order         = "2,5,5"
+```
+
+Please set the server_flags in .tfvars file:
+```
+server_flags   = "secrets-encryption: true\n"
+```
+
+Note/TODO: k3s external db fails working with etcd only node. Refer: https://docs.k3s.io/datastore/ha
+
+
 ### Dual-Stack Testing
 
 - Required vars for `*.tfvars` file
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -30,7 +30,7 @@ agent1 ->  agent/worker node
 Note/TODO: k3s external db fails working with etcd only node. Refer: https://docs.k3s.io/datastore/ha
 
 ### Secret-Encryption tests Setup Requirements/Assumptions
-We need a split role setup for this test:
+For patch validation test runs, we need a split role setup for this test:
 1 Etcd ONLY node
 2 Control Plane ONLY node
 1 Agent node
@@ -47,13 +47,20 @@ etcd_worker_nodes  = 0
 cp_only_nodes      = 2  # control plane only node count
 cp_worker_nodes    = 0
 # Numbers 1-6 correspond to: all-roles (1), etcd-only (2), etcd-cp (3), etcd-worker (4), cp-only (5), cp-worker (6).
-role_order         = "2,5,5"
+role_order         = "2,5"
 ```
+Please note, we can also run this test on a regular HA setup - 3 all-roles server, 1 worker node. (without split roles)
 
-Please set the server_flags in .tfvars file:
+Please set the server_flags in .tfvars file for k3s:
 ```
 server_flags   = "secrets-encryption: true\n"
 ```
+In case of rke2 - do not leave this empty for now.
+At least set a "token: secret\n" for both server and agent flags as a workaround for:
+https://github.com/rancher/distros-test-framework/issues/86
+
+For versions 1.26 and 1.27 - we run the traditional tests only: prepare/rotate/renenrypt (TEST_TYPE gets set to 'classic' in env var. We use this to determine which tests to run.)
+For versions 1.28 and greater - we run both the traditional tests and new method - rotate-keys (TEST_TYPE gets set to 'both' in env var)
 
 Note/TODO: k3s external db fails working with etcd only node. Refer: https://docs.k3s.io/datastore/ha
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -55,9 +55,6 @@ Please set the server_flags in .tfvars file for k3s:
 ```
 server_flags   = "secrets-encryption: true\n"
 ```
-In case of rke2 - do not leave this empty for now.
-At least set a "token: secret\n" for both server and agent flags as a workaround for:
-https://github.com/rancher/distros-test-framework/issues/86
 
 For versions 1.26 and 1.27 - we run the traditional tests only: prepare/rotate/reencrypt (TEST_TYPE gets set to 'classic' in env var. We use this to determine which tests to run.)
 For versions 1.28 and greater - we run both the traditional tests and new method - rotate-keys (TEST_TYPE gets set to 'both' in env var)
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -59,7 +59,7 @@ In case of rke2 - do not leave this empty for now.
 At least set a "token: secret\n" for both server and agent flags as a workaround for:
 https://github.com/rancher/distros-test-framework/issues/86
 
-For versions 1.26 and 1.27 - we run the traditional tests only: prepare/rotate/renenrypt (TEST_TYPE gets set to 'classic' in env var. We use this to determine which tests to run.)
+For versions 1.26 and 1.27 - we run the traditional tests only: prepare/rotate/reencrypt (TEST_TYPE gets set to 'classic' in env var. We use this to determine which tests to run.)
 For versions 1.28 and greater - we run both the traditional tests and new method - rotate-keys (TEST_TYPE gets set to 'both' in env var)
 
 Note/TODO: k3s external db fails working with etcd only node. Refer: https://docs.k3s.io/datastore/ha
--- a/entrypoint/secretsencrypt/encryption_suite_test.go
+++ b/entrypoint/secretsencrypt/encryption_suite_test.go
@@ -2,6 +2,7 @@ package secretsencrypt
 
 import (
 	"flag"
+	"fmt"
 	"os"
 	"testing"
 
@@ -33,6 +34,14 @@ func TestSecretsEncryptionSuite(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Secrets Encryption Test Suite")
 }
+
+var _ = BeforeSuite(func() {
+	if err := config.SetEnv(shared.BasePath() + fmt.Sprintf("/config/%s.tfvars", cfg.Product)); err != nil {
+		Expect(err).To(BeNil(), fmt.Sprintf("error loading tf vars: %v\n", err))
+	}
+	Expect(os.Getenv("server_flags")).To(ContainSubstring("secrets-encryption:"),
+		"FATAL: Add secrets-encryption:true to server_flags for this test")
+})
 
 var _ = AfterSuite(func() {
 	g := GinkgoT()
--- a/entrypoint/secretsencrypt/encryption_suite_test.go
+++ b/entrypoint/secretsencrypt/encryption_suite_test.go
@@ -1,0 +1,44 @@
+package secretsencrypt
+
+import (
+	"flag"
+	"os"
+	"testing"
+
+	"github.com/rancher/distros-test-framework/config"
+	"github.com/rancher/distros-test-framework/factory"
+	"github.com/rancher/distros-test-framework/pkg/customflag"
+	"github.com/rancher/distros-test-framework/shared"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var cfg *config.Product
+
+func TestMain(m *testing.M) {
+	var err error
+	flag.Var(&customflag.ServiceFlag.ClusterConfig.Destroy, "destroy", "Destroy cluster after test")
+	flag.Parse()
+
+	cfg, err = shared.EnvConfig()
+	if err != nil {
+		return
+	}
+
+	os.Exit(m.Run())
+}
+
+func TestSecretsEncryptionSuite(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Secrets Encryption Test Suite")
+}
+
+var _ = AfterSuite(func() {
+	g := GinkgoT()
+	if customflag.ServiceFlag.ClusterConfig.Destroy {
+		status, err := factory.DestroyCluster(g)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(status).To(Equal("cluster destroyed"))
+	}
+})
--- a/entrypoint/secretsencrypt/encryption_suite_test.go
+++ b/entrypoint/secretsencrypt/encryption_suite_test.go
@@ -40,9 +40,10 @@ var _ = BeforeSuite(func() {
 	if err := config.SetEnv(shared.BasePath() + fmt.Sprintf("/config/%s.tfvars", cfg.Product)); err != nil {
 		Expect(err).To(BeNil(), fmt.Sprintf("error loading tf vars: %v\n", err))
 	}
-	Expect(os.Getenv("server_flags")).To(ContainSubstring("secrets-encryption:"),
-		"FATAL: Add secrets-encryption:true to server_flags for this test")
-
+	if cfg.Product == "k3s" {
+		Expect(os.Getenv("server_flags")).To(ContainSubstring("secrets-encryption:"),
+			"ERROR: Add secrets-encryption:true to server_flags for this test")
+	}
 	version := os.Getenv(fmt.Sprintf("%s_version", cfg.Product))
 	if strings.Contains(version, "1.27") || strings.Contains(version, "1.26") {
 		os.Setenv("TEST_TYPE", "classic")
--- a/entrypoint/secretsencrypt/encryption_suite_test.go
+++ b/entrypoint/secretsencrypt/encryption_suite_test.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/rancher/distros-test-framework/config"
@@ -41,6 +42,14 @@ var _ = BeforeSuite(func() {
 	}
 	Expect(os.Getenv("server_flags")).To(ContainSubstring("secrets-encryption:"),
 		"FATAL: Add secrets-encryption:true to server_flags for this test")
+
+	version := os.Getenv(fmt.Sprintf("%s_version", cfg.Product))
+	if strings.Contains(version, "1.27") || strings.Contains(version, "1.26") {
+		os.Setenv("TEST_TYPE", "classic")
+	} else {
+		os.Setenv("TEST_TYPE", "both")
+	}
+
 })
 
 var _ = AfterSuite(func() {
--- a/entrypoint/secretsencrypt/encryption_test.go
+++ b/entrypoint/secretsencrypt/encryption_test.go
@@ -52,8 +52,8 @@ var _ = Describe("Test:", func() {
 
 var _ = AfterEach(func() {
 	if CurrentSpecReport().Failed() {
-		fmt.Printf("\nFAILED! %s\n", CurrentSpecReport().FullText())
+		fmt.Printf("\nFAILED! " + CurrentSpecReport().FullText() + "\n")
 	} else {
-		fmt.Printf("\nPASSED! %s\n", CurrentSpecReport().FullText())
+		fmt.Printf("\nPASSED! " + CurrentSpecReport().FullText() + "\n")
 	}
 })
--- a/entrypoint/secretsencrypt/encryption_test.go
+++ b/entrypoint/secretsencrypt/encryption_test.go
@@ -1,0 +1,59 @@
+package secretsencrypt
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+
+	"github.com/rancher/distros-test-framework/pkg/assert"
+	"github.com/rancher/distros-test-framework/pkg/testcase"
+)
+
+var _ = Describe("Test:", func() {
+
+	It("Start Up with no issues", func() {
+		testcase.TestBuildCluster(GinkgoT())
+	})
+
+	It("Validate Nodes", func() {
+		testcase.TestNodeStatus(
+			assert.NodeAssertReadyStatus(),
+			nil,
+		)
+	})
+
+	It("Validate Pods", func() {
+		testcase.TestPodStatus(
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus(),
+		)
+	})
+
+	It("Validate Secrets Encryption", func() {
+		testcase.TestSecretsEncrypt()
+	})
+
+	It("Validate Nodes", func() {
+		testcase.TestNodeStatus(
+			assert.NodeAssertReadyStatus(),
+			nil,
+		)
+	})
+
+	It("Validate Pods", func() {
+		testcase.TestPodStatus(
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus(),
+		)
+	})
+})
+
+var _ = AfterEach(func() {
+	if CurrentSpecReport().Failed() {
+		fmt.Printf("\nFAILED! %s\n", CurrentSpecReport().FullText())
+	} else {
+		fmt.Printf("\nPASSED! %s\n", CurrentSpecReport().FullText())
+	}
+})
--- a/entrypoint/secretsencrypt/encryption_test.go
+++ b/entrypoint/secretsencrypt/encryption_test.go
@@ -42,11 +42,11 @@ var _ = Describe("Test:", func() {
 	})
 
 	It("Validate Pods", func() {
-		testcase.TestPodStatus(nil, nil, nil)
-		// assert.PodAssertRestart(),
-		// assert.PodAssertReady(),
-		// assert.PodAssertStatus(),
-		// )
+		testcase.TestPodStatus(
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus(),
+		)
 	})
 })
 
--- a/entrypoint/secretsencrypt/encryption_test.go
+++ b/entrypoint/secretsencrypt/encryption_test.go
@@ -31,7 +31,7 @@ var _ = Describe("Test:", func() {
 	})
 
 	It("Validate Secrets Encryption", func() {
-		testcase.TestSecretsEncrypt()
+		testcase.TestSecretsEncryption()
 	})
 
 	It("Validate Nodes", func() {
--- a/entrypoint/secretsencrypt/encryption_test.go
+++ b/entrypoint/secretsencrypt/encryption_test.go
@@ -42,11 +42,11 @@ var _ = Describe("Test:", func() {
 	})
 
 	It("Validate Pods", func() {
-		testcase.TestPodStatus(
-			assert.PodAssertRestart(),
-			assert.PodAssertReady(),
-			assert.PodAssertStatus(),
-		)
+		testcase.TestPodStatus(nil, nil, nil)
+		// assert.PodAssertRestart(),
+		// assert.PodAssertReady(),
+		// assert.PodAssertStatus(),
+		// )
 	})
 })
 
--- a/pkg/testcase/secretsencrypt.go
+++ b/pkg/testcase/secretsencrypt.go
@@ -23,7 +23,9 @@ func TestSecretsEncryption() {
 
 	errSecret := shared.CreateSecret("secret1", "default")
 	Expect(errSecret).NotTo(HaveOccurred(), "error creating secret")
+
 	shared.LogLevel("INFO", "TEST: 'CLASSIC' Secrets Encryption method")
+
 	index := len(nodes) - 1
 	secretsEncryptOps("prepare", product, nodes[index].ExternalIP, ips)
 	secretsEncryptOps("rotate", product, nodes[index].ExternalIP, ips)
@@ -56,7 +58,7 @@ func secretsEncryptOps(action, product, cpIp string, ips []string) {
 		waitEtcdErr := shared.WaitForPodsRunning(5, 4, false)
 		if waitEtcdErr != nil {
 			shared.LogLevel("WARN", "pods not up after 20 seconds.")
-			if i != len(ips) {
+			if i != len(ips)-1 {
 				shared.LogLevel("DEBUG", "continue service restarts")
 			}
 		}
@@ -92,11 +94,13 @@ func waitForHashMatch(cpIp, product string, defaultTime time.Duration, times int
 		}
 		if secretEncryptStatus != "" && strings.Contains(secretEncryptStatus, "All hashes match") {
 			shared.LogLevel("DEBUG", "Total sleep time before hashes matched: %d seconds", i*int(defaultTime))
+
 			return secretEncryptStatus, nil
 		}
 		time.Sleep(defaultTime * time.Second)
 	}
 	shared.LogLevel("WARN", "Hashes did not match after %d seconds", times*int(defaultTime))
+
 	return secretEncryptStatus, errGetStatus
 }
 
@@ -142,6 +146,7 @@ func getNodeIps(nodes []shared.Node) []string {
 			"Node details: name: %s status: %s roles: %s external ip: %s",
 			node.Name, node.Status, node.Roles, node.ExternalIP)
 	}
+
 	return nodeIps
 }
 
@@ -166,5 +171,6 @@ func logEncryptionFileContents(ips []string, product string) error {
 			return shared.ReturnLogError(fmt.Sprintf("Error cat of %s", stateFile))
 		}
 	}
+
 	return nil
 }
--- a/pkg/testcase/secretsencrypt.go
+++ b/pkg/testcase/secretsencrypt.go
@@ -132,7 +132,8 @@ func logEncryptionFileContents(ips []string, product string) error {
 			return shared.ReturnLogError(fmt.Sprintf("Error cat of %s", configFile))
 		}
 		currentTime := time.Now()
-		Expect(configStdOut).To(ContainSubstring(fmt.Sprintf("aescbckey-%s", currentTime.Format("2006-01-02"))))
+		Expect(configStdOut).To(ContainSubstring(fmt.Sprintf("aescbckey-%s",
+			currentTime.Format("2006-01-02"))))
 		_, errState := shared.RunCommandOnNode(cmdShowState, ip)
 		if errState != nil {
 			return shared.ReturnLogError(fmt.Sprintf("Error cat of %s", stateFile))
--- a/pkg/testcase/secretsencrypt.go
+++ b/pkg/testcase/secretsencrypt.go
@@ -1,0 +1,137 @@
+package testcase
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/rancher/distros-test-framework/shared"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestSecretsEncrypt() {
+	etcdNodes, errGetEtcd := shared.GetNodesByRoles("etcd")
+	Expect(etcdNodes).NotTo(BeEmpty())
+	Expect(errGetEtcd).NotTo(HaveOccurred(), "error getting etcd nodes")
+
+	cpNodes, errGetCP := shared.GetNodesByRoles("control-plane")
+	Expect(cpNodes).NotTo(BeEmpty())
+	Expect(errGetCP).NotTo(HaveOccurred(), "error getting control plane nodes")
+
+	product, err := shared.Product()
+	Expect(err).NotTo(HaveOccurred(), "error getting product from config")
+
+	ips := getNodeIps(etcdNodes, cpNodes)
+
+	errSecret := shared.CreateSecret("secret1", "default")
+	Expect(errSecret).NotTo(HaveOccurred(), "error creating secret")
+
+	secretsEncryptOps("prepare", product, cpNodes[0].ExternalIP, ips)
+	secretsEncryptOps("rotate", product, cpNodes[0].ExternalIP, ips)
+	secretsEncryptOps("reencrypt", product, cpNodes[0].ExternalIP, ips)
+	secretsEncryptOps("rotate-keys", product, cpNodes[0].ExternalIP, ips)
+}
+
+func secretsEncryptOps(action, product, cpIp string, ips []string) {
+	shared.LogLevel("INFO", fmt.Sprintf("TEST: Secrets-Encryption: %s", action))
+	_, errStatusB4 := shared.SecretEncryptOps("status", cpIp, product)
+	Expect(errStatusB4).NotTo(HaveOccurred(), "error getting secret-encryption status before action")
+
+	stdOutput, err := shared.SecretEncryptOps(action, cpIp, product)
+	Expect(err).NotTo(HaveOccurred(), "error: secret-encryption: "+action)
+	verifyStdOut(action, stdOutput)
+	if (action == "reencrypt") || (action == "rotate-keys") {
+		time.Sleep(20 * time.Second)
+	}
+
+	for _, node := range ips {
+		var nodearr []string
+		nodearr = append(nodearr, node)
+		nodeIp, errRestart := shared.ManageService(product, "restart", "server", nodearr)
+		Expect(errRestart).NotTo(HaveOccurred(), "error restart service for node: "+nodeIp)
+		// Order of reboot matters. Etcd first then control plane nodes.
+		// Little lag needed between node restarts to avoid issues.
+		shared.LogLevel("INFO", "Sleep 10 seconds - wait before restarting next node in cluster")
+		time.Sleep(10 * time.Second)
+	}
+	switch product {
+	case "k3s":
+		shared.LogLevel("INFO", "Sleep 30 seconds - wait for services to come up")
+		time.Sleep(30 * time.Second)
+	case "rke2":
+		shared.LogLevel("INFO", "Sleep 60 seconds - wait for services to come up")
+		time.Sleep(60 * time.Second)
+	}
+
+	stdStatusOut, errStatus := shared.SecretEncryptOps("status", cpIp, product)
+	Expect(errStatus).NotTo(HaveOccurred(), "error getting secret-encryption status")
+	verifyStatusOutput(action, stdStatusOut)
+
+	errLog := logEncryptionFileContents(ips, product)
+	Expect(errLog).NotTo(HaveOccurred())
+}
+
+func verifyStdOut(action, stdout string) {
+	switch action {
+	case "prepare":
+		Expect(stdout).To(ContainSubstring("prepare completed successfully"))
+	case "rotate":
+		Expect(stdout).To(ContainSubstring("rotate completed successfully"))
+	case "reencrypt":
+		Expect(stdout).To(ContainSubstring("reencryption started"))
+	case "rotate-keys":
+		Expect(stdout).To(ContainSubstring("keys rotated, reencryption started"))
+	}
+}
+
+func verifyStatusOutput(action, stdout string) {
+	switch action {
+	case "prepare":
+		Expect(stdout).To(ContainSubstring("Encryption Status: Enabled"))
+		Expect(stdout).To(ContainSubstring("Current Rotation Stage: prepare"))
+		Expect(stdout).To(ContainSubstring("Server Encryption Hashes: All hashes match"))
+	case "rotate":
+		Expect(stdout).To(ContainSubstring("Encryption Status: Enabled"))
+		Expect(stdout).To(ContainSubstring("Current Rotation Stage: rotate"))
+		Expect(stdout).To(ContainSubstring("Server Encryption Hashes: All hashes match"))
+	case "reencrypt":
+		Expect(stdout).To(ContainSubstring("Encryption Status: Enabled"))
+		Expect(stdout).To(ContainSubstring("Current Rotation Stage: reencrypt_finished"))
+		Expect(stdout).To(ContainSubstring("Server Encryption Hashes: All hashes match"))
+	case "rotate-keys":
+		Expect(stdout).To(ContainSubstring("Encryption Status: Enabled"))
+		Expect(stdout).To(ContainSubstring("Current Rotation Stage: reencrypt_finished"))
+		Expect(stdout).To(ContainSubstring("Server Encryption Hashes: All hashes match"))
+	}
+}
+
+func getNodeIps(etcdNodes, cpNodes []shared.Node) []string {
+	var sarray []string
+	for _, node := range etcdNodes {
+		sarray = append(sarray, node.ExternalIP)
+	}
+	for _, node := range cpNodes {
+		sarray = append(sarray, node.ExternalIP)
+	}
+	return sarray
+}
+
+func logEncryptionFileContents(ips []string, product string) error {
+	configFile := fmt.Sprintf("/var/lib/rancher/%s/server/cred/encryption-config.json", product)
+	stateFile := fmt.Sprintf("/var/lib/rancher/%s/server/cred/encryption-state.json", product)
+	cmdShowConfig := fmt.Sprintf("sudo cat %s", configFile)
+	cmdShowState := fmt.Sprintf("sudo cat %s", stateFile)
+	for _, ip := range ips {
+		sout, errConfig := shared.RunCommandOnNode(cmdShowConfig, ip)
+		if errConfig != nil {
+			return shared.ReturnLogError(fmt.Sprintf("Error cat of %s", configFile))
+		}
+		currentTime := time.Now()
+		Expect(sout).To(ContainSubstring(fmt.Sprintf("aescbckey-%s", currentTime.Format("2006-01-02"))))
+		_, errState := shared.RunCommandOnNode(cmdShowState, ip)
+		if errState != nil {
+			return shared.ReturnLogError(fmt.Sprintf("Error cat of %s", stateFile))
+		}
+	}
+	return nil
+}
--- a/pkg/testcase/secretsencrypt.go
+++ b/pkg/testcase/secretsencrypt.go
@@ -21,7 +21,7 @@ func TestSecretsEncryption() {
 
 	ips := getNodeIps(nodes)
 
-	errSecret := shared.CreateSecret("secret1", "default")
+	errSecret := shared.CreateSecret("secret1", "default", false)
 	Expect(errSecret).NotTo(HaveOccurred(), "error creating secret")
 
 	shared.LogLevel("INFO", "TEST: 'CLASSIC' Secrets Encryption method")
--- a/pkg/testcase/secretsencrypt.go
+++ b/pkg/testcase/secretsencrypt.go
@@ -2,6 +2,8 @@ package testcase
 
 import (
 	"fmt"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/rancher/distros-test-framework/shared"
@@ -25,11 +27,14 @@ func TestSecretsEncrypt() {
 
 	errSecret := shared.CreateSecret("secret1", "default")
 	Expect(errSecret).NotTo(HaveOccurred(), "error creating secret")
-
+	shared.LogLevel("INFO", "TEST: 'CLASSIC' Secrets Encryption method")
 	secretsEncryptOps("prepare", product, cpNodes[0].ExternalIP, ips)
 	secretsEncryptOps("rotate", product, cpNodes[0].ExternalIP, ips)
 	secretsEncryptOps("reencrypt", product, cpNodes[0].ExternalIP, ips)
-	secretsEncryptOps("rotate-keys", product, cpNodes[0].ExternalIP, ips)
+	if strings.Contains(os.Getenv("TEST_TYPE"), "both") {
+		shared.LogLevel("INFO", "TEST: 'NEW' Secrets Encryption method")
+		secretsEncryptOps("rotate-keys", product, cpNodes[0].ExternalIP, ips)
+	}
 }
 
 func secretsEncryptOps(action, product, cpIp string, ips []string) {
--- a/pkg/testcase/secretsencrypt.go
+++ b/pkg/testcase/secretsencrypt.go
@@ -54,7 +54,7 @@ func secretsEncryptOps(action, product, cpIp string, nodes []shared.Node) {
 		Expect(errRestart).NotTo(HaveOccurred(), "error restart service for node: "+nodeIp)
 		// Order of reboot matters. Etcd first then control plane nodes.
 		// Little lag needed between node restarts to avoid issues.
-		waitEtcdErr := shared.WaitForPodsRunning(5, 4, false)
+		waitEtcdErr := shared.WaitForPodsRunning(5, 4)
 		if waitEtcdErr != nil {
 			shared.LogLevel("WARN", "pods not up after 20 seconds.")
 			if i != len(nodes)-1 {
@@ -64,12 +64,12 @@ func secretsEncryptOps(action, product, cpIp string, nodes []shared.Node) {
 	}
 	switch product {
 	case "k3s":
-		waitPodsErr := shared.WaitForPodsRunning(5, 6, false)
+		waitPodsErr := shared.WaitForPodsRunning(5, 6)
 		if waitPodsErr != nil {
 			shared.LogLevel("WARN", "pods not up after 30 seconds")
 		}
 	case "rke2":
-		waitPodsErr := shared.WaitForPodsRunning(5, 12, false)
+		waitPodsErr := shared.WaitForPodsRunning(5, 12)
 		if waitPodsErr != nil {
 			shared.LogLevel("WARN", "pods not up after 60 seconds")
 		}
@@ -104,7 +104,7 @@ func waitForHashMatch(cpIp, product string, defaultTime time.Duration, times int
 }
 
 // verifyActionStdOut Verifies secrets-encryption action outputs
-// Verfies std outputs of: sudo k3s|rke2 secrets-encryption prepare|rotate|reencrypt|rotate-keys actions
+// Verifies std outputs of: sudo k3s|rke2 secrets-encryption prepare|rotate|reencrypt|rotate-keys actions
 func verifyActionStdOut(action, stdout string) {
 	switch action {
 	case "prepare":
@@ -118,7 +118,7 @@ func verifyActionStdOut(action, stdout string) {
 	}
 }
 
-// verifyStatusStdOut Verfies secrets-encryption status outputs post different actions.
+// verifyStatusStdOut Verifies secrets-encryption status outputs post different actions.
 // Verifies std output of: sudo k3s|rke2 secrets-encryption status
 // post the action -prepare|rotate|reencrypt|rotate-keys and restart services have been completed
 func verifyStatusStdOut(action, stdout string) {
--- a/pkg/testcase/secretsencrypt.go
+++ b/pkg/testcase/secretsencrypt.go
@@ -86,16 +86,15 @@ func waitForHashMatch(cpIp, product string, defaultTime time.Duration, times int
 	var secretEncryptStatus string
 	var errGetStatus error
 	for i := 1; i <= times; i++ {
-		secretEncryptStatus, errGetStatus := shared.SecretEncryptOps("status", cpIp, product)
+		secretEncryptStatus, errGetStatus = shared.SecretEncryptOps("status", cpIp, product)
 		if errGetStatus != nil {
 			shared.LogLevel("DEBUG", "error getting secret-encryption status. Retry.")
 		}
 		if secretEncryptStatus != "" && strings.Contains(secretEncryptStatus, "All hashes match") {
 			shared.LogLevel("DEBUG", "Total sleep time before hashes matched: %d seconds", i*int(defaultTime))
 			return secretEncryptStatus, nil
-		} else {
-			time.Sleep(defaultTime * time.Second)
 		}
+		time.Sleep(defaultTime * time.Second)
 	}
 	shared.LogLevel("WARN", "Hashes did not match after %d seconds", times*int(defaultTime))
 	return secretEncryptStatus, errGetStatus
@@ -139,7 +138,9 @@ func getNodeIps(nodes []shared.Node) []string {
 	var nodeIps []string
 	for _, node := range nodes {
 		nodeIps = append(nodeIps, node.ExternalIP)
-		shared.LogLevel("DEBUG", "Node details: name: %s status: %s roles: %s external ip: %s", node.Name, node.Status, node.Roles, node.ExternalIP)
+		shared.LogLevel("DEBUG",
+			"Node details: name: %s status: %s roles: %s external ip: %s",
+			node.Name, node.Status, node.Roles, node.ExternalIP)
 	}
 	return nodeIps
 }
--- a/pkg/testcase/secretsencrypt.go
+++ b/pkg/testcase/secretsencrypt.go
@@ -125,9 +125,7 @@ func verifyStatusOutput(action, stdout string) {
 		Expect(stdout).To(ContainSubstring("Current Rotation Stage: prepare"))
 	case "rotate":
 		Expect(stdout).To(ContainSubstring("Current Rotation Stage: rotate"))
-	case "reencrypt":
-		Expect(stdout).To(ContainSubstring("Current Rotation Stage: reencrypt_finished"))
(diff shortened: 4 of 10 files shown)


shared/product.go
```diff
@@ -111,3 +112,26 @@ func CertRotate(product string, ips []string) (string, error) {
 
 	return "", nil
 }
+
```
fmoral2, line 115: honestly i think `SystemCtlCmd` , `ManageService` ,  `CertRotate`, and now `SecretEncryptOps` does not makes much sense here , since here should be really related to product functions itself

if we go think like, " oh but they use a product or something like that " then almost half of the code should live here lol

so i would really suggest that those go to cluster file

shared/product.go
```diff
@@ -111,3 +112,26 @@ func CertRotate(product string, ips []string) (string, error) {
...
+		"enable":      fmt.Sprintf("sudo %s secrets-encrypt enable", product),
+		"disable":     fmt.Sprintf("sudo %s secrets-encrypt disable", product),
+		"prepare":     fmt.Sprintf("sudo %s secrets-encrypt prepare", product),
+		"rotate":      fmt.Sprintf("sudo %s secrets-encrypt rotate", product),
+		"reencrypt":   fmt.Sprintf("sudo %s secrets-encrypt reencrypt", product),
+		"rotate-keys": fmt.Sprintf("sudo %s secrets-encrypt rotate-keys", product),
+	}
+
+	stdout, err := RunCommandOnNode(secretEncryptCmd[action], ip)
+
```
fmoral2, line 128: empty line with no need

shared/product.go
```diff
@@ -111,3 +112,26 @@ func CertRotate(product string, ips []string) (string, error) {
...
+	if err != nil {
+		return ip, ReturnLogError(fmt.Sprintf("FATAL: secrets-encryption %s action failed", action), err)
```
fmoral2, line 130: why returning IP on error ? are you using it even on error flow ?

shared/product.go
```diff
@@ -111,3 +112,26 @@ func CertRotate(product string, ips []string) (string, error) {
...
+		"status":      fmt.Sprintf("sudo %s secrets-encrypt status", product),
+		"enable":      fmt.Sprintf("sudo %s secrets-encrypt enable", product),
+		"disable":     fmt.Sprintf("sudo %s secrets-encrypt disable", product),
+		"prepare":     fmt.Sprintf("sudo %s secrets-encrypt prepare", product),
+		"rotate":      fmt.Sprintf("sudo %s secrets-encrypt rotate", product),
+		"reencrypt":   fmt.Sprintf("sudo %s secrets-encrypt reencrypt", product),
+		"rotate-keys": fmt.Sprintf("sudo %s secrets-encrypt rotate-keys", product),
+	}
+
+	stdout, err := RunCommandOnNode(secretEncryptCmd[action], ip)
```
fmoral2, line 127: is there any other more meaningful name to this var instead of stdout ? we could be more specific here

shared/cluster.go
```diff
@@ -584,3 +584,20 @@ func processNodeArgs(nodeArgs string) (nodeArgsMapSlice []map[string]string) {
...
 }
+
+func CreateSecret(secret, namespace string) error {
+	kubectl := fmt.Sprintf("kubectl --kubeconfig %s", KubeConfigFile)
+	if namespace == "" {
+		namespace = "default"
+	}
+	cmd := fmt.Sprintf("%s create secret generic %s -n %s --from-literal=mykey=mydata",
+		kubectl, secret, namespace)
+	stdout, err := RunCommandHost(cmd)
```
fmoral2, line 595: same comment as  
"

is there any other more meaningful name to this var instead of stdout ? we could be more specific here
"

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,137 @@
...
+	errSecret := shared.CreateSecret("secret1", "default")
+	Expect(errSecret).NotTo(HaveOccurred(), "error creating secret")
+
+	secretsEncryptOps("prepare", product, cpNodes[0].ExternalIP, ips)
+	secretsEncryptOps("rotate", product, cpNodes[0].ExternalIP, ips)
+	secretsEncryptOps("reencrypt", product, cpNodes[0].ExternalIP, ips)
+	secretsEncryptOps("rotate-keys", product, cpNodes[0].ExternalIP, ips)
+}
+
+func secretsEncryptOps(action, product, cpIp string, ips []string) {
```
fmoral2, line 35: can we return error please ? since has some possible areas that could go wrong, would be good to send the error above

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,137 @@
...
+
+	stdStatusOut, errStatus := shared.SecretEncryptOps("status", cpIp, product)
+	Expect(errStatus).NotTo(HaveOccurred(), "error getting secret-encryption status")
+	verifyStatusOutput(action, stdStatusOut)
+
+	errLog := logEncryptionFileContents(ips, product)
+	Expect(errLog).NotTo(HaveOccurred())
+}
+
+func verifyStdOut(action, stdout string) {
```
fmoral2, line 74: same comment- 


can we return error please ? since has some possible areas that could go wrong, would be good to send the error above

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,137 @@
...
+	case "rotate":
+		Expect(stdout).To(ContainSubstring("rotate completed successfully"))
+	case "reencrypt":
+		Expect(stdout).To(ContainSubstring("reencryption started"))
+	case "rotate-keys":
+		Expect(stdout).To(ContainSubstring("keys rotated, reencryption started"))
+	}
+}
+
+func verifyStatusOutput(action, stdout string) {
```
fmoral2, line 87: same comment


can we return error please ? since has some possible areas that could go wrong, would be good to send the error above

you could use default as error

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,137 @@
...
+		Expect(stdout).To(ContainSubstring("Server Encryption Hashes: All hashes match"))
+	case "rotate-keys":
+		Expect(stdout).To(ContainSubstring("Encryption Status: Enabled"))
+		Expect(stdout).To(ContainSubstring("Current Rotation Stage: reencrypt_finished"))
+		Expect(stdout).To(ContainSubstring("Server Encryption Hashes: All hashes match"))
+	}
+}
+
+func getNodeIps(etcdNodes, cpNodes []shared.Node) []string {
+	var sarray []string
```
fmoral2, line 109: sorry lol, maybe missing something 

what is sarray ?  shared array ?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return sarray
+}
+
+func logEncryptionFileContents(ips []string, product string) error {
+	configFile := fmt.Sprintf("/var/lib/rancher/%s/server/cred/encryption-config.json", product)
+	stateFile := fmt.Sprintf("/var/lib/rancher/%s/server/cred/encryption-state.json", product)
+	cmdShowConfig := fmt.Sprintf("sudo cat %s", configFile)
+	cmdShowState := fmt.Sprintf("sudo cat %s", stateFile)
+	for _, ip := range ips {
```
fmoral2, line 124: empty line before the for loop

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,137 @@
...
+		sout, errConfig := shared.RunCommandOnNode(cmdShowConfig, ip)
```
fmoral2, line 125: what would be sout ?  can we name it better ?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,137 @@
...
+	Expect(err).NotTo(HaveOccurred(), "error: secret-encryption: "+action)
+	verifyStdOut(action, stdOutput)
+	if (action == "reencrypt") || (action == "rotate-keys") {
+		time.Sleep(20 * time.Second)
+	}
+
+	for _, node := range ips {
+		var nodearr []string
+		nodearr = append(nodearr, node)
+		nodeIp, errRestart := shared.ManageService(product, "restart", "server", nodearr)
```
fmoral2, line 50: if you are sending " nodearr" each loop iteration , why creating a slice and appending ?

pkg/testcase/secretsencrypt.go
fmoral2: @aganesh-suse  i resolve the other ones, this one we still need to return err, since we are only chekcing strings

pkg/testcase/secretsencrypt.go
fmoral2: sure, but  "node" its a new var every loop right ?

why need the slice and append?

pkg/testcase/secretsencrypt.go
aganesh-suse: we dont have any errors here. we are passing the stdout we already got, and the action. we are only checking the content of the stdout. there is no error generation other than the Expect errors. unless thats what you want me to pass? like catch the expect error and verify?

pkg/testcase/secretsencrypt.go
aganesh-suse: i was thinkin string array in my mind. i could probably rename if to nodeIpsArr or some such. :)

pkg/testcase/secretsencrypt.go
fmoral2: https://go.dev/play/p/TJTUXucDSq8

pkg/testcase/secretsencrypt.go
aganesh-suse: will rename it to configStdOut

pkg/testcase/secretsencrypt.go
aganesh-suse: the manageservice expects []string, and node is a string.

pkg/testcase/secretsencrypt.go
fmoral2:
```
stdStatusOut, errStatus := shared.SecretEncryptOps("status", cpIp, product)
	Expect(errStatus).NotTo(HaveOccurred(), "error getting secret-encryption status")
	verifyStatusOutput(action, stdStatusOut)
```


so you are sending a stdout to verifyStatusOutput , which you dont know what might be the value, you are just validating that has no error....but the value of the string you are not.

so on the verifyStatusOutput you just check if strings contains right? 

and if the value does not contains ?

pkg/testcase/secretsencrypt.go
aganesh-suse: the manageservice was built to reboot all nodes at once. not to handle one by one.

pkg/testcase/secretsencrypt.go
fmoral2: never mind, forget me , i misunsdertood, you are sending the status already

thats good!  sorry

pkg/testcase/secretsencrypt.go
aganesh-suse: func shared.ManageService(product string, action string, nodeType string, ips []string) (string, error)

fmoral2: got you

but still not need to append right? 


```
nodearr := []string{node}
 nodeIp, errRestart := shared.ManageService(product, "restart", "server", nodearr)
```

shared/product.go
aganesh-suse: so the commands that go :
sudo k3s --help 
sudo rke2 --help

they all belong here. They are testing k3s/rke2 product commands. 
the cluster ones are all kubectl operations.

shared/product.go
aganesh-suse: secretsEncryptStdOut ?

shared/product.go
fmoral2: agree, my point was like , they are related to product only because of how we deal stuff in code and cause we have shared functions to rke2/k3s ..... but they are actually stuff/functionalities/aux  of the cluster itself 

got it ?

fmoral2: like if i would do it now i would put manageservice AND systemctlcmd in AUX.go, AND  rotate and encrypt in cluster

makes sense or no ?

shared/product.go
fmoral2: yeah anything that could relate to the output

shared/product.go
aganesh-suse: aux.go is where we put auxilliary helper stuff. like logging or basic skeleton stuff. putting manageservice and systemcmdctl here - is like stuffing something somewhere cos we dont have a better place to put it. 
and rotate and encrypt is not cluster kubectl ops. i feel it makes sense to give purpose and value to product and k3s/rke2 commands themselves - cos "that is what we are testing". So it makes sense to give importance to what we are testing as products and give them the attention they deserve and keep them at product.go
End of the day, it boils down to perspectives. Its all "shared.x" anyway. 
Either way, it doesnt make sense to change this in this PR. If you still disagree, we should circle back on design for another discussion with team and keep it as a separate PR for a design change.

pkg/testcase/secretsencrypt.go
aganesh-suse: fixed.

pkg/testcase/secretsencrypt.go
aganesh-suse: added

shared/product.go
aganesh-suse: fixed

shared/product.go
aganesh-suse: fixed. returning ""

shared/product.go
aganesh-suse: fixed

shared/cluster.go
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
```diff
@@ -106,28 +105,29 @@ func verifyStatusOutput(action, stdout string) {
 }
 
 func getNodeIps(etcdNodes, cpNodes []shared.Node) []string {
-	var sarray []string
+	var nodeIpsArr []string
```
fmoral2, line 108: sorry think i commented here but seems it doesnt

lets naming better accordingly to use case x language idiomacy 
nodeIPs , its great 

no need to specify the type on the name, execpt in more complex flow or logic 
also in golang this is not an array, its a slice --  to declare an array you need to add the lenght

fmoral2: for sure, it serves for any other places that u might have this also

fmoral2:
last comment for me, and looks great to go ! 
**https://github.com/rancher/distros-test-framework/pull/77#discussion_r1516018889**

entrypoint/secretsencrypt/encryption_suite_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	os.Exit(m.Run())
+}
+
+func TestSecretsEncryptionSuite(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Secrets Encryption Test Suite")
+}
+
```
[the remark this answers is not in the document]
aganesh-suse: Love this change request: 
```
[BeforeSuite] [FAILED] [0.001 seconds]
[BeforeSuite]
/go/src/github.com/rancher/distros-test-framework/entrypoint/secretsencrypt/encryption_suite_test.go:38

  [FAILED] FATAL: Add secrets-encryption:true to server_flags for this test
  Expected
      <string>:
  to contain substring
      <string>: secrets-encryption:
  In [BeforeSuite] at: /go/src/github.com/rancher/distros-test-framework/entrypoint/secretsencrypt/encryption_suite_test.go:44 @ 03/07/24 19:26:02.722
------------------------------

Summarizing 1 Failure:
  [FAIL] [BeforeSuite]
```

pkg/testcase/secretsencrypt.go
aganesh-suse: fixed

entrypoint/secretsencrypt/encryption_suite_test.go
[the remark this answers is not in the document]
aganesh-suse: i went ahead used the same to do a version check and make decision on which test to run. by default run classic method. if not 1.27 or 1.26 run both classic and new methods. (rotate-keys to test/not to test descision)

entrypoint/secretsencrypt/encryption_suite_test.go
```diff
@@ -0,0 +1,62 @@
...
+var _ = BeforeSuite(func() {
+	if err := config.SetEnv(shared.BasePath() + fmt.Sprintf("/config/%s.tfvars", cfg.Product)); err != nil {
+		Expect(err).To(BeNil(), fmt.Sprintf("error loading tf vars: %v\n", err))
+	}
+	Expect(os.Getenv("server_flags")).To(ContainSubstring("secrets-encryption:"),
+		"FATAL: Add secrets-encryption:true to server_flags for this test")
+
+	version := os.Getenv(fmt.Sprintf("%s_version", cfg.Product))
+	if strings.Contains(version, "1.27") || strings.Contains(version, "1.26") {
+		os.Setenv("TEST_TYPE", "classic")
```
mdrahman-suse, line 48: Can we document this env flag in the testing.md and also add what `classic` and `both` means?

entrypoint/secretsencrypt/encryption_suite_test.go
```diff
@@ -0,0 +1,62 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Secrets Encryption Test Suite")
+}
+
+var _ = BeforeSuite(func() {
+	if err := config.SetEnv(shared.BasePath() + fmt.Sprintf("/config/%s.tfvars", cfg.Product)); err != nil {
+		Expect(err).To(BeNil(), fmt.Sprintf("error loading tf vars: %v\n", err))
+	}
+	Expect(os.Getenv("server_flags")).To(ContainSubstring("secrets-encryption:"),
+		"FATAL: Add secrets-encryption:true to server_flags for this test")
```
mdrahman-suse, line 44: Can we use log-level here instead of printing `FATAL...`?

entrypoint/secretsencrypt/encryption_test.go
```diff
@@ -0,0 +1,59 @@
...
+	It("Validate Pods", func() {
+		testcase.TestPodStatus(
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus(),
+		)
+	})
+
+	It("Validate Secrets Encryption", func() {
+		testcase.TestSecretsEncrypt()
```
mdrahman-suse, line 34: nit: Can we rename to `TestSecretsEncryption`?

docs/testing.md
```diff
@@ -29,6 +29,35 @@ agent1 ->  agent/worker node
 
 Note/TODO: k3s external db fails working with etcd only node. Refer: https://docs.k3s.io/datastore/ha
 
+### Secret-Encryption tests Setup Requirements/Assumptions
+We need a split role setup for this test:
```
mdrahman-suse, line 33: Do we want to restrict the tests to have this config? We should be able to run the test in non-split mode as well, right?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,143 @@
...
+	"os"
+	"strings"
+	"time"
+
+	"github.com/rancher/distros-test-framework/shared"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestSecretsEncrypt() {
```
mdrahman-suse, line 14: Will this work if we have a setup without split mode?

scripts/test_runner.sh
```diff
@@ -14,7 +14,7 @@ function validate_test_image() {
 
 function validate_dir(){
   case "$TEST_DIR" in
-       upgradecluster|versionbump|mixedoscluster|dualstack|validatecluster|createcluster|selinux|certrotate|restartservice)
+       upgradecluster|versionbump|mixedoscluster|dualstack|validatecluster|createcluster|selinux|certrotate|secretsencrypt|restartservice)
```
mdrahman-suse, line 17: Nit: This will keep getting bigger, can we multiline it?

entrypoint/secretsencrypt/encryption_suite_test.go
aganesh-suse: Updated testing.md to reflect the same.

scripts/test_runner.sh
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,143 @@
...
+		time.Sleep(10 * time.Second)
+	}
+	switch product {
+	case "k3s":
+		shared.LogLevel("INFO", "Sleep 30 seconds - wait for services to come up")
+		time.Sleep(30 * time.Second)
+	case "rke2":
+		shared.LogLevel("INFO", "Sleep 60 seconds - wait for services to come up")
+		time.Sleep(60 * time.Second)
+	}
```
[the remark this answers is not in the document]
aganesh-suse: added 2 levels of wait/query model. First we do a get pod running status(call to WaitForPodsRunning). After this, we wait for the "All hashes match" ( call to waitForHashMatch). Only after they work, we proceed with further verifications.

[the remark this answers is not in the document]
aganesh-suse: get running pods alone is not enough, since secrets-encryption hash match seems to take longer. hence the 2 level query/wait model.

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,143 @@
...
+		time.Sleep(20 * time.Second)
+	}
+
+	for _, node := range ips {
+		nodearr := []string{node}
+		nodeIp, errRestart := shared.ManageService(product, "restart", "server", nodearr)
+		Expect(errRestart).NotTo(HaveOccurred(), "error restart service for node: "+nodeIp)
+		// Order of reboot matters. Etcd first then control plane nodes.
+		// Little lag needed between node restarts to avoid issues.
+		shared.LogLevel("INFO", "Sleep 10 seconds - wait before restarting next node in cluster")
```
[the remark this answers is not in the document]
aganesh-suse: Fixed.

pkg/testcase/secretsencrypt.go
aganesh-suse: fixed

docs/testing.md
aganesh-suse: fixed. Added more verbage to suggest we can test on both. Made it clear, that the split role is for patch validation test case.

entrypoint/secretsencrypt/encryption_test.go
aganesh-suse: fixed

entrypoint/secretsencrypt/encryption_suite_test.go
aganesh-suse: This message is within Expect - it is not allowing to add an extra shared.LogLevel here. So changed from FATAL to Error. Hope this is okay?

pkg/testcase/secretsencrypt.go
rancher-max: resolved from my perspective but I'll let @mdrahman-suse mark resolved if he feels the same

pkg/testcase/secretsencrypt.go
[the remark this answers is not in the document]
rancher-max: Nice, I like that approach

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,170 @@
...
+	nodes, errGetNodes := shared.GetNodesByRoles("etcd", "control-plane")
+	Expect(nodes).NotTo(BeEmpty())
+	Expect(errGetNodes).NotTo(HaveOccurred(), "error getting etcd/control-plane nodes")
+
+	product, err := shared.Product()
+	Expect(err).NotTo(HaveOccurred(), "error getting product from config")
+
+	ips := getNodeIps(nodes)
+
+	errSecret := shared.CreateSecret("secret1", "default")
```
fmoral2, line 24: can we have break line here ? separating code block concerns 
```

	errSecret := shared.CreateSecret("secret1", "default")
	Expect(errSecret).NotTo(HaveOccurred(), "error creating secret")

	shared.LogLevel("INFO", "TEST: 'CLASSIC' Secrets Encryption method")

	index := len(nodes) - 1
	secretsEncryptOps("prepare", product, nodes[index].ExternalIP, ips)
	secretsEncryptOps("rotate", product, nodes[index].ExternalIP, ips)
	secretsEncryptOps("reencrypt", product, nodes[index].ExternalIP, ips)

```

docs/testing.md
```diff
@@ -29,6 +29,42 @@ agent1 ->  agent/worker node
...
+
+Please set the server_flags in .tfvars file for k3s:
+```
+server_flags   = "secrets-encryption: true\n"
+```
+In case of rke2 - do not leave this empty for now.
+At least set a "token: secret\n" for both server and agent flags as a workaround for:
+https://github.com/rancher/distros-test-framework/issues/86
+
+For versions 1.26 and 1.27 - we run the traditional tests only: prepare/rotate/renenrypt (TEST_TYPE gets set to 'classic' in env var. We use this to determine which tests to run.)
```
fmoral2, line 62: where are we loading this var `TEST_TYPE` ?

fmoral2: on the suite file ?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,170 @@
...
+func secretsEncryptOps(action, product, cpIp string, ips []string) {
+	shared.LogLevel("INFO", fmt.Sprintf("TEST: Secrets-Encryption: %s", action))
+	_, errStatusB4 := shared.SecretEncryptOps("status", cpIp, product)
+	Expect(errStatusB4).NotTo(HaveOccurred(), "error getting secret-encryption status before action")
+
+	stdOutput, err := shared.SecretEncryptOps(action, cpIp, product)
+	Expect(err).NotTo(HaveOccurred(), "error: secret-encryption: "+action)
+	verifyStdOut(action, stdOutput)
+	if (action == "reencrypt") || (action == "rotate-keys") {
+		shared.LogLevel("DEBUG", "reencrypt op needs some time to complete - Sleep for 20 seconds before service restarts")
```
fmoral2, line 47: why this message is DEBUG and not INFO , on your pov ?

aganesh-suse: Well, if someone doesnt want this seen eventually, when we have the option to drop debug log lines.. wanted to leave it in debug instead of info - in case folks may consider such messages noisy and not necessary.

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,170 @@
...
+	for i, node := range ips {
+		nodearr := []string{node}
+		nodeIp, errRestart := shared.ManageService(product, "restart", "server", nodearr)
+		Expect(errRestart).NotTo(HaveOccurred(), "error restart service for node: "+nodeIp)
+		// Order of reboot matters. Etcd first then control plane nodes.
+		// Little lag needed between node restarts to avoid issues.
+		waitEtcdErr := shared.WaitForPodsRunning(5, 4, false)
+		if waitEtcdErr != nil {
+			shared.LogLevel("WARN", "pods not up after 20 seconds.")
+			if i != len(ips) {
```
fmoral2, line 59: this is statement is not always true ?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,170 @@
...
+	switch product {
+	case "k3s":
+		waitPodsErr := shared.WaitForPodsRunning(5, 6, false)
+		if waitPodsErr != nil {
+			shared.LogLevel("WARN", "pods not up after 30 seconds")
+		}
+	case "rke2":
+		waitPodsErr := shared.WaitForPodsRunning(5, 12, false)
+		if waitPodsErr != nil {
+			shared.LogLevel("WARN", "pods not up after 60 seconds")
```
fmoral2, line 73: if the pods are not up after 60 sec, what we are doing with the "error "  ?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,170 @@
...
+	Expect(errLog).NotTo(HaveOccurred())
+}
+
+func waitForHashMatch(cpIp, product string, defaultTime time.Duration, times int) (string, error) {
+	var secretEncryptStatus string
+	var errGetStatus error
+	for i := 1; i <= times; i++ {
+		secretEncryptStatus, errGetStatus = shared.SecretEncryptOps("status", cpIp, product)
+		if errGetStatus != nil {
+			shared.LogLevel("DEBUG", "error getting secret-encryption status. Retry.")
```
fmoral2, line 91: this is not a WARN ?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		if secretEncryptStatus != "" && strings.Contains(secretEncryptStatus, "All hashes match") {
+			shared.LogLevel("DEBUG", "Total sleep time before hashes matched: %d seconds", i*int(defaultTime))
```
fmoral2, line 94: why DEBUG  instead of INFO ?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,170 @@
...
+			return secretEncryptStatus, nil
+		}
+		time.Sleep(defaultTime * time.Second)
+	}
+	shared.LogLevel("WARN", "Hashes did not match after %d seconds", times*int(defaultTime))
```
fmoral2, line 99: can we give a space before return ?

pkg/testcase/secretsencrypt.go
aganesh-suse: ah... good catch.

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,170 @@
...
+
+func TestSecretsEncryption() {
+	nodes, errGetNodes := shared.GetNodesByRoles("etcd", "control-plane")
+	Expect(nodes).NotTo(BeEmpty())
+	Expect(errGetNodes).NotTo(HaveOccurred(), "error getting etcd/control-plane nodes")
+
+	product, err := shared.Product()
+	Expect(err).NotTo(HaveOccurred(), "error getting product from config")
+
+	ips := getNodeIps(nodes)
```
fmoral2, line 22: why getnodeIps if your  `nodes` already has it? 

```
	nodes, errGetNodes := shared.GetNodesByRoles("etcd", "control-plane")

	for _, ip := range nodes {
		ip.ExternalIP = ""

	}

```

pkg/testcase/secretsencrypt.go
aganesh-suse: no. the first few times, we wont get the secret-encryption status. it will be a few tries before the results start coming.

shared/cluster.go
```diff
@@ -633,3 +633,64 @@ func FetchToken(ip string) (string, error) {
...
+	if podReady+podNotReady != len(pods) {
+		LogLevel("DEBUG", "Length of pods %d != Ready pods: %d + Not Ready Pods: %d", len(pods), podReady, podNotReady)
+	}
+	if podNotReady == 0 {
+		podsRunningStatus = true
+	}
+	return podsRunningStatus
+}
+
+// Wait for pods to reach running state.
```
fmoral2, line 681: Comment should have the following format 'WaitForPodsRunning ...' 
the complete name of the func

shared/cluster.go
```diff
@@ -633,3 +633,64 @@ func FetchToken(ip string) (string, error) {
...
+		kubectl, secret, namespace)
+	createStdOut, err := RunCommandHost(cmd)
+	if err != nil {
+		return ReturnLogError("failed to create secret: \n%w", err)
+	}
+	if strings.Contains(createStdOut, "failed to create secret") {
+		return ReturnLogError("failed to create secret: \n%w", err)
+	}
+	LogLevel("DEBUG", "Create Secret Output: %s", createStdOut)
+	return nil
```
fmoral2, line 652: space before return, sorry idk why lint is not working on this, will take a look, but we have a clausule for that there

shared/cluster.go
```diff
@@ -633,3 +633,64 @@ func FetchToken(ip string) (string, error) {
...
+			LogLevel("DEBUG", "Pod Not Ready. Pod details: Name: %s Status: %s", pod.Name, pod.Status)
+		}
+	}
+	if podReady+podNotReady != len(pods) {
+		LogLevel("DEBUG", "Length of pods %d != Ready pods: %d + Not Ready Pods: %d", len(pods), podReady, podNotReady)
+	}
+	if podNotReady == 0 {
+		podsRunningStatus = true
+	}
+	return podsRunningStatus
```
fmoral2, line 678: https://github.com/rancher/distros-test-framework/pull/77/files#r1554261840

shared/cluster.go
```diff
@@ -633,3 +633,64 @@ func FetchToken(ip string) (string, error) {
...
+	cmd := fmt.Sprintf("%s create secret generic %s -n %s --from-literal=mykey=mydata",
+		kubectl, secret, namespace)
+	createStdOut, err := RunCommandHost(cmd)
+	if err != nil {
+		return ReturnLogError("failed to create secret: \n%w", err)
+	}
+	if strings.Contains(createStdOut, "failed to create secret") {
+		return ReturnLogError("failed to create secret: \n%w", err)
+	}
+	LogLevel("DEBUG", "Create Secret Output: %s", createStdOut)
```
fmoral2, line 651: so this is a public function, which means to be consumed.

lets talk about it on monday with everyone, but we should try to only log on the provider if its something very occasional and unexpected or a complicated flow.

so if i am a consumer of this function, and dont want or cant have a log from my secret, wherever the log will be output, i have no option, if i consume this func, mandatory i will have this log .

shared/cluster.go
```diff
@@ -633,3 +633,64 @@ func FetchToken(ip string) (string, error) {
...
+	return nil
+}
+
+func checkPodStatus(print bool) bool {
```
fmoral2, line 655: https://github.com/rancher/distros-test-framework/blob/e189754e84556cbd9d0c9b3b695967d07c665e01/pkg/assert/pod.go#L61
https://github.com/rancher/distros-test-framework/blob/e189754e84556cbd9d0c9b3b695967d07c665e01/pkg/assert/pod.go#L121

shared/product.go
```diff
@@ -103,11 +107,34 @@ func ManageService(product, action, nodeType string, ips []string) (string, erro
...
+
+	secretsEncryptStdOut, err := RunCommandOnNode(secretEncryptCmd[action], ip)
+	if err != nil {
+		return "", ReturnLogError(fmt.Sprintf("secrets-encryption %s action failed", action), err)
+	}
+	if strings.Contains(secretsEncryptStdOut, "fatal") {
+		return "", ReturnLogError(fmt.Sprintf("secrets-encryption %s action failed", action))
+	}
+	LogLevel("DEBUG", "%s output:\n %s", action, secretsEncryptStdOut)
+	return secretsEncryptStdOut, nil
```
fmoral2, line 139: space before return

pkg/testcase/secretsencrypt.go
fmoral2: got it tks

pkg/testcase/secretsencrypt.go
fmoral2: ok tks

pkg/testcase/secretsencrypt.go
fmoral2: ok tks

docs/testing.md
aganesh-suse: https://github.com/rancher/distros-test-framework/pull/77/files#diff-af432fd2bd869597ab2606f5075fa3962df0a81684e1ce8402c6091831d9d7e4R49

aganesh-suse: Yes in the before suite section, based on version, we set the env var.

pkg/testcase/secretsencrypt.go
aganesh-suse: just making my life easier, working with a simple string array, instead of using the whole object. Every time, i ll hv to reference, and look for the external ip. also, the shared methods take string arrays as inputs. just thought, keep the datatype needed simple and uniform across all methods.

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		time.Sleep(defaultTime * time.Second)
+	}
+	shared.LogLevel("WARN", "Hashes did not match after %d seconds", times*int(defaultTime))
+	return secretEncryptStatus, errGetStatus
+}
+
+func verifyStdOut(action, stdout string) {
+	switch action {
+	case "prepare":
```
fmoral2, line 105: so here in `verifyStdOut` and `verifyStatusOutput` are the same "actions"  right, could we think in a better design for this ? 
idk but maybe i structure that holds action or the messages 
map them and check all if exists and if contains what u want  in one single place ? 

this rn is too long and repetitive unnecessary 
if you need ideas let me know i can also help

pkg/testcase/secretsencrypt.go
fmoral2: sorry but this is creating extra functions unnecessary you already have the data u need....on pov makes no sense cause u are already also using it bellow
```
	secretsEncryptOps("prepare", product, nodes[index].ExternalIP, ips)
```

fmoral2: but thats up to you

aganesh-suse: so the original getips method, was because, i was getting the etcd nodes first and then the control plane nodes, and getting their individual ips and stitching them together. but after Max suggested i can get them together, this method became simple. But all methods here and the shared methods are using the ips array. If I modify it now, it will be a major rework of all the methods/calls and retesting. I wont be able to complete it this cycle. Will have to get pushed later. I would like to leave this as is for now.

aganesh-suse: so the original getips method, was because, i was getting the etcd nodes first and then the control plane nodes, and getting their individual ips and stitching them together. but after Max suggested i can get them together, this method became simple. But all methods here and the shared methods are using the ips array. If I modify it now, it will be a major rework of all the methods/calls and retesting. I wont be able to complete it this cycle. Will have to get pushed later. I would like to leave this as is for now.

pkg/testcase/secretsencrypt.go
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
aganesh-suse: No, they are not. 
verifyStdOut is the 'std output' before reboots for different actions - the outputs of secrets-encrypt prepare/rotate etc. 
Then we perform reboot services.  
The verifyStatusOutput is after - when we get the "secrets-encryption status". They have different contents/expectations.

aganesh-suse: Sample Outputs: 
verifyStdOut verifies: 
```
$ sudo k3s secrets-encrypt rotate 
 rotate completed successfully
```

verifyStatusOutput verifies: 
```
$ sudo k3s secrets-encrypt status 
 Encryption Status: Enabled
Current Rotation Stage: rotate
Server Encryption Hashes: All hashes match

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey-2024-04-04T21:30:30Z
        AES-CBC   aescbckey
```

shared/cluster.go
aganesh-suse: fixed

shared/cluster.go
aganesh-suse: added a print bool - to give the option of printing vs not for now. The day we can actually use the implementation for "DEBUG" like its supposed to, this will hopefully not be a problem anymore.

shared/product.go
aganesh-suse: fixed

shared/cluster.go
aganesh-suse: Both of them use Expects - which means, they will fail. These waits are meant for NOT waiting too long, and NOT erroring/failing. It is meant to be forgiving. I cannot use either of those for this reason.

shared/cluster.go
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
aganesh-suse: In verifyStatusOutput, moved 
```
	Expect(stdout).To(ContainSubstring("Encryption Status: Enabled"))
	Expect(stdout).To(ContainSubstring("Server Encryption Hashes: All hashes match"))
```
to outside the switch case. They are common for all actions.

aganesh-suse: Also renamed the methods to avoid confusion and added method comments. 
verifyStdOut -> verifyActionStdOut
verifyStatusOutput -> verifyStatusStdOut

pkg/testcase/secretsencrypt.go
aganesh-suse: fixed

shared/cluster.go
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
fmoral2: > No, they are not. verifyStdOut is the 'std output' before reboots for different actions - the outputs of secrets-encrypt prepare/rotate etc. Then we perform reboot services. The verifyStatusOutput is after - when we get the "secrets-encryption status". They have different contents/expectations.


sorry .... my bad, i was not clear on my comment

i was saying that both functions has the same actions to perform on it , regardless the purpose or whatever ....then based on that my suggestion to improve code design to avoid over usage os switch statement which can easily lead to code duplication, making maintenance harder sometimes. 

so here would be really fit some table driven approach as simpler method, but could also use something like strategy pattern. 

this is only a recommendation, since its a test case file not an implementation of nothing shareable or etc

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		time.Sleep(defaultTime * time.Second)
+	}
+	shared.LogLevel("WARN", "Hashes did not match after %d seconds", times*int(defaultTime))
+
+	return secretEncryptStatus, errGetStatus
+}
+
+// verifyActionStdOut Verifies secrets-encryption action outputs
+// Verfies std outputs of: sudo k3s|rke2 secrets-encryption prepare|rotate|reencrypt|rotate-keys actions
```
fmoral2, line 107: typo - Verfies

shared/cluster.go
```diff
@@ -633,3 +633,69 @@ func FetchToken(ip string) (string, error) {
...
+	}
+	if podNotReady == 0 {
+		podsRunningStatus = true
+	}
+
+	return podsRunningStatus
+}
+
+// WaitForPodsRunning Waits for pods to reach running state.
+func WaitForPodsRunning(defaultTime time.Duration, times int, print bool) error {
```
fmoral2, line 686: every call of this function is passing false to print, 
Screenshot 2024-04-08 at 09 39 21


-  so to avoid unnecessary complexity, on the usage  pass "false". directly 

```
func checkPodStatus(print bool) bool {
	podsRunningStatus := false
	pods, errGetPods := GetPods(false)
```

shared/cluster.go
fmoral2: ok but the usage of this func which is on the `WaitForPodsRunning` on line 696 
```
if !podsRunning {
		return ReturnLogError("All pods were not up at the end of wait period %d", int(defaultTime)*times)
	}
```

you are returning error if they are not up in that period of time, which is exactly what those functions do,  right ?

if your concern its the duration , thats a design discussion now to improve that, not duplicating code .

if am still missing something please let me know.

shared/cluster.go
fmoral2: so you mentioned somewhere else ( which i totally agree ) that we in the future will have a "start option" mode to differentiate what to log or not

we should deal that on that implementation, not adding arbitrary ifs boolean in the code 

also adding the print does not solve the issue i mentioned , its just a workaround that would need refactor later on 

 + you are adding unnecessary signature complexity in favor of a just workaround only

shared/cluster.go
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,161 @@
...
+	case "rotate":
+		Expect(stdout).To(ContainSubstring("rotate completed successfully"))
+	case "reencrypt":
+		Expect(stdout).To(ContainSubstring("reencryption started"))
+	case "rotate-keys":
+		Expect(stdout).To(ContainSubstring("keys rotated, reencryption started"))
+	}
+}
+
+// verifyStatusStdOut Verfies secrets-encryption status outputs post different actions.
```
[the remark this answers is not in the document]
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
aganesh-suse: i feel like thats overkill for the purpose here, plus they r methods no one else will re-use - ever - it belongs within testfile like you mentioned. Will keep this suggestion in mind for the future. Thanks.

fmoral2: no p at all

shared/cluster.go
fmoral2: idk why but my comments was not uploaded here 

so , the user of this is `WaitForPodsRunning` its returning error here if the pods are not running, which is exact the same thing on the exist functions already do.  if your concern its the time duration, this should be a design enhancement but not creating extra stuff duplicating and doing exact the same only changing the duration. 
```
	}
	if !podsRunning {
		return ReturnLogError("All pods were not up at the end of wait period %d", int(defaultTime)*times)
	}
```

fmoral2: i think i am not understanding well the requirement sorry 
 can u explain me? why then u need all of this, if you are actually doing nothing with the result ( success or not ) ?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,161 @@
...
+
+	secretEncryptStatus, errGetStatus := waitForHashMatch(cpIp, product, 5, 36) // Max 3 minute wait time for hash to match
+	Expect(errGetStatus).NotTo(HaveOccurred(), "error getting secret-encryption status")
+	verifyStatusStdOut(action, secretEncryptStatus)
+
+	errLog := logEncryptionFileContents(nodes, product)
+	Expect(errLog).NotTo(HaveOccurred())
+}
+
+func waitForHashMatch(cpIp, product string, defaultTime time.Duration, times int) (string, error) {
```
fmoral2, line 86: if this is a private function, and you are only using in one context, why pass default time as parameter instead of just use the value u want on the sleep ?

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,161 @@
...
+			shared.LogLevel("WARN", "pods not up after 30 seconds")
+		}
+	case "rke2":
+		waitPodsErr := shared.WaitForPodsRunning(5, 12)
+		if waitPodsErr != nil {
+			shared.LogLevel("WARN", "pods not up after 60 seconds")
+		}
+	}
+
+	secretEncryptStatus, errGetStatus := waitForHashMatch(cpIp, product, 5, 36) // Max 3 minute wait time for hash to match
```
fmoral2, line 78: why this 36 times?

shared/cluster.go
rancher-max: This came from my ask -- the reasoning was so that we could have a "wait up to X seconds for this condition" as previously this was just "sleep X" where X was an arbitrary number that had worked in local tests but might not work always.

PodStatusRunning assertion does wait up to 30 seconds, so it's possible that we can still use that but maybe increase the timeout on that function to match what is needed here in the maximum case. I'm fine either way and will let @aganesh-suse decide what's best there.

rancher-max (dismissed):

shared/cluster.go
aganesh-suse: We had a hard-coded wait time before. Max suggested we avoid this. 
We need time buffers between node restarts. And post node restart - we have to wait before we run the secrets-encryption test. Otherwise the tests will fail. How do we quantify this time? How long a wait is a good time? 
We could query for the pods to come up. - But this is not enough. Since the secrets-encryption processes after the pods come up. So then we query for the hash match as well - before we start the test. This ends up being a wait for pods running + wait for hash match. 

We are waiting till the the pods come up. If they come up quickly, `we save some wait time` and can get on with the test faster.  Instead of a blind wait for 30 seconds say, we now `monitor & log` - which reduces wait time so far to like 15 to 20 seconds in different test runs. 
Also, this is a way to simply `know the pod status before we started the secret-encryption test`. 
 If they dont come up by the time, that is fine too. since we will do restarts anyway for other servers. 
Plus we have a wait time for the hash match later - which will be enough time for the pods to have come up. 

Also, at the end we have the Test for Pod Status - which will cover the error for pods not up after secret encryption.

pkg/testcase/secretsencrypt.go
aganesh-suse: we should never hard code. that is not good practice. even if it is only used in one place. 
also it follows with the wait pod method pattern. so it is not confusing to people.

pkg/testcase/secretsencrypt.go
aganesh-suse: to we have a max wait time of 3 minutes - as mentioned in the comment in the code.

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,161 @@
...
+	for i, node := range nodes {
+		nodearr := []string{node.ExternalIP}
+		nodeIp, errRestart := shared.ManageService(product, "restart", "server", nodearr)
+		Expect(errRestart).NotTo(HaveOccurred(), "error restart service for node: "+nodeIp)
+		// Order of reboot matters. Etcd first then control plane nodes.
+		// Little lag needed between node restarts to avoid issues.
+		waitEtcdErr := shared.WaitForPodsRunning(5, 4)
+		if waitEtcdErr != nil {
+			shared.LogLevel("WARN", "pods not up after 20 seconds.")
+			if i != len(nodes)-1 {
```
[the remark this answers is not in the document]
aganesh-suse: nope it was for logging when the restarts where continuing and it would not have logged for the last time - when there is no more node restarts happening. but i removed this log line to avoid any confusion for anyone.

rancher-max (approved):

pkg/testcase/secretsencrypt.go
fmoral2: sorry this is not hard coding....actually thinking on your perspctive what u doing is hard coding as well, since the only difference is that instead of making it simple and where it should be used its passed already, you are passing a value that never changes as parameter..... so in your POV this is also  hard coding, but instead of declaring it on the usage you are declaring elsewhere, and passing it along.

but regardless, this is not hard coding at all, this is you using a value that has no interaction anywhere else, like no one in any way can change this value right? so you are just coding something and using a spefic value for it, which is totally normal. Now if you have a value that can/should/probably needs to be changed,overwrited or whatever , and you are adding the value fixed there, THEN its hard coding.

shared/cluster.go
fmoral2: ok , if we are using this a shared stuff , we need to contemplate other conditionals as well like the ones from here 
https://github.com/rancher/distros-test-framework/blob/e189754e84556cbd9d0c9b3b695967d07c665e01/pkg/testcase/pod.go#L34

shared/cluster.go
```diff
@@ -633,3 +633,70 @@ func FetchToken(ip string) (string, error) {
...
+	pods, errGetPods := GetPods(print)
+	if errGetPods != nil || len(pods) == 0 {
+		LogLevel("DEBUG", "Error getting pods. Retry.")
+		return podsRunningStatus
+	}
+	podReady := 0
+	podNotReady := 0
+	for _, pod := range pods {
+		if pod.Status == "Running" || pod.Status == "Completed" {
+			podReady = podReady + 1
```
fmoral2, line 669: lets simplify like
` podReady++` 

`podNotReady++`

shared/cluster.go
```diff
@@ -633,3 +633,70 @@ func FetchToken(ip string) (string, error) {
...
+		podsRunningStatus = true
+	}
+
+	return podsRunningStatus
+}
+
+// WaitForPodsRunning Waits for pods to reach running state.
+func WaitForPodsRunning(defaultTime time.Duration, times int) error {
+	var podsRunning bool
+	print := false
```
fmoral2, line 688: no need of this var 

`podsRunning = checkPodStatus(false)`

docs/testing.md
mdrahman-suse, line 62: Typo `reencrypt`

entrypoint/secretsencrypt/encryption_suite_test.go
mdrahman-suse: Ty

pkg/testcase/secretsencrypt.go
```diff
@@ -0,0 +1,158 @@
...
+	Expect(nodes).NotTo(BeEmpty())
+	Expect(errGetNodes).NotTo(HaveOccurred(), "error getting etcd/control-plane nodes")
+
+	product, err := shared.Product()
+	Expect(err).NotTo(HaveOccurred(), "error getting product from config")
+
+	errSecret := shared.CreateSecret("secret1", "default", false)
+	Expect(errSecret).NotTo(HaveOccurred(), "error creating secret")
+
+	shared.LogLevel("INFO", "TEST: 'CLASSIC' Secrets Encryption method")
```
mdrahman-suse, line 25: `INFO` Should be `info` as defined in the LogLevel func

shared/cluster.go
aganesh-suse: The etcd node pods wont come up until the control node plane joins the cluster. If we assert fail for pods running here, it would have been premature - and we would have a failed test for a non-bug.

fmoral2: so if this  is too specific for your use case, this 2 funcs should live on the test case file, not here.

rancher-max: I feel like it's fine in this cluster.go file in this case -- since this is a shared function that is at cluster level operations. It's not really a testcase since there aren't any `Expect` statements

pkg/testcase/secretsencrypt.go
rancher-max: yeah there's not a right or wrong answer here, more of a "which paradigm do we want to use"? In the case of `waitForX` functions, we have one already merged in the framework today `WaitForInstanceRunning` which sets the time vars in the function. This PR has two new ones: `waitForHashMatch` and `WaitForPodsRunning` which both set the time vars as part of the function signature. 

Having it as part of the function signature allows other areas to call with different wait times. Having it as defined within the function says "wait N time for X". If we expect to use these functions a bit more, I personally like having it as part of the function signature, because it's likely that not everything will need to wait the same amount of time. But if we don't expect to use it anymore, or if we expect that future uses require the same amount of wait time, then we should definitely define the time within the function itself.

shared/cluster.go
```diff
@@ -633,3 +633,70 @@ func FetchToken(ip string) (string, error) {
...
+	}
+	if podNotReady == 0 {
+		podsRunningStatus = true
+	}
+
+	return podsRunningStatus
+}
+
+// WaitForPodsRunning Waits for pods to reach running state.
+func WaitForPodsRunning(defaultTime time.Duration, times int) error {
```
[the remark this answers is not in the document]
rancher-max: Sleep should be fine here no? It's still idiomatic to use sleep to wait a certain number of seconds and that seems the most simple to me 👀 Ticker is a little bit different in that it takes into account execution time of the code, but I think in this case we're fine if we wait a little bit longer as well so sleep should still be good!

shared/cluster.go
fmoral2: not a test case but a specific auxiliar to it ...so should live in the same file

if we have at least one other context,scneario,test that could use this func, then its ok to be here. but as far i am seeing, difficult to other place be using it, except the encryp

shared/cluster.go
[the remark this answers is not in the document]
fmoral2: agree , ticker does not apply that much here no ...but the retry, retry delay would fit nicer

pkg/testcase/secretsencrypt.go
fmoral2: no the point here its that this func its not public, so it will only be consumed in this specific case, and on this specific case its been used a single value with no way to override or change...
so no need to pass as parameter, it only complicates more the scenario

rancher-max: That's true yeah, this function itself is private, indicating it will only be used here, so I agree the time parameters should be defined within the function!

mdrahman-suse:
@aganesh-suse please address the comments and resolve the merge conflicts

entrypoint/secretsencrypt/encryption_test.go
```diff
@@ -0,0 +1,59 @@
...
+	It("Validate Nodes", func() {
+		testcase.TestNodeStatus(
+			assert.NodeAssertReadyStatus(),
+			nil,
+		)
+	})
+
+	It("Validate Pods", func() {
+		testcase.TestPodStatus(
+			assert.PodAssertRestart(),
```
fmoral2, line 46: ,

briandowns:
Non-blocking things I noticed:

Use `+` to combine strings rather than `fmt.printf(...)`. You will see a noticeable change in performance. If you're curious why, I'm happy to break it down further.

Use the "0 value" for variable declarations to avoid "magic numbers". E.g. Prefer `var x int` over `x := 0`, prefer `var t bool` over `t := false`, prefer `var s []string` over `s := []string{}`. The code looks and reads cleaner and more consistent.

fmoral2 (approved):

fmoral2:
have approved only just to remove the changes request block.

entrypoint/secretsencrypt/encryption_test.go
aganesh-suse: please elaborate on the change needed here?

shared/cluster.go
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
aganesh-suse: fixed

shared/cluster.go
aganesh-suse: fixed

shared/cluster.go
```diff
@@ -633,3 +633,70 @@ func FetchToken(ip string) (string, error) {
...
+
+	return nil
+}
+
+func checkPodStatus(print bool) bool {
+	podsRunningStatus := false
+	pods, errGetPods := GetPods(print)
+	if errGetPods != nil || len(pods) == 0 {
+		LogLevel("DEBUG", "Error getting pods. Retry.")
+		return podsRunningStatus
```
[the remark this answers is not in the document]
aganesh-suse: fixed

shared/cluster.go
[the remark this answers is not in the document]
aganesh-suse: fixed

shared/cluster.go
aganesh-suse: fixed

docs/testing.md
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
aganesh-suse: fixed

aganesh-suse:
> [#77 (comment)](https://github.com/rancher/distros-test-framework/pull/77#discussion_r1555796980)

fixed

aganesh-suse:
> @aganesh-suse please address the comments and resolve the merge conflicts

fixed

shared/cluster.go
```diff
@@ -655,3 +656,64 @@ func PrintGetAll() {
...
 		"  -------------------\n\n%v\n\n\n\n", res)
 }
+
+func CreateSecret(secret, namespace string) error {
+	kubectl := fmt.Sprintf("kubectl --kubeconfig %s", KubeConfigFile)
+	if namespace == "" {
+		namespace = "default"
+	}
+	cmd := fmt.Sprintf("%s create secret generic %s -n %s --from-literal=mykey=mydata",
+		kubectl, secret, namespace)
```
fmoral2, line 666: can we add a break line here? lets try at least on shared stuff maintain code blocks with context of usage and etc , gives a better and clear reading of the code.

shared/cluster.go
```diff
@@ -655,3 +656,64 @@ func PrintGetAll() {
 	fmt.Printf("\n\n\n-----------------  Results from kubectl get all -A -o wide"+
 		"  -------------------\n\n%v\n\n\n\n", res)
 }
+
+func CreateSecret(secret, namespace string) error {
```
fmoral2, line 660: if am using this function somewhere else, and forget to send secret or send a empty string, is that ok ? 
or should we validate that ?

shared/cluster.go
```diff
@@ -655,3 +656,64 @@ func PrintGetAll() {
...
+		return ReturnLogError("failed to create secret: \n%w", err)
+	}
+	if strings.Contains(createStdOut, "failed to create secret") {
+		return ReturnLogError("failed to create secret: \n%w", err)
+	}
+
+	return nil
+}
+
+func checkPodStatus() bool {
```
fmoral2, line 678: some space blocks here as well 
```
func checkPodStatus() bool {
	pods, errGetPods := GetPods(false)
	if errGetPods != nil || len(pods) == 0 {
		LogLevel("DEBUG", "Error getting pods. Retry.")
		return false
	}

	podReady := 0
	podNotReady := 0
	for _, pod := range pods {
		if pod.Status == "Running" || pod.Status == "Completed" {
			podReady++
		} else {
			podNotReady++
			LogLevel("DEBUG", "Pod Not Ready. Pod details: Name: %s Status: %s", pod.Name, pod.Status)
		}
	}

	if podReady+podNotReady != len(pods) {
		LogLevel("DEBUG", "Length of pods %d != Ready pods: %d + Not Ready Pods: %d", len(pods), podReady, podNotReady)
	}
	if podNotReady == 0 {
		return true
	}

	return true
}

```

shared/cluster.go
```diff
@@ -655,3 +656,64 @@ func PrintGetAll() {
...
+
+	return true
+}
+
+// WaitForPodsRunning Waits for pods to reach running state.
+func WaitForPodsRunning(defaultTime time.Duration, attempts uint) error {
+	return retry.Do(
+		func() error {
+			if !checkPodStatus() {
+				return ReturnLogError("Not all pods are ready yet")
```
fmoral2, line 709: error messages should be lower case ... "not all.... "

shared/cluster.go
fmoral2: so here , on shared pkg we should have functions that can be used in elsewhere, this function in conjunction with `WaitForPodsRunning` can we use in other places ? 
because as i can see its a very specific use case for the encryption test right ? 

@rancher-max  @aganesh-suse

shared/product.go
```diff
@@ -78,10 +79,13 @@ func ManageService(product, action, nodeType string, ips []string) (string, erro
 		if getError != nil {
 			return ip, getError
 		}
-		_, err := RunCommandOnNode(cmd, ip)
+		manageServiceOut, err := RunCommandOnNode(cmd, ip)
 		if err != nil {
 			return ip, err
 		}
+		if manageServiceOut != "" {
```
fmoral2, line 86: this debug stuff you deleted from the other thread but added here on the same situation?

shared/product.go
```diff
@@ -91,11 +95,36 @@ func ManageService(product, action, nodeType string, ips []string) (string, erro
 func CertRotate(product string, ips []string) (string, error) {
 	for _, ip := range ips {
 		cmd := fmt.Sprintf("sudo %s certificate rotate", product)
-		_, err := RunCommandOnNode(cmd, ip)
+		certRotateOut, err := RunCommandOnNode(cmd, ip)
 		if err != nil {
 			return ip, err
 		}
+		LogLevel("DEBUG", "On %s, cert rotate output:\n %s", ip, certRotateOut)
```
fmoral2, line 102: same - https://github.com/rancher/distros-test-framework/pull/77/files#r1592959926

shared/product.go
```diff
@@ -91,11 +95,36 @@ func ManageService(product, action, nodeType string, ips []string) (string, erro
...
+	}
+
+	secretsEncryptStdOut, err := RunCommandOnNode(secretEncryptCmd[action], ip)
+	if err != nil {
+		return "", ReturnLogError(fmt.Sprintf("secrets-encryption %s action failed", action), err)
+	}
+	if strings.Contains(secretsEncryptStdOut, "fatal") {
+		return "", ReturnLogError(fmt.Sprintf("secrets-encryption %s action failed", action))
+	}
+	LogLevel("DEBUG", "%s output:\n %s", action, secretsEncryptStdOut)
```
fmoral2, line 127: https://github.com/rancher/distros-test-framework/pull/77/files#r1592959926

shared/product.go
```diff
@@ -91,11 +95,36 @@ func ManageService(product, action, nodeType string, ips []string) (string, erro
 func CertRotate(product string, ips []string) (string, error) {
```
fmoral2, line 95: also here, i know thats not the context now, but we are not validating if by any chance we receive ips empty

shared/product.go
aganesh-suse: you wanted to remove the stdout for RunCommand Logging for security purposes. Your recommendation was to add better verbage per the action. And hence am adding during the actual action for context and debugging purposes.

shared/product.go
aganesh-suse: https://github.com/rancher/distros-test-framework/pull/77/files#r1597120785

shared/product.go
aganesh-suse: https://github.com/rancher/distros-test-framework/pull/77/files#r1597120785

shared/cluster.go
aganesh-suse: fixed

rancher-max (approved):

fmoral2 (approved):

PR #79, opened 2024-03-11
 Refactor dependency injection - cluster and env 

fmoral2:
#### Proposed Changes ####

- https://github.com/rancher/qa-tasks/issues/1069 

- This has many files changed cause it needed to change due to dependency, but whats changed in fact is very small changes:
- Env and Cluster dependencies

- Refactor how we inject dependencies in a way of create a more standardized and clear model which basically is:

    . Send dependencies top down injecting in the main and passing where needed

  Benefits:

   . Have a better visibility and understanding of your deps.
   . Have a easier way to follow through to debug and maintain
   . Standardize
   . Have deps available without need to duplicate code
   . Simplify process of adding dependencies ( now you just need to add and pass along in the main)

   

#### Testing ####

Checklist:

- yes 

- yes 

- will open a PR only for README 
- n/a

- yes 

- yes


fmoral2:
https://github.com/rancher/qa-tasks/issues/1069

rancher-max:
Needs rebase on main again!

rancher-max (commented):
I think this looks good after rebase. We might not want to delete all the versionbump tests here -- I know the versionbump PR updated those so perhaps after rebase that'll be resolved, otherwise maybe keep canal and cilium tests there?

fmoral2:
> I think this looks good after rebase. We might not want to delete all the versionbump tests here -- I know the versionbump PR updated those so perhaps after rebase that'll be resolved, otherwise maybe keep canal and cilium tests there?

it will all be solved after rebase yep

fmoral2:
> Needs rebase on main again!

i will wait all PRs get merged, so i dont complicate people lives with this updates.

VestigeJ (dismissed):

config/reader.go
```diff
@@ -4,57 +4,69 @@ import (
...
 	"strings"
 	"sync"
+
+	"github.com/rancher/distros-test-framework/pkg/logger"
 )
 
 var (
 	product *Product
 	once    sync.Once
+	l       = logger.AddLogger()
```
mdrahman-suse, line 18: can we rename it to `logger`

fmoral2: here we cant, its the pkg name

i can rename to `log`, if thats better

entrypoint/createcluster/createcluster_suite_test.go
```diff
@@ -7,23 +7,21 @@ import (
...
 	flag.Parse()
 
-	_, err = shared.EnvConfig()
-	if err != nil {
-		return
-	}
+	cluster = factory.ClusterConfig()
 
-	os.Exit(m.Run())
+	exitCode := m.Run()
```
mdrahman-suse, line 23: Any reason why we are adding a variable here and replacing the one liner?

entrypoint/deployrancher/rancher_suite_test.go
```diff
@@ -17,12 +17,11 @@ import (
 )
 
 var (
-	cfg   *config.Product
-	flags *customflag.FlagConfig
+	cluster *factory.Cluster
+	flags   *customflag.FlagConfig
```
mdrahman-suse, line 21: Nit: extra spacing?

fmoral2: why ?  they are on the same level 

```
var (
	cluster *factory.Cluster
	flags   *customflag.FlagConfig
)
```

entrypoint/createcluster/createcluster_suite_test.go
fmoral2: yeah its just a cleaner way to not tie this error directly , like first get whats happening then you exit
so if you need to deal with some extra test stuff, like log ,clean something or whatever you have a way to do so +

right now, makes no diff logically, just make more readable

rancher-max: I was going to comment the same thing 😅 I prefer the one liner myself, but I'm fine either way here.

fmoral2: well let me know....this PR its really getting stuck here

fmoral2: here on the PRs, not on this comment i meant

VestigeJ: I also think we should keep the one-liner os.Exit(m.Run())

factory/cluster.go
```diff
@@ -4,29 +4,32 @@ import (
...
 	"github.com/gruntwork-io/terratest/modules/terraform"
 
-	"github.com/rancher/distros-test-framework/shared"
-
-	. "github.com/onsi/ginkgo/v2"
+	"github.com/rancher/distros-test-framework/config"
+	"github.com/rancher/distros-test-framework/pkg/logger"
 )
 
+var l = logger.AddLogger()
```
mdrahman-suse, line 17: Rename to log

pkg/assert/node.go
```diff
@@ -28,7 +28,7 @@ func NodeAssertVersionTypeUpgrade(c customflag.FlagConfig) NodeAssertFunc {
 
 // assertVersion returns the NodeAssertFunc for asserting version
 func assertVersion(c customflag.FlagConfig) NodeAssertFunc {
-	fmt.Printf("Asserting Version: %s\n", c.InstallMode.Version)
+	shared.LogLevel("\ninfo", "Asserting Version: %s\n", c.InstallMode.Version)
```
mdrahman-suse, line 31: typo `\ninfo`?

pkg/assert/pod.go
```diff
@@ -101,7 +102,7 @@ func ValidatePodIPByLabel(labels, expected []string) {
 func ValidatePodIPsByLabel(label string, expected []string) {
 	cmd := "kubectl get pods -l " + label +
 		` -o jsonpath='{range .items[*]}{.status.podIPs[*].ip}{" "}{end}'` +
-		" --kubeconfig=" + shared.KubeConfigFile
+		" --kubeconfig=" + factory.KubeConfigFile
```
mdrahman-suse, line 105: can we store and fetch KubeConfigFile from cluster var?

entrypoint/createcluster/createcluster_suite_test.go
fmoral2: sure, removing then

factory/config.go
```diff
@@ -30,11 +30,13 @@ type Cluster struct {
 	NumAgents     int
 	FQDN          string
 	Config        clusterConfig
-	AwsEc2        awsEc2Config
+	AwsConfig     awsConfig
```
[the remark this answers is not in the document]
fmoral2: i think it was my bad on refactor other stuff naming , i really agree also. fixed tks

pkg/assert/node.go
fmoral2: nope, its to break line before the log

pkg/assert/pod.go
fmoral2: this one as per how its used now, i would say lets keep global

pkg/assert/node.go
mdrahman-suse: should probably use Enums/Constants for represent the log level

fmoral2: could be a possible refactor in future

factory/cluster.go
```diff
@@ -88,29 +96,31 @@ func newCluster(g GinkgoTInterface) (*Cluster, error) {
...
+func DestroyCluster() (string, error) {
+	cfg, err := config.AddEnv()
 	if err != nil {
 		return "", err
 	}
-	varDir, err = filepath.Abs(shared.BasePath() +
+
+	_, callerFilePath, _, _ := runtime.Caller(0)
+	dir := filepath.Join(filepath.Dir(callerFilePath), "..")
+	varDir, err := filepath.Abs(dir +
```
rancher-max, line 107: possible to use BasePath function here at all?

rancher-max (dismissed):
Most of the stuff I have is nitpicky, so I've approved unless you want to talk about any of it or change any of it based on the comments.

Good job on this! Lots of refactoring for it all.

entrypoint/upgradecluster/upgradesuc_test.go
```diff
@@ -54,41 +57,43 @@ var _ = Describe("SUC Upgrade Tests:", func() {
...
 		It("Verifies Traefik IngressRoute before upgrade using old GKV", func() {
-			testcase.TestIngressRoute(true, false, "traefik.containo.us/v1alpha1")
+			testcase.TestIngressRoute(cluster, true, false, "traefik.containo.us/v1alpha1")
 		})
 	}
 
 	It("\nUpgrade via SUC", func() {
-		err := testcase.TestUpgradeClusterSUC(customflag.ServiceFlag.SUCUpgradeVersion.String())
-		Expect(err).NotTo(HaveOccurred())
+		shared.PrintClusterState()
```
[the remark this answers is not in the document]
fmoral2: no, lets remove!  tks

shared/cluster.go
```diff
@@ -47,21 +43,32 @@ func ManageWorkload(action string, workloads ...string) error {
...
+		return err
+	}
+
+	if envErr := config.SetEnv(BasePath() + fmt.Sprintf("/config/%s.tfvars",
+		prod)); envErr != nil {
+		return ReturnLogError("error setting env: %w\n", envErr)
+	}
+	arch := os.Getenv("arch")
+
+	resourceDir := BasePath() + "/workloads/" + arch
```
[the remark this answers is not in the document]
fmoral2: oh please no ! lol 

i added like this to prevent adding a "random" param on the signature, keeping this lighter , since this function its called a lot of times everywhere 
ans this is not a optional var today, so we are always going to have this env arch

factory/config.go
```diff
@@ -57,34 +59,20 @@ type generalConfig struct {
...
-	cfg, err := loadConfig()
-	if err != nil {
-		return nil, "", shared.ReturnLogError("error loading config: %w\n", err)
-	}
-
-	var varDir string
-	var tfDir string
+func addTerraformOptions(cfg *config.Product) (*terraform.Options, string, error) {
+	_, callerFilePath, _, _ := runtime.Caller(0)
+	dir := filepath.Join(filepath.Dir(callerFilePath), "..")
```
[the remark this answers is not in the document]
fmoral2: today we cant cause of circular dependency between them

```
# github.com/rancher/distros-test-framework/entrypoint/createcluster
package github.com/rancher/distros-test-framework/entrypoint/createcluster (test)
        imports github.com/rancher/distros-test-framework/factory
        imports github.com/rancher/distros-test-framework/shared
        imports github.com/rancher/distros-test-framework/factory: import cycle not allowed
```

factory/cluster.go
fmoral2: https://github.com/rancher/distros-test-framework/pull/79#discussion_r1638123337

rancher-max (approved):

mdrahman-suse:
I see the use of this `cluster` variable in almost all the the tests so I was wondering if it makes sense to have it initialized as a global var during the test lifecycle instead of needing to pass it along every time we write the test. Sharing my thought and we can discuss if needed...

fmoral2:
> I see the use of this `cluster` variable in almost all the the tests and suite so I was wondering if it makes sense to have it initialized as a global var during the test lifecycle instead of needing to pass it along every time we write the test. Sharing my thought and we can discuss if needed...


 what are the gains you see initializing  it somewhere else and injecting here as global?

mdrahman-suse:
> > I see the use of this `cluster` variable in almost all the the tests and suite so I was wondering if it makes sense to have it initialized as a global var during the test lifecycle instead of needing to pass it along every time we write the test. Sharing my thought and we can discuss if needed...
> 
> what are the gains you see initializing it somewhere else and injecting here as global?

The gain from my perspective is not to provide that var each time we are defining the tests as we have to use that var in the test anyways. This thought came to my mind as I have used DI with Java before and remember to simply inject the dependency (as an interface) on the test file itself (for us it is similar to testcase pkg).

fmoral2:
> > > I see the use of this `cluster` variable in almost all the the tests and suite so I was wondering if it makes sense to have it initialized as a global var during the test lifecycle instead of needing to pass it along every time we write the test. Sharing my thought and we can discuss if needed...
> > 
> > 
> > what are the gains you see initializing it somewhere else and injecting here as global?
> 
> The gain from my perspective is not to provide that var each time we are defining the tests as we have to use that var in the test anyways. This thought came to my mind as I have used DI with Java before and remember to simply inject the dependency (as an interface) on the test file itself (for us it is similar to testcase pkg). Sharing an example here with Java https://www.vogella.com/tutorials/DependencyInjection/article.html Again this is just as I understand DI

so we dont have that kind of stuff in go because one of the ideas is exactly do the opposite of that like prefering using more explicitness and simplicity, like creating your stuff , handling yourself and passing it through ...making things easier to track .
```
New( my interface for some stuff  ) *stuff 
{. return the interface of my stuff } 
```
  those kind of annotations from that link also for me just bring more complexity cause it can hide potential data or stuff from your obecjt

mdrahman-suse:
> > > > I see the use of this `cluster` variable in almost all the the tests and suite so I was wondering if it makes sense to have it initialized as a global var during the test lifecycle instead of needing to pass it along every time we write the test. Sharing my thought and we can discuss if needed...
> > > 
> > > 
> > > what are the gains you see initializing it somewhere else and injecting here as global?
> > 
> > 
> > The gain from my perspective is not to provide that var each time we are defining the tests as we have to use that var in the test anyways. This thought came to my mind as I have used DI with Java before and remember to simply inject the dependency (as an interface) on the test file itself (for us it is similar to testcase pkg). Sharing an example here with Java https://www.vogella.com/tutorials/DependencyInjection/article.html Again this is just as I understand DI
> 
> so we dont have that kind of stuff in go because one of the ideas is exactly do the opposite of that like prefering using more explicitness and simplicity, like creating your stuff , handling yourself and passing it through ...making things easier to track .
> 
> ```
> New( my interface for some stuff  ) *stuff 
> {. return the interface of my stuff } 
> ```
> 
> those kind of annotations from that link also for me just bring more complexity cause it can hide potential data or stuff from your obecjt

Gotcha! Thanks for clarifying. Looks great overall 👍

mdrahman-suse (approved):
LGTM as long as it has been rebased with the recent changed and ran on Jenkins.

mdrahman-suse:
I think we should run the batch test on this branch before we merge, WDYT?

rancher-max:
Good idea -- I can do that when I run the tests on the RCs tomorrow!

fmoral2:
> I think we should run the batch test on this branch before we merge, WDYT? CC. @rancher-max @fmoral2

good idea

rancher-max (dismissed):

mdrahman-suse (approved):
LGTM

rancher-max (approved):

PR #85, opened 2024-04-03
Update helm repo for rancher 2.7

mdrahman-suse:
#### Proposed Changes ####

This change adds the ability to add helm args on deploy rancher cluster creation. It is needed to enable creating rancher cluster with user provided helm args.

- May be, good to document

#### Types of Changes ####

- Improvement on an existing functionality

#### Testing ####

Checklist:
- Ok

- Ok

- TBA

- Ok

- TBA

- TBA

### DIFF
--- a/entrypoint/deployrancher/rancher_suite_test.go
+++ b/entrypoint/deployrancher/rancher_suite_test.go
@@ -30,7 +30,6 @@ func TestMain(m *testing.M) {
 	flag.StringVar(&flags.ExternalFlag.HelmChartsFlag.RepoName, "chartsRepoName", "rancher-latest", "rancher helm repo name")
 	flag.StringVar(&flags.ExternalFlag.HelmChartsFlag.RepoUrl, "chartsRepoUrl", "https://releases.rancher.com/server-charts/latest", "rancher helm repo url")
 	flag.StringVar(&flags.ExternalFlag.HelmChartsFlag.Args, "chartsArgs", "", "rancher helm additional args, comma separated")
-	flag.StringVar(&flags.ExternalFlag.RancherImage, "rancherImage", "", "rancher image name or registry url")
 	flag.StringVar(&flags.ExternalFlag.RancherVersion, "rancherVersion", "v2.8.0", "rancher version that will be deployed on the cluster")
 	flag.Parse()
 
--- a/entrypoint/deployrancher/rancher_suite_test.go
+++ b/entrypoint/deployrancher/rancher_suite_test.go
@@ -23,6 +23,7 @@ func TestMain(m *testing.M) {
 	flag.Var(&customflag.ServiceFlag.ClusterConfig.Destroy, "destroy", "Destroy cluster after test")
 	flag.StringVar(&customflag.ServiceFlag.ExternalFlag.CertManagerVersion, "certManagerVersion", "v1.11.0", "cert-manager version that will be deployed on the cluster")
 	flag.StringVar(&customflag.ServiceFlag.ExternalFlag.RancherHelmVersion, "rancherHelmVersion", "v2.8.0", "rancher helm chart version to use to deploy rancher manager")
+	flag.StringVar(&customflag.ServiceFlag.ExternalFlag.RancherHelmArgs, "rancherHelmArgs", "", "rancher helm additional args that can be added on the rancher manager")
 	flag.StringVar(&customflag.ServiceFlag.ExternalFlag.RancherImageVersion, "rancherImageVersion", "v2.8.0", "rancher version that will be deployed on the cluster")
 	flag.Parse()
 
--- a/entrypoint/deployrancher/rancher_suite_test.go
+++ b/entrypoint/deployrancher/rancher_suite_test.go
@@ -16,15 +16,22 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var cfg *config.Product
+var (
+	cfg   *config.Product
+	flags *customflag.FlagConfig
+)
 
 func TestMain(m *testing.M) {
 	var err error
-	flag.Var(&customflag.ServiceFlag.ClusterConfig.Destroy, "destroy", "Destroy cluster after test")
-	flag.StringVar(&customflag.ServiceFlag.ExternalFlag.CertManagerVersion, "certManagerVersion", "v1.11.0", "cert-manager version that will be deployed on the cluster")
-	flag.StringVar(&customflag.ServiceFlag.ExternalFlag.RancherHelmVersion, "rancherHelmVersion", "v2.8.0", "rancher helm chart version to use to deploy rancher manager")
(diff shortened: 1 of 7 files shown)


pkg/testcase/helm.go
```diff
@@ -44,16 +44,34 @@ func TestDeployCertManager(version string) {
...
+		addRepoCmd = "helm repo add rancher-prime https://charts.optimus.rancher.io/server-charts/latest && " +
+			"helm repo update"
+		installRancherCmd = fmt.Sprintf("kubectl create namespace cattle-system --kubeconfig=%s " +
+			"&& helm install rancher rancher-prime/rancher " +
+			"--set bootstrapPassword=admin " +
+			"--set rancherImage=%s " + 
+			"--set 'extraEnv[0].name=CATTLE_AGENT_IMAGE' " + 
+			"--set 'extraEnv[0].value=%s-agent:%s'",
+			shared.KubeConfigFile, imgReg, imgReg, imageVersion)
+	}
```
rancher-max, line 66: Instead of this, can we just make the helm repo a flag? Since this registry isn't what we want to use all of the time (not every release is a prime-only release). So I'm imagining that line 49 looks like: `helm repo add rancher " + helmRepo` and `helmRepo` is a flag that is defaulted to this `latest` but we can pass in staging, optimus, etc. instead.

rancher-max (changes requested):
One suggestion but everything else looks good. Make sure to `go fmt ./...` as well!

entrypoint/deployrancher/rancher_suite_test.go
```diff
@@ -23,6 +23,7 @@ func TestMain(m *testing.M) {
 	flag.Var(&customflag.ServiceFlag.ClusterConfig.Destroy, "destroy", "Destroy cluster after test")
 	flag.StringVar(&customflag.ServiceFlag.ExternalFlag.CertManagerVersion, "certManagerVersion", "v1.11.0", "cert-manager version that will be deployed on the cluster")
 	flag.StringVar(&customflag.ServiceFlag.ExternalFlag.RancherHelmVersion, "rancherHelmVersion", "v2.8.0", "rancher helm chart version to use to deploy rancher manager")
+	flag.StringVar(&customflag.ServiceFlag.ExternalFlag.RancherHelmArgs, "rancherHelmArgs", "", "rancher helm additional args that can be added on the rancher manager")
```
fmoral2, line 26: can you add on testing.md which args can be used and whats the usage of those ?

fmoral2: and how to use

pkg/testcase/helm.go
```diff
@@ -44,16 +44,34 @@ func TestDeployCertManager(version string) {
...
+	if helmArgs != "" {
+		installRancherCmd += helmArgs
+	}
+	installRancherCmd += " -n cattle-system --set global.cattle.psp.enabled=false " +
 		fmt.Sprintf("--set hostname=%s --set rancherImageTag=%s --version=%s --kubeconfig=%s",
 			cluster.FQDN, imageVersion, helmVersion, shared.KubeConfigFile)
 
+	fmt.Println("Install command: ", installRancherCmd)
```
fmoral2, line 74: if its not a data structure very specific that would only fits in println instead our logs, lets only use shared.LogLevel for logging info or debug

pkg/customflag/config.go
```diff
@@ -147,12 +155,12 @@ func (d *destroyFlag) Set(value string) error {
 	return nil
 }
 
-func (e *externalConfigFlag) String() string {
-	return e.SonobuoyVersion
-}
+// func (e *externalConfigFlag) String() string {
```
fmoral2, line 158: forgot comments

entrypoint/deployrancher/rancher_suite_test.go
```diff
@@ -52,6 +60,16 @@ var _ = BeforeSuite(func() {
...
 			"Wrong value passed in tfvars for 'server_flags'")
 	}
+
+	// Check helm chart repo
+	res, err := shared.CheckHelmRepo(
+		flags.ExternalFlag.HelmChartsFlag.RepoName,
+		flags.ExternalFlag.HelmChartsFlag.RepoUrl,
+		flags.ExternalFlag.HelmChartsFlag.Version)
+	fmt.Println(res)
+	fmt.Println(err)
```
rancher-max, line 70: looks like debug lines that should probably be removed 😇

pkg/testcase/helm.go
```diff
@@ -88,3 +111,29 @@ func TestDeployRancher(helmVersion, imageVersion string) {
...
+		for _, arg := range argsSlice {
+			if !strings.Contains(finalArgs, arg) {
+				finalArgs += fmt.Sprintf("--set %s ", arg)
+			}
+		}
+	} else {
+		finalArgs = fmt.Sprintf("--set %s ", helmArgs)
+	}
+
+	if strings.Contains(flags.ExternalFlag.RancherVersion, "v2.7.12") {
```
rancher-max, line 128: Is this only valid for this one version? Or should we remove this and expect any `extraEnv` stuff to be added into the helmChartArgs themselves when relevant?

mdrahman-suse: I think so as I have not seen those args being used / needed in v2.8+

mdrahman-suse: But if needed, we can remove this check and let user define those args for v2.7.x

rancher-max: yeah let's remove it and document in testing.md! I think it'll be relevant in other versions later on so I'd rather be future-proof here

mdrahman-suse: Updated

rancher-max (commented):
ahh I think we still want the `rancherImageVersion` for the cases when we need to pull, for example, v2.8.3 for the chart, but use `v2.8-head` for the rancher image. But the change you made for the removal of the extra args bit for the one specific version looks good to me.

mdrahman-suse:
> ahh I think we still want the `rancherImageVersion` for the cases when we need to pull, for example, v2.8.3 for the chart, but use `v2.8-head` for the rancher image. But the change you made for the removal of the extra args bit for the one specific version looks good to me.

We can add that as the `rancherVersion` parameter

rancher-max:
> > ahh I think we still want the `rancherImageVersion` for the cases when we need to pull, for example, v2.8.3 for the chart, but use `v2.8-head` for the rancher image. But the change you made for the removal of the extra args bit for the one specific version looks good to me.
> 
> We can add that as the `rancherVersion` parameter. So for charts we have `chartsVersion` and for rancher we have `rancherVersion`

`rancherVersion` is the helm chart version, but not necessarily the image version used in the resulting deployment 😬 It generally is, but not always

rancher-max (approved):
I didn't recognize the variable name changes 🤦‍♂️ This looks great! Ignore my previous comment 😄

mdrahman-suse:
> I didn't recognize the variable name changes 🤦‍♂️ This looks great! Ignore my previous comment 😄

No worries that is my bad, shouldve mentioned on the description 🙇

fmoral2 (approved):

PR #88, opened 2024-04-08
return statement lint back + misspell

fmoral2:
#### Proposed Changes ####

- IDK in which moment we stop having this on lint, but i am adding it back.

#### Testing ####

Checklist:


.golangci.yaml
```diff
@@ -9,6 +9,8 @@ linters:
     - dupl
     - gosimple
     - staticcheck
+    - nlreturn
+    - misspell
```
rancher-max, line 13: Does this not complain about things like `kubectl`, `k3s`, `rke2`, etc.?

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/distros-test-framework
 
-go 1.20
+go 1.21
+
+toolchain go1.21.8
```
rancher-max, line 5: I think this needs a rebase onto main.. this should already be present in go.mod there

pkg/assert/validate.go
```diff
@@ -33,6 +33,7 @@ func validate(exec func(string) (string, error), args ...string) error {
 			if err != nil {
 				shared.LogLevel("error", "error from runAssertion():\n %s\n", err)
 				close(errorsChan)
+
```
rancher-max, line 36: This one shouldn't need a new line since it should fall within the block-size of 4

pkg/assert/validate.go
```diff
@@ -54,6 +55,7 @@ func runAssertion(
 		res, err := exec(cmd)
 		if err != nil {
 			errorsChan <- err
+
```
rancher-max, line 58: Also no new line necessary because block-size < 4 I think

.golangci.yaml
fmoral2: apparently no lol

pkg/assert/validate.go
fmoral2: block size of 4 only counts form the beginning of the func, this is to avoid needing extra line in too small funcs or just returning funcs

.golangci.yaml
rancher-max: Thank goodness!

pkg/assert/validate.go
rancher-max: oohh interesting! I'm fine with this then!I also checked the tests on the repo (https://github.com/ssgreg/nlreturn/blob/master/testdata/src/p/p.go) and it looks like we can still have hings bunched at least for the one-line error returns, like:
```go
// some code

_, err := someFunc()
if err != nil {
  return fmt.Sprintf("There was an error: %v", err)
}

// some more code
```

So that's nice and clear!

rancher-max (commented):
LGTM after rebase!

VestigeJ (approved):

pkg/assert/validate.go
fmoral2: agreed 
yep that was the idea

VestigeJ (approved):

rancher-max (approved):

Issue #93, opened 2024-04-16
issue on creating single cluster with cilium cni 


fmoral2:
if is a single node cluster on cni: cilium, always have a pod that keeps in pending  , which is ok due to only one node.

needs to fix

PR #95, opened 2024-04-17
fix cilium single node

fmoral2:
#### Proposed Changes ####
- fix https://github.com/rancher/distros-test-framework/issues/93

#### Testing ####

Checklist:

### DIFF
--- a/pkg/testcase/pod.go
+++ b/pkg/testcase/pod.go
@@ -13,6 +13,11 @@ import (
 
 const statusCompleted = "Completed"
 
+var (
+	ciliumPodsRunning    = 0
+	ciliumPodsNotRunning = 0
+)
+
 // TestPodStatus test the status of the pods in the cluster using custom assert functions
 func TestPodStatus(
 	podAssertRestarts assert.PodAssertFunc,
@@ -39,6 +44,7 @@ func processPodStatus(
 	podAssertRestarts, podAssertReady, podAssertStatus assert.PodAssertFunc,
 ) {
 	cluster := factory.ClusterConfig(GinkgoT())
+	var ciliumPod bool
 
 	// process Helm install status that should be completed.
 	if strings.Contains(pod.Name, "helm-install") {
@@ -56,6 +62,7 @@ func processPodStatus(
 	} else if strings.Contains(pod.Name, "cilium-operator") && cluster.Config.Product == "rke2" &&
 		cluster.NumServers == 1 && cluster.NumAgents == 0 {
 		processCiliumStatus(pod)
+		ciliumPod = true
 
 		// process other pods status that should be running.
 	} else {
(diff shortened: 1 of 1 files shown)


pkg/testcase/pod.go
```diff
@@ -36,14 +38,25 @@ func processPodStatus(
...
+		// process system-upgrade apply status thats should be completed or errors bellow.
 	} else if strings.Contains(pod.Name, "apply") && strings.Contains(pod.NameSpace, "system-upgrade") {
 		g.Expect(pod.Status).Should(SatisfyAny(
 			ContainSubstring("Unknown"),
 			ContainSubstring("Init:Error"),
 			Equal(statusCompleted),
 		), pod.Name)
+
+		// process cilium-operator status that should be at least one running pod.
+	} else if strings.Contains(pod.Name, "cilium-operator") && cluster.Config.Product == "rke2" && cluster.NumServers == 1 {
```
rancher-max, line 56: This should probably be the aggregate of the number of servers and workers I think, since it'll have both pods running in the case of 1 server 1 worker or just 2 servers

mdrahman-suse: So this check wont be triggered if the server_count is more than 1 even with 0 workers?

mdrahman-suse: Also I think it will be simpler to check `cni: cilium` to  be present in `server_flags` tf var, just a suggestion

fmoral2: > So this check wont be triggered if the server_count is more than 1 even with 0 workers?


this check its going to be triggered if:

- product  rke2 + has cilium operator pod + it has only one server +  ( 0 agents,  what i mentioned above )

fmoral2: > Also I think it will be simpler to check cni: cilium to be present in server_flags tf var, just a suggestion

but i will need to also check for servers and agents, so since we are here validating pods and we already have a pod structure available, i choose  to be more specific.

mdrahman-suse: Got it, I wanted to understand if we should always do this check regardless of the server/agent count and be consistent to check the CNI specific pods across other CNI's as well

fmoral2: no need, only when 1 server alone

pkg/testcase/pod.go
```diff
@@ -57,3 +71,11 @@ func processPodStatus(
 		}
 	}
 }
+
+func processCiliumStatus(pod shared.Pod) {
```
rancher-max, line 75: I'm stuck here because it will still return a PASS based on this code if BOTH pods are in Pending, whereas it should only return PASS if ONE pod is in Pending and the other one is Running. Can we update this to somehow reflect that instead?

fmoral2: good catch, i was not counting on that

pkg/testcase/pod.go
```diff
@@ -56,4 +72,14 @@ func processPodStatus(
 			podAssertStatus(g, pod)
 		}
 	}
+	Expect(ciliumPodsRunning).Should(BeNumerically(">=", 1), "no cilium pods running")
```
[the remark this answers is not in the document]
mdrahman-suse: Should add it on the else if block where we are checking for cilium-operator

pkg/testcase/pod.go
```diff
@@ -56,4 +77,27 @@ func processPodStatus(
 			podAssertStatus(g, pod)
 		}
 	}
+
+	// at least one cilium pod should be running.
+	if ciliumPod {
```
mdrahman-suse, line 82: Wondering if we can move this code block inside `processCiliumStatus` to scope it within that func, that way we would not require the global vars on this file (which does not have any usage except for cilium based tests)

fmoral2: and on thiws func, we need a way to only run this when it has cilium pods

mdrahman-suse (approved):
LGTM

rancher-max (approved):

PR #97, opened 2024-04-23
improve install fail

fmoral2:
#### Proposed Changes ####
- https://github.com/rancher/qa-tasks/issues/894 

#### Testing ####

Checklist:

### DIFF
--- a/factory/cluster.go
+++ b/factory/cluster.go
@@ -2,6 +2,7 @@ package factory
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"strconv"
 
@@ -18,8 +19,17 @@ func ClusterConfig(g GinkgoTInterface) *Cluster {
 		var err error
 		cluster, err = newCluster(g)
 		if err != nil {
-			err = shared.ReturnLogError("error getting cluster: %w\n", err)
-			g.Errorf("%s", err)
+			status, destroyErr := DestroyCluster(g)
+			if destroyErr != nil {
+				shared.LogLevel("error", "error destroying cluster: %w\n", destroyErr)
+				return
+			}
+			if status != "cluster destroyed" {
+				shared.LogLevel("error", "cluster not destroyed: %s\n", status)
+				os.Exit(1)
+			}
+			shared.LogLevel("error", "building cluster failed!: %w\nmoving to start destroy operation\n", err)
+			os.Exit(1)
 		}
 	})
 
@@ -56,7 +66,7 @@ func newCluster(g GinkgoTInterface) (*Cluster, error) {
 	shared.LogLevel("info", "\nCreating cluster\n")
 	_, err = terraform.InitAndApplyE(g, terraformOptions)
 	if err != nil {
-		shared.LogLevel("error", "\nCreating cluster Failed!!!\n")
+		shared.LogLevel("error", "\nTerraform apply Failed: %w", err)
 		return nil, err
 	}
 
--- a/factory/cluster.go
+++ b/factory/cluster.go
@@ -19,6 +19,7 @@ func ClusterConfig(g GinkgoTInterface) *Cluster {
 		var err error
 		cluster, err = newCluster(g)
 		if err != nil {
+			shared.LogLevel("error", "building cluster failed!: %w\nmoving to start destroy operation\n", err)
 			status, destroyErr := DestroyCluster(g)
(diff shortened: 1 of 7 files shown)


modules/install/join_rke2_master.sh
```diff
@@ -97,25 +96,55 @@ cis_setup() {
   fi
 }
 
-install() {
+
+export_variables() {
   export "$install_mode"="$version"
-  if [ -n "$install_method" ]; then
-    export INSTALL_RKE2_METHOD="$install_method"
+   if [ -n "$install_method" ]; then
```
rancher-max, line 102: remove extra space in front of the `if`

factory/cluster.go
```diff
@@ -18,8 +19,17 @@ func ClusterConfig(g GinkgoTInterface) *Cluster {
...
+			status, destroyErr := DestroyCluster(g)
+			if destroyErr != nil {
+				shared.LogLevel("error", "error destroying cluster: %w\n", destroyErr)
+				return
+			}
+			if status != "cluster destroyed" {
+				shared.LogLevel("error", "cluster not destroyed: %s\n", status)
+				os.Exit(1)
+			}
+			shared.LogLevel("error", "building cluster failed!: %w\nmoving to start destroy operation\n", err)
```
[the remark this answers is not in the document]
fmoral2: actually makes no difference here  but i do think its clearer moving this up, tks

modules/install/join_rke2_agent.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/bash
 # This script is used to join one or more nodes as agents
 echo "$@"
-# the following lines are to enable debug mode
-set -x
```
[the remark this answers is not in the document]
fmoral2: my pov is that is no longer needed since we are now moving to a good shape of "info" available, since we are already printing/echoing usefull stuff and also errors . but let me know

modules/install/k3s_master.sh
```diff
@@ -93,22 +91,43 @@ disable_cloud_setup() {
...
 
-install() {
+export_variables() {
   export "$install_mode"="$version"
+   alias k=kubectl
+}
+
+install_k3s() {
+  url="https://get.k3s.io"
+  params="INSTALL_K3S_TYPE='server'"
```
ShylajaDevadiga, line 101: We don't need this as we pass "server" in install_command in LN 108/110

modules/install/k3s_master.sh
```diff
@@ -93,22 +91,43 @@ disable_cloud_setup() {
...
+         sudo journalctl -xeu k3s.service | grep -i "failed\|fatal"
+         exit 1
     fi
-  fi
+}
+
+install() {
+  export_variables
+  install_k3s
+  sleep 20
```
ShylajaDevadiga, line 129: We need to handle this differently if wait is needed. Adds to install time which may not be necessary

ShylajaDevadiga (commented):
Nit: Lets run through indentation on all scripts.

modules/install/join_rke2_agent.sh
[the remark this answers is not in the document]
rancher-max: yeah I'm totally fine without it, just wanted to make sure it was intentional!

modules/install/join_k3s_agent.sh
```diff
@@ -82,16 +80,33 @@ disable_cloud_setup() {
...
     curl -sfL https://get.k3s.io | sh -s - agent
   fi
-  sleep 15
+}
+
+check_service() {
+  if systemctl is-active --quiet k3s-agent; then
+      printf "K3s is running on agent node ip: %s\n" "$public_ip"
+  else
+      printf "K3s failed to start on agent node ip %s\n" "$public_ip"
```
mdrahman-suse, line 99: Can we keep this message consistent? In some cases we are using `joining` and in some we are not.

mdrahman-suse (changes requested):
One suggested change on adding consistent message across failures, else LGTM

modules/install/k3s_master.sh
fmoral2: i decrease to 10 s, but this is just a time to wait the install complete. nothing more to do here

rancher-max (approved):

ShylajaDevadiga (approved):

PR #102, opened 2024-05-13
Add support for kine in rke2

mdrahman-suse:
#### Proposed Changes ####

Adding support for the new kine feature implemented in rke2

- No, Need to update JJB to add the new external datastore variables

#### Types of Changes ####

- New functionality to add external datastore in rke2

#### Testing ####

Checklist:
- Ok

- Ok

- Ok

- N/A

- Ok

- TBA, also run with different CNI's and DB types


rancher-max (approved):

pkg/testcase/cluster.go
```diff
@@ -25,16 +25,17 @@ func TestBuildCluster(g GinkgoTInterface) {
...
 		res, err := shared.RunCommandOnNode(cmd, cluster.ServerIPs[0])
 		Expect(err).NotTo(HaveOccurred())
 		Expect(res).Should(ContainSubstring(cluster.Config.RenderedTemplate))
 
-		etcd, err := shared.RunCommandHost("cat /var/lib/rancher/k3s/server/db/etcd/config",
-			cluster.ServerIPs[0])
-		// TODO: validate also after fix https://github.com/k3s-io/k3s/issues/8744
-		Expect(etcd).Should(ContainSubstring(" No such file or directory"))
-		Expect(err).To(HaveOccurred())
+		// TODO: This fails for rke2
```
fmoral2, line 33: have u forgot this ?

factory/config.go
```diff
@@ -129,14 +129,12 @@ func loadTFconfig(
...
 
-	if cfg.Product == "k3s" {
-		c.Config.DataStore = terraform.GetVariableAsStringFromVarFile(g, varDir, "datastore_type")
-		if c.Config.DataStore == "external" {
-			c.Config.ExternalDb = terraform.GetVariableAsStringFromVarFile(g, varDir, "external_db")
-			c.Config.RenderedTemplate = terraform.Output(g, terraformOptions, "rendered_template")
-		} else if c.Config.DataStore == "" {
-			return nil, shared.ReturnLogError("datastore should not be empty \n%w", err)
-		}
+	c.Config.DataStore = terraform.GetVariableAsStringFromVarFile(g, varDir, "datastore_type")
```
fmoral2, line 132: so its all the same config for k3s external dbs? on vars and in the terraform ?

mdrahman-suse: Yes it is the same as k3s

pkg/testcase/cluster.go
mdrahman-suse: Nope it actually fails for rke2 as the etcd folder is created in that path regardless, should raise a ticket and add it here

fmoral2: and this 
```
// // TODO: validate also after fix https://github.com/k3s-io/k3s/issues/8744
```

you probably can remove as well

rancher-max (dismissed):

factory/cluster.go
```diff
@@ -18,7 +19,7 @@ func ClusterConfig(g GinkgoTInterface) *Cluster {
 	once.Do(func() {
 		var err error
 		cluster, err = newCluster(g)
-		if err != nil {
+		if err != nil && customflag.ServiceFlag.ClusterConfig.Destroy {
```
fmoral2, line 22: here you are saying &&  but we want to destroy here even if `customflag.ServiceFlag.ClusterConfig.Destroy` is empty

fmoral2: this flag is already being used on suites - https://github.com/rancher/distros-test-framework/blob/6b77ea38bbd8b277012736a360c6a03069880595/entrypoint/certrotate/certrotate_suite_test.go#L39

mdrahman-suse: So I would like to have the option to keep the cluster in case I want to debug the failure. Should we add a separate debug flag for that then? Else right now its destroying on failures by default and I have no way to debug why it failed

fmoral2: so the  "install/ initialization " failures  will be logged pretty much every time for you 

do we want on to still have the cluster up to debug further? 

if so, yeah we should leave as you did with `&&` 

when i add that i was thinking that we wanted always when errors come from the install/ init tf and etc stuff, we wanted to only log and kill

fmoral2: if we go that way you proposed , i would say that does not even need to add the kill cluster in here
since we are already passing kill=true , its a overhead and duplicate stuff

fmoral2: if thats the case what i am saying is then go like 
```

func ClusterConfig() *Cluster {
	once.Do(func() {
		var err error
		cluster, err = newCluster()
		if err != nil {
			l.Errorf("building cluster failed!: %vn", err)
			os.Exit(1)
		}
	})

	return cluster
}

Comment (mdrahman-suse) [reply factory/cluster.go:22]: Yeah so there are couple of types of failure we can get, these comes to my mind:

1.  Failure sure to terraform variable/value mismatch (like unsupported or missing value on tf.vars)
2. Failure on the installation process
3. Failure on the test itself

So, for 1. we can destroy the cluster as it does not make sense to proceed as the var/value is wrong
But for 2 and 3, we should have the ability to debug further to find where and why it failed even though we are logging it. Will need the cluster to share the logs as well to the devs for debugging purpose. Hence I was proposing a `debug` flag that can assist with keeping or destroying the cluster... lmk what you think?

Comment (fmoral2) [reply factory/cluster.go]: i thought i had answered before , sorry 

the debug option would be the optimal here, but for now i would say lets only destroy if we pass the true flag in?

Comment (fmoral2) [reply factory/cluster.go]: if we found error we exit always, but destroy only if flag ==true

File: scripts/test_runner.sh [hunk 4]
```diff
@@ -82,21 +82,21 @@ if [ -n "${TEST_DIR}" ]; then
...
+        go test -timeout=65m -v -count=1 ./entrypoint/dualstack/... -destroy "${DESTROY}"
     elif [  "${TEST_DIR}" = "createcluster" ]; then
-        go test -timeout=60m -v -count=1 ./entrypoint/createcluster/...
+        go test -timeout=60m -v -count=1 ./entrypoint/createcluster/... -destroy "${DESTROY}"
     elif [ "${TEST_DIR}" = "validatecluster" ]; then
-        go test -timeout=65m -v -count=1 ./entrypoint/validatecluster/...
+        go test -timeout=65m -v -count=1 ./entrypoint/validatecluster/... -destroy "${DESTROY}"
     elif [ "${TEST_DIR}" = "selinux" ]; then
-        go test -timeout=65m -v -count=1 ./entrypoint/selinux/...
+        go test -timeout=65m -v -count=1 ./entrypoint/selinux/... -destroy "${DESTROY}"
```
fmoral2, line 91: can we add the flag on the suite as well? 

i see selinux and restart service without

scripts/test_runner.sh
```diff
@@ -82,21 +82,21 @@ if [ -n "${TEST_DIR}" ]; then
             [ -n "${RANCHERVERSION}" ] && OPTS+=(-rancherVersion "${RANCHERVERSION}")
       go test "${OPTS[@]}"
     elif [ "${TEST_DIR}" = "dualstack" ]; then
-        go test -timeout=65m -v -count=1 ./entrypoint/dualstack/...
+        go test -timeout=65m -v -count=1 ./entrypoint/dualstack/... -destroy "${DESTROY}"
```
fmoral2, line 85: if you are adding ` -destroy "${DESTROY}"` here on test runner, you need to also refactor makefile + docker run file which are the consumers of it 

right now you will get an error trying to start tests locally cause is expecting this flag

mdrahman-suse: I though we need to add it here for the batch test to work but realize now that its set to true in the suite itself, so I guess I can remove it then from here

fmoral2: jenkins dont use this file

pkg/assert/node.go
```diff
@@ -21,7 +21,7 @@ func NodeAssertVersionTypeUpgrade(c customflag.FlagConfig) NodeAssertFunc {
 		return assertCommit(c)
 	}
 
-	return func(g Gomega, node shared.Node) {
+	return func(_ Gomega, _ shared.Node) {
```
fmoral2, line 24: why this change?

mdrahman-suse: As suggested by linter
```
```

fmoral2: ok , so lets remove `	return func(Gomega, shared.Node) {` , not ommit _ 

_  is supposed to be used when u need a parameter on the signature but its not going to use it inside.

is slightly different from what linter said

pkg/template/helper.go
```diff
@@ -69,45 +69,45 @@ func AddTestCases(cluster *factory.Cluster, names []string) ([]testCase, error)
 		"TestInternodeConnectivityMixedOS": func(applyWorkload, deleteWorkload bool) {
 			testcase.TestInternodeConnectivityMixedOS(cluster, applyWorkload, deleteWorkload)
 		},
-		"TestSonobuoyMixedOS": func(applyWorkload, deleteWorkload bool) {
+		"TestSonobuoyMixedOS": func(_, deleteWorkload bool) {
```
fmoral2, line 72: on here honestly i would not remove that, cause right now its showing us the  "conversion" to the expected signature

fmoral2: the expected signature of the template (not the test itself) thats why the conversion

mdrahman-suse: Updated

pkg/assert/node.go
mdrahman-suse: Ok added it back and not seeing ci-linter complaining 🤷

VestigeJ (approved):

pkg/template/helper.go
```diff
@@ -107,7 +107,7 @@ func AddTestCases(cluster *factory.Cluster, names []string) ([]testCase, error)
 	for _, name := range names {
 		name = strings.TrimSpace(name)
 		if name == "" {
-			testCases = append(testCases, func(applyWorkload, deleteWorkload bool) {})
+			testCases = append(testCases, func(_, _ bool) {})
```
fmoral2, line 110: lets get this back also please

mdrahman-suse: Updated

rancher-max (approved):

PR #110, opened 2024-06-06
fix arg for suc upgrade

aganesh-suse:
#### Proposed Changes ####

1. missed a typo in the if condition for suc upgrade test. 
2. hard code the dualstack alone to be jenkins-elliptic-validation since it is in a different region and needs different pem value
3. add rke2_windows_agent job in the batch jobs triggered. 

#### Testing ####

Checklist:

### DIFF
--- a/scripts/Jenkinsfile_batch_test
+++ b/scripts/Jenkinsfile_batch_test
@@ -48,6 +48,8 @@ node {
             test_dir.each { test_name ->
               def actual_testdir_name = test_name
               def job_test_args = "${TEST_ARGS}"
+              def job_aws_ssh_key_name = "${AWS_SSH_KEY_NAME}"
+              def job_aws_ssh_pem_key = "${AWS_SSH_PEM_KEY}"
               if ( test_name == "upgrademanual" || test_name == "upgradereplacement" || test_name == "upgradesuc"){
                 actual_testdir_name = "upgradecluster"
                 if ( test_name == "upgradesuc") {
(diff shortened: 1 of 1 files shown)


mdrahman-suse:
I see that the batch test is missing rke2_windows job, can we add that?

rancher-max (dismissed):

scripts/Jenkinsfile_batch_test
```diff
@@ -50,7 +50,7 @@ node {
               def job_test_args = "${TEST_ARGS}"
               if ( test_name == "upgrademanual" || test_name == "upgradereplacement" || test_name == "upgradesuc"){
                 actual_testdir_name = "upgradecluster"
-                if ( test_name == "sucupgrade") {
+                if ( test_name == "upgradesuc") {
                   job_test_args = job_test_args + " -tags=${test_name} -sucUpgradeVersion ${VERSION_OR_COMMIT}"
```
fmoral2, line 54: shouldnt we do something like this here? 
```
              if ( test_name == "upgrademanual" || test_name == "upgradereplacement" || test_name == "upgradesuc"){
                actual_testdir_name = "upgradecluster"

                if ( test_name == "upgradesuc") {
                  job_test_args = job_test_args + " -tags=${test_name} -sucUpgradeVersion ${VERSION_OR_COMMIT}"
                } else {
                  job_test_args = job_test_args + " -tags=${test_name} -installVersionOrCommit ${VERSION_OR_COMMIT}"
                }

                  job_test_args = job_test_args +  " -channel ${UPGRADE_CHANNEL} "
              }
```

Since in all we use channel and avoid code duplication

fmoral2: since the  upgrade flag is the same for manual and replacement

aganesh-suse: fixed it. the reason i did it though is right now only manual upgrade jenkins job was using it. the rest did not use a channel flag in the test args. as long as the go test takes the channel flag, there should not be a problem.

mdrahman-suse (approved):
LGTM

VestigeJ (approved):
LGTM

PR #114, opened 2024-06-14
Adding test for flannel version bump

mdrahman-suse:
#### Proposed Changes ####

Added separate test for flannel cni

#### Types of Changes ####

Added new test for flannel cni

#### Testing ####

Checklist:
No


fmoral2:
we need to remove then now, from the components file

mdrahman-suse:
> we need to remove then now, from the components file and add this on the version bump jenkins job

But we still want to keep flannel in component bump for canal validation, isn't it? As for canal, we validate flannel and calico

fmoral2:
> > we need to remove then now, from the components file and add this on the version bump jenkins job
> 
> But we still want to keep flannel in component bump for canal validation, isn't it? As for canal, we validate flannel and calico

true, maybe just remove for k3s and for rke2 rename it ?

entrypoint/versionbump/componentsbump_test.go
```diff
@@ -14,20 +14,16 @@ import (
...
-	flannelK3s        = "/var/lib/rancher/k3s/data/current/bin/flannel, "
-	calico            = "kubectl get node -o yaml : | grep 'hardened-calico' -A1, "
-	ingressController = "kubectl get node -o yaml : | grep 'nginx-ingress-controller' -A1, "
-	corednsRke2       = "kubectl get node -o yaml : | grep 'hardened-coredns' -A1, "
-	coreDnsk3s        = "kubectl get node -o yaml : | grep 'mirrored-coredns' -A1, "
-	metricsServer     = "kubectl get node -o yaml : | grep 'metrics-server' -A1, "
-	etcdRke2          = "kubectl get node -o yaml : | grep 'hardened-etcd' -A1, "
-	containerd        = "kubectl get node -o yaml : | grep  containerd, "
+	kgn               = "kubectl get node -o yaml"
+	flannel           = kgn + " : | grep 'hardened-flannel' -A1, "
```
fmoral2, line 18: maybe canal-flannel ? the name

pkg/template/helper.go
```diff
@@ -123,6 +123,7 @@ func currentProductVersion() (string, error) {
 	if err != nil {
 		return "", shared.ReturnLogError("failed to get product: %w", err)
 	}
+	fmt.Println(version)
```
fmoral2, line 126: why this print?

mdrahman-suse: To display the version on the logs, helps adding it to issue validation

fmoral2: why have that already under debug flag here 
https://github.com/rancher/distros-test-framework/blob/b098805b7b4c04d8d638593adc23eb8e71e1d04a/pkg/template/processor.go#L86

 we also have it here - https://github.com/rancher/distros-test-framework/blob/b098805b7b4c04d8d638593adc23eb8e71e1d04a/entrypoint/versionbump/versionbump_suite_test.go#L81


but if you want add it yet , lets try to add as log not printing , since doesnt need any especial formatting , lets try to keep consistency using logs to return info, not print

mdrahman-suse: Go it, I was not seeing it getting printed after this command was executed on the logs 
```
18:16:56  PASSED! Flannel Version bump: Validate Pod
18:16:56  [38;5;10m•[0m[37mDEBU[0m[2024-06-14T23:16:42Z] Execute: rke2 -v on <ip-addr>           
18:16:56  
18:16:56  Command:
```
so I thought may be it was not implemented. Will update

VestigeJ (dismissed):

entrypoint/versionbump/flannelbump_test.go
```diff
@@ -0,0 +1,76 @@
...
+			cluster,
+			assert.PodAssertRestart(),
+			assert.PodAssertReady(),
+			assert.PodAssertStatus())
+	})
+
+	cmd := "kubectl get node -o yaml : | grep 'hardened-flannel' -A1"
+	if cluster.Config.Product == "k3s" {
+		cmd = "/var/lib/rancher/k3s/data/current/bin/flannel"
+	}
```
rancher-max, line 39: Can we put this in the `It` block below where it's used so that it is scoped only to where it is used?

pkg/template/helper.go
fmoral2: sorry got confused lol

do you still need to print? if so , lets use log.level()

mdrahman-suse: Yes I think we should print, updated

entrypoint/versionbump/flannelbump_test.go
mdrahman-suse: Updated

fmoral2 (approved):

VestigeJ (approved):

entrypoint/versionbump/componentsbump_test.go
```diff
@@ -14,20 +14,16 @@ import (
...
-	flannelK3s        = "/var/lib/rancher/k3s/data/current/bin/flannel, "
-	calico            = "kubectl get node -o yaml : | grep 'hardened-calico' -A1, "
-	ingressController = "kubectl get node -o yaml : | grep 'nginx-ingress-controller' -A1, "
-	corednsRke2       = "kubectl get node -o yaml : | grep 'hardened-coredns' -A1, "
-	coreDnsk3s        = "kubectl get node -o yaml : | grep 'mirrored-coredns' -A1, "
-	metricsServer     = "kubectl get node -o yaml : | grep 'metrics-server' -A1, "
-	etcdRke2          = "kubectl get node -o yaml : | grep 'hardened-etcd' -A1, "
-	containerd        = "kubectl get node -o yaml : | grep  containerd, "
+	kgn               = "kubectl get node -o yaml"
+	canalFlannel      = kgn + " : | grep 'hardened-flannel' -A1, "
```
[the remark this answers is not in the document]
mdrahman-suse: We can but not sure if we are using them anywhere else... also this allows scoping these vars within the test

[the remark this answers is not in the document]
fmoral2: we can use the `kgn`  like this in other tests like cilium:
`						Cmd:                  kgn + " : | grep mirrored-cilium  -A1,kubectl get node -o yaml : | grep hardened-cni-plugins -A1",`

so if you want to use like this we dont need even to change this anywhere else, because it still visible for all other tests on this pkg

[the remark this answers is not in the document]
fmoral2: but tbh not seeing any huge gain on that , since those commands are not stuff we are going to change regularly and so on

ShylajaDevadiga (approved):

entrypoint/versionbump/componentsbump_test.go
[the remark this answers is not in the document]
rancher-max: I don't have a _super_ strong opinion here, but I do kind of like the explicit `kubectl get node` since it's easy for someone who is less familiar with the commands we use and the processes to understand what is being run exactly if they look at the code.

PR #125, opened 2024-07-11
Fix for split-role SUC and edge case around binary location

rancher-max:
#### Proposed Changes ####

Two fixes:
1. split-role SUC does not have a `master` role. Both split-role and standard have the `control-plane` role, so this should be used instead.
2. There is an edge case where sometimes the data-dir binary file will be returned first instead of the /usr/... file. This causes failures during at least cluster-reset, so it should be ignored when retrieving the file.

#### Types of Changes ####

Fix

#### Testing ####

Checklist:
N/A

✅ 

N/A

N/A

Ran cluster-reset, create, validate, and suc upgrade.

#### Linked Issues ####

N/A


fmoral2:
have u ran in multiple OS.s ?

fmoral2 (approved):

rancher-max:
> have u ran in multiple OS.s ?

I ran with RHEL and SLES only

ShylajaDevadiga (approved):

PR #142, opened 2024-08-09
Reboot.improve

fmoral2:
#### Proposed Changes ####

-  created a auxiliar method waitInstanceStop to validate the actual stopping and not needing to rely on the output from StopInstances(input) because it can be inconsistent and time wise  and + not needing of eventually function
 
- improve performance with go routines
- added also "reboot on worker nodes"
- added a method for releases Elastic IPs after tests
- added some empty/nil validations
- added validation for the var itself before start terraform 

#### Testing ####

Checklist:


pkg/aws/aws.go
```diff
@@ -146,39 +146,138 @@ func (c Client) DeleteInstance(ip string) error {
...
+		}
+	}
 
-				return nil
+	return "", shared.ReturnLogError("no instances found with the public IP address: ")
+}
+
+func (c Client) StopInstance(instanceID string) error {
+	if instanceID == "" {
+		return shared.ReturnLogError("must sent a instance ID")
```
ShylajaDevadiga, line 185: nit: Can we rephrase this to be more clear? (and other locations)

pkg/aws/aws.go
```diff
@@ -146,39 +146,138 @@ func (c Client) DeleteInstance(ip string) error {
...
+	if startErr != nil {
+		return shared.ReturnLogError("timed out on start instance %s: %v", instanceID, startErr)
+	}
+
+	return nil
+}
+
+func (c Client) ReleaseElasticIps(ipAddress string) error {
+	if ipAddress == "" {
+		return shared.ReturnLogError("must sent an ip address")
```
ShylajaDevadiga, line 229: same

ShylajaDevadiga (dismissed):
small change otherwise LGTM

mdrahman-suse (approved):
LGTM

PR #148, opened 2024-08-20
Fix.lb

fmoral2:
#### Testing ####

Checklist:


pkg/assert/validate.go
```diff
@@ -60,27 +54,29 @@ func runAssertion(
...
 			if err != nil {
-				i++
-				shared.LogLevel("warn", "error from exec runAssertion: %v\n with res: %s\nRetrying...", err, res)
-				if i > 5 {
-					errorsChan <- shared.ReturnLogError("error from exec runAssertion: %v\nwith res: %s\n", err, res)
-					return shared.ReturnLogError("error from exec runAssertion: %v\nwith res: %s\n", err, res)
+				shared.LogLevel("warn", "error from exec runAssertion: %v", err)
+				shared.LogLevel("warn", "Retrying...executing command: %s\n on retry count: %d\n", cmd, retry)
+				if retry > 5 {
+					return shared.ReturnLogError("error from exec runAssertion after 10 retries: %v\n", err)
```
mdrahman-suse, line 79: Is it 10 or 5 retries?

fmoral2: 5 sorry, was testing

updated

pkg/testcase/service.go
```diff
@@ -81,14 +81,14 @@ func TestServiceLoadBalancer(applyWorkload, deleteWorkload bool) {
 	Expect(err).NotTo(HaveOccurred(), err)
 
 	getAppLoadBalancer := "kubectl get pods -n test-loadbalancer  " +
-		"--field-selector=status.phase=Running --kubeconfig="
+		"--field-selector=status.phase=Running "
```
aganesh-suse, line 84: on some setups kubectl may not work without providing the kubeconfig. double checking removing this is essential here.

pkg/assert/validate.go
```diff
@@ -60,27 +54,29 @@ func runAssertion(
...
 		select {
 		case <-timeout:
 			timeoutErr := shared.ReturnLogError("timeout reached for command:\n%s\n"+
 				"Trying to assert with:\n%s\nExpected value: %s\n", cmd, res, assert)
-			errorsChan <- timeoutErr
 
 			return timeoutErr
 		case <-ticker:
-			i := 0
+			retry++
```
aganesh-suse, line 72: assuming that retry need not be initialized here to zero - that ticker somehow knows the var takes care of it?

aganesh-suse (approved):
LGTM

pkg/testcase/service.go
fmoral2: this is only used here, and its used in  func on NODE, which does not need

pkg/assert/validate.go
fmoral2: the ticker actually has nothing to do with it....

we are just increment this var in case we get back on this select ( which in this case is a ticker )

fmoral2: maybe i should explain better

one thing is the  ticker period of time 
another is the retry inside the err != nil 

thats why would make no sense have a big retry, since the timeout would finish first 

if we are still trying inside the ticker times it run , and still getting error, then we use the incremented var to break the loop some time ( after 5 retries ) 

if no error on next iteration we just increment the var , do nothing with it and exit the select/loop 

if no error never, still creating a var and again not using it and exiting the select/loop

mdrahman-suse (approved):
LGTM

PR #149, opened 2024-08-20
replace hardcoded rke2 to product var instead

aganesh-suse:
#### Proposed Changes ####

Upgrade Node replacement test fails today while looking up FetchToken since it has rke2 hardcoded. 
Updating FetchToken to have product arg and updated the call to the method as well. 

No

#### Testing ####

Checklist:

### DIFF
--- a/pkg/testcase/upgradenodereplacement.go
+++ b/pkg/testcase/upgradenodereplacement.go
@@ -45,7 +45,7 @@ func TestUpgradeReplaceNode(cluster *shared.Cluster, version string) {
 	Expect(scpErr).NotTo(HaveOccurred(), scpErr)
 	shared.LogLevel("info", "Scp files to new server nodes done\n")
 
-	serverLeaderIP := cluster.ServerIPs[0]
+	serverLeaderIP := newExternalServerIps[0]
 	token, err := shared.FetchToken(cluster.Config.Product, serverLeaderIP)
 	Expect(err).NotTo(HaveOccurred(), err)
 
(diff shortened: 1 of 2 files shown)


fmoral2 (dismissed):

pkg/testcase/upgradenodereplacement.go
```diff
@@ -46,7 +46,7 @@ func TestUpgradeReplaceNode(cluster *shared.Cluster, version string) {
 	shared.LogLevel("info", "Scp files to new server nodes done\n")
 
 	serverLeaderIP := cluster.ServerIPs[0]
```
mdrahman-suse, line 48: qq… we would also need to assign the new external ips to the cluster.ServerIPs, isnt't it? else it would show the previous ips...

pkg/testcase/upgradenodereplacement.go
```diff
@@ -45,7 +45,7 @@ func TestUpgradeReplaceNode(cluster *shared.Cluster, version string) {
 	Expect(scpErr).NotTo(HaveOccurred(), scpErr)
 	shared.LogLevel("info", "Scp files to new server nodes done\n")
 
-	serverLeaderIP := cluster.ServerIPs[0]
+	serverLeaderIP := newExternalServerIps[0]
```
mdrahman-suse, line 48: @aganesh-suse sorry please revert this change, will revisit this

fmoral2: yes please

aganesh-suse: i suspect that it is talking to the old IP to get the token to copy on to the new set of nodes maybe. so i dont think we should be using the new IP for this line. reverting the change.

fmoral2: https://github.com/rancher/distros-test-framework/pull/149#discussion_r1723988444

fmoral2 (approved):

mdrahman-suse (approved):
LGTM

PR #154, opened 2024-08-30
Fix daemonset test for upgrades

mdrahman-suse:
#### Proposed Changes ####

Daemoset test fails after upgrade due to miscalculation of indexing

No documentation update needed

#### Types of Changes ####

Refactor, make robust query to fetch data from pod object

#### Testing ####

Checklist:

### DIFF
--- a/pkg/testcase/daemonset.go
+++ b/pkg/testcase/daemonset.go
@@ -19,12 +19,10 @@ func TestDaemonset(applyWorkload, deleteWorkload bool) {
 	}
 	pods, _ := shared.GetPods(false)
 
-	cmd := fmt.Sprintf(`
-		kubectl get pods -n test-daemonset -o wide --kubeconfig="%s" \
-		| grep -A10 NODE | awk 'NR>1 {print $7}'
-		`,
-		shared.KubeConfigFile,
-	)
+	cmd := "kubectl get pods -n test-daemonset" +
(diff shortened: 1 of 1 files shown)


VestigeJ (dismissed):

ShylajaDevadiga (approved):

pkg/testcase/daemonset.go
```diff
@@ -19,12 +19,10 @@ func TestDaemonset(applyWorkload, deleteWorkload bool) {
...
 	pods, _ := shared.GetPods(false)
 
-	cmd := fmt.Sprintf(`
-		kubectl get pods -n test-daemonset -o wide --kubeconfig="%s" \
-		| grep -A10 NODE | awk 'NR>1 {print $7}'
-		`,
-		shared.KubeConfigFile,
-	)
+	cmd := "kubectl get pods -n test-daemonset" +
+		` -o jsonpath='{range .items[*]}{.spec.nodeName}{"\n"}{end}'` +
```
endawkins, line 23: What is the new command doing in contrast to the previous one?

endawkins (approved):

pkg/testcase/daemonset.go
mdrahman-suse: So previously it was taking the node name using grep and parsing the output, this change uses pod objects to get the same data it expects…

ShylajaDevadiga (approved):

PR #163, opened 2024-09-13
Updates kubeconfig

fmoral2:
#### Proposed Changes ####
- minor refactor
- adding kubeconfig on reboot nodes 

#### Testing ####

Checklist:


docs/testing.md
```diff
@@ -117,8 +117,22 @@ chartsArgs rancherImage=<image or url>,extraEnv[0].name=CATTLE_AGENT_IMAGE,extra
...
+
+
+ ### Testing Reboot instances with EIP ###
+
+- Required vars for in `*.tfvars` file:
+```
+ create_eip =   true
+ ```
+- Optional vars for in `.env` file: Will be to not release EIPs after the test, so you can reuse the kubeconfig file.
+Please note that using this option you will need to manually release the EIPs.
```
fmoral2, line 135: using to true

docs/testing.md
```diff
@@ -117,8 +117,22 @@ chartsArgs rancherImage=<image or url>,extraEnv[0].name=CATTLE_AGENT_IMAGE,extra
...
+ ### Testing Reboot instances with EIP ###
+
+- Required vars for in `*.tfvars` file:
+```
+ create_eip =   true
+ ```
+- Optional vars for in `.env` file: Will be to not release EIPs after the test, so you can reuse the kubeconfig file.
+Please note that using this option set to `false` you will need to manually release the EIPs.
+ ```
+ RELEASE_EIP=false
```
aganesh-suse, line 137: is it possible to add this as a flag instead of in the .env file? and set the default value to true. if we will do this as a later change, when we add it to jenkins, we should at least create a ticket to track the same. (since we will need this only when we are using this in a batch job)
Even otherwise, the default should be true. and we can set to false only when using in a batch job.

aganesh-suse: On second thoughts, it should be fine to keep it in the .env file. Never mind. Am overthinking this.

aganesh-suse (approved):
LGTM

PR #165, opened 2024-09-16
Update test to accept channel from test args

mdrahman-suse:
#### Testing ####

Checklist:

### DIFF
--- a/pkg/testcase/upgradenodereplacement.go
+++ b/pkg/testcase/upgradenodereplacement.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/rancher/distros-test-framework/pkg/aws"
+	"github.com/rancher/distros-test-framework/pkg/customflag"
 	"github.com/rancher/distros-test-framework/shared"
 
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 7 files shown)


fmoral2:
nice catch!

you will probably need to update `/scripts/test_runner.sh` as well

pkg/testcase/upgradenodereplacement.go
```diff
@@ -74,7 +78,8 @@ func TestUpgradeReplaceNode(cluster *shared.Cluster, version string) {
 
 func nodeReplaceAgents(
 	cluster *shared.Cluster,
-	version string,
+	version,
```
ShylajaDevadiga, line 81: have channel in same line?

pkg/testcase/upgradenodereplacement.go
```diff
@@ -287,7 +292,8 @@ func joinRemainServers(
 	oldServerIPs []string,
 	serverLeaderIp,
 	token,
-	version string,
+	version,
+	channel string,
```
ShylajaDevadiga, line 296: same. add to line above with version

pkg/testcase/upgradenodereplacement.go
mdrahman-suse: I think its better keep the vars in separate line

pkg/testcase/upgradenodereplacement.go
mdrahman-suse: I think its better keep the vars in separate line

fmoral2 (approved):

Issue #164, opened 2024-09-16
[Feature] Terraform destroy is not working on node replacement


mdrahman-suse:
As the instances are getting created via AWS cli, so even when terraform destroy flag is set to true, the instances does not get destroyed as terraform does not know about those instances. 
We need to handle this scenario to not overuse resources

PR #166, opened 2024-09-17
Allow deletion of replaced nodes

mdrahman-suse:
#### Proposed Changes ####

Address: https://github.com/rancher/distros-test-framework/issues/164

No
#### Types of Changes ####

Fix

#### Testing ####

Checklist:

### DIFF
--- a/entrypoint/restartservice/restartservice_test.go
+++ b/entrypoint/restartservice/restartservice_test.go
@@ -100,11 +100,6 @@ var _ = Describe("Test:", func() {
 			testcase.TestServiceLoadBalancer(false, true)
 		})
 	}
-
-	It("Displays Cluster details after tests", func() {
-		testcase.DisplayClusterDetails()
-	})
-
 })
 
 var _ = AfterEach(func() {
--- a/entrypoint/restartservice/restartservice_test.go
+++ b/entrypoint/restartservice/restartservice_test.go
@@ -101,8 +101,8 @@ var _ = Describe("Test:", func() {
 		})
 	}
 
-	It("Displays Cluster details", func() {
-		testcase.TestDisplayClusterDetails()
+	It("Displays Cluster details after tests", func() {
+		testcase.DisplayClusterDetails()
 	})
 
 })
--- a/entrypoint/upgradecluster/upgradenodereplacement_test.go
+++ b/entrypoint/upgradecluster/upgradenodereplacement_test.go
@@ -14,7 +14,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 )
 
-var _ = Describe("Upgrade Node Replcement Test:", Ordered, func() {
+var _ = Describe("Upgrade Node Replacement Test:", Ordered, func() {
 	It("Start Up with no issues", func() {
 		testcase.TestBuildCluster(cluster)
 	})
--- a/entrypoint/upgradecluster/upgradenodereplacement_test.go
+++ b/entrypoint/upgradecluster/upgradenodereplacement_test.go
@@ -4,12 +4,10 @@ package upgradecluster
 
 import (
 	"fmt"
-	"sync"
 
 	"github.com/rancher/distros-test-framework/pkg/assert"
 	"github.com/rancher/distros-test-framework/pkg/aws"
 	"github.com/rancher/distros-test-framework/pkg/testcase"
-	"github.com/rancher/distros-test-framework/shared"
 
 	. "github.com/onsi/ginkgo/v2"
 )
@@ -85,32 +83,11 @@ var _ = Describe("Upgrade Node Replacement Test:", Ordered, func() {
 
 	AfterAll(func() {
 		if flags.Destroy {
-			deleteAWSResources()
+			aws.DeleteEC2Instances(cluster)
 		}
 	})
 
 })
-
-func deleteAWSResources() {
-	ips := shared.FetchNodeExternalIPs()
-	awsClient, err = aws.AddClient(cluster)
-	if err != nil {
-		shared.LogLevel("error", "error creating aws client: %w\n", err)
-	}
-	var wg sync.WaitGroup
-	for _, ip := range ips {
-		wg.Add(1)
-		go func(ip string) {
-			defer wg.Done()
-			nodeDelErr := awsClient.DeleteInstance(ip)
-			if nodeDelErr != nil {
-				shared.LogLevel("error", "on deleting node with ip: %v, got error %w", ip, nodeDelErr)
-				return
-			}
-		}(ip)
-	}
-	wg.Wait()
-}
 
 var _ = AfterEach(func() {
 	if CurrentSpecReport().Failed() {
--- a/entrypoint/upgradecluster/upgradenodereplacement_test.go
+++ b/entrypoint/upgradecluster/upgradenodereplacement_test.go
@@ -79,12 +79,6 @@ var _ = Describe("Test:", func() {
 			testcase.TestServiceLoadBalancer(false, true)
 		})
 	}
-
-	if flags.Destroy {
-		It("Deletes replaced nodes", func() {
-			testcase.DeleteReplacedNodes(cluster)
-		})
-	}
 })
 
 var _ = AfterEach(func() {
--- a/entrypoint/upgradecluster/upgradenodereplacement_test.go
+++ b/entrypoint/upgradecluster/upgradenodereplacement_test.go
@@ -79,10 +79,6 @@ var _ = Describe("Test:", func() {
 			testcase.TestServiceLoadBalancer(false, true)
 		})
 	}
-
-	// Delete aws nodes created by cli
-	testcase.DeleteAWSInstance(cluster)
-	
 })
 
 var _ = AfterEach(func() {
@@ -91,4 +87,21 @@ var _ = AfterEach(func() {
 	} else {
 		fmt.Printf("\nPASSED! %s\n\n", CurrentSpecReport().FullText())
 	}
+})
+
+var _ = AfterAll(func() {
+	ips := shared.FetchNodeExternalIPs()
+	var wg sync.WaitGroup
+	for _, ip := range ips {
+		wg.Add(1)
+		go func(ip string) {
+			defer wg.Done()
+			nodeDelErr := awsClient.DeleteInstance(ip)
+			if nodeDelErr != nil {
+				shared.LogLevel("error", "on deleting node with ip: %v, got error %w", ip, nodeDelErr)
+				return
+			}
+		}(ip)
+	}
+	wg.Wait()
 })
--- a/entrypoint/upgradecluster/upgradenodereplacement_test.go
+++ b/entrypoint/upgradecluster/upgradenodereplacement_test.go
@@ -79,6 +79,10 @@ var _ = Describe("Test:", func() {
 			testcase.TestServiceLoadBalancer(false, true)
 		})
 	}
+
+	It("Deletes nodes created by aws cli", func() {
+		testcase.DeleteAWSInstance(cluster)
+	})
 })
 
 var _ = AfterEach(func() {
--- a/entrypoint/upgradecluster/upgradenodereplacement_test.go
+++ b/entrypoint/upgradecluster/upgradenodereplacement_test.go
@@ -7,13 +7,14 @@ import (
 	"sync"
 
 	"github.com/rancher/distros-test-framework/pkg/assert"
+	"github.com/rancher/distros-test-framework/pkg/aws"
 	"github.com/rancher/distros-test-framework/pkg/testcase"
 	"github.com/rancher/distros-test-framework/shared"
 
 	. "github.com/onsi/ginkgo/v2"
 )
 
-var _ = Describe("Test:", func() {
+var _ = Describe("Upgrade Node Replcement Test:", Ordered, func() {
 	It("Start Up with no issues", func() {
 		testcase.TestBuildCluster(cluster)
 	})
@@ -81,18 +82,21 @@ var _ = Describe("Test:", func() {
 			testcase.TestServiceLoadBalancer(false, true)
 		})
 	}
+
+	AfterAll(func() {
+		if flags.Destroy {
+			deleteAWSResources()
+		}
+	})
+
 })
 
-var _ = AfterEach(func() {
-	if CurrentSpecReport().Failed() {
-		fmt.Printf("\nFAILED! %s\n\n", CurrentSpecReport().FullText())
-	} else {
-		fmt.Printf("\nPASSED! %s\n\n", CurrentSpecReport().FullText())
-	}
-})
-
-var _ = AfterAll(func() {
+func deleteAWSResources() {
 	ips := shared.FetchNodeExternalIPs()
+	awsClient, err = aws.AddClient(cluster)
+	if err != nil {
+		shared.LogLevel("error", "error creating aws client: %w\n", err)
+	}
 	var wg sync.WaitGroup
 	for _, ip := range ips {
 		wg.Add(1)
@@ -106,4 +110,12 @@ var _ = AfterAll(func() {
 		}(ip)
 	}
 	wg.Wait()
+}
+
+var _ = AfterEach(func() {
+	if CurrentSpecReport().Failed() {
+		fmt.Printf("\nFAILED! %s\n\n", CurrentSpecReport().FullText())
+	} else {
+		fmt.Printf("\nPASSED! %s\n\n", CurrentSpecReport().FullText())
+	}
 })
--- a/entrypoint/upgradecluster/upgradenodereplacement_test.go
+++ b/entrypoint/upgradecluster/upgradenodereplacement_test.go
@@ -4,9 +4,11 @@ package upgradecluster
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/rancher/distros-test-framework/pkg/assert"
 	"github.com/rancher/distros-test-framework/pkg/testcase"
+	"github.com/rancher/distros-test-framework/shared"
 
 	. "github.com/onsi/ginkgo/v2"
 )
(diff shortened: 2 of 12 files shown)


endawkins (dismissed):
LGTM

VestigeJ (dismissed):

aganesh-suse (approved):
LGTM

pkg/testcase/upgradenodereplacement.go
```diff
@@ -598,3 +598,25 @@ func buildRke2Cmd(
 
 	return cmd, nil
 }
+
+// DeleteReplacedNodes deletes the new nodes created as replacements.
+func DeleteReplacedNodes(cluster *shared.Cluster) {
```
fmoral2, line 603: this really should not live here as a test, you are narrow it down too much

this should live on cluster config, and we have destroy cluster there, we should have a function `deleteAwsInstances` and call it wherever we want on suites that might need ,  not being specific for node replace

fmoral2 (changes requested):
delete func and flag

pkg/testcase/upgradenodereplacement.go
```diff
@@ -598,3 +598,25 @@ func buildRke2Cmd(
...
+	externalIPs := shared.FetchNodeExternalIPs()
```
fmoral2, line 604: you dont need this func since you have the cluster data 

```
   var ips []string
    ips = append(cluster.ServerIPs, ips...)
    ips = append(cluster.AgentIPs, ips...)


for _, ip : range ips ....

```

pkg/testcase/upgradenodereplacement.go
```diff
@@ -598,3 +598,25 @@ func buildRke2Cmd(
...
+		wg.Add(1)
+		go func(ip string) {
+			defer wg.Done()
+			nodeDelErr := awsClient.DeleteInstance(ip)
+			if nodeDelErr != nil {
+				shared.LogLevel("error", "on deleting node with ip: %v, got error %w", ip, nodeDelErr)
+				return
+			}
+		}(ip)
+		wg.Wait()
```
fmoral2, line 620: the wait should be outside for loop, this way you are actually doing nothing

pkg/testcase/upgradenodereplacement.go
```diff
@@ -598,3 +598,25 @@ func buildRke2Cmd(
 
 	return cmd, nil
 }
+
+// DeleteReplacedNodes deletes the new nodes created as replacements.
+func DeleteReplacedNodes(cluster *shared.Cluster) {
+	externalIPs := shared.FetchNodeExternalIPs()
+	awsClient, err := aws.AddAWSClient(cluster)
+	Expect(err).NotTo(HaveOccurred(), "error creating aws client: %s", err)
```
fmoral2, line 606: to print error  %v or %w

pkg/testcase/upgradenodereplacement.go
mdrahman-suse: So this cannot be added to shared/clusterconfig.go due to circular dependency with aws.go... any other suggestion where this can live?

pkg/testcase/upgradenodereplacement.go
mdrahman-suse: FYI I copied the code from here which has similar wait inside the for loop... so this needs to be updated as well then? https://github.com/rancher/distros-test-framework/blob/main/entrypoint/rebootinstances/rebootinstances_suite_test.go#L74

mdrahman-suse:
Updated please re-review

pkg/testcase/upgradenodereplacement.go
fmoral2: yeah makes sense. since this is actual aws implementation, should live in aws.go? 

then on suite 
```

var _ = AfterSuite(func() {
	if customflag.ServiceFlag.Destroy {
		aws.DeleteAwsNodes(cluster)
	}
})

```

pkg/testcase/upgradenodereplacement.go
fmoral2: yes please, sorry about that , doing that way we are actually not using the async benefits

fmoral2: if you dont mind updating, tks

pkg/testcase/upgradenodereplacement.go
fmoral2: also on suite you will need to a var for aws here 
```
var (
	kubeconfig string
	flags      *customflag.FlagConfig
	cluster    *shared.Cluster
)
```

to be able to reuse it on delete function and also pass it to `testcase.TestUpgradeReplaceNode(cluster,aws,  flags)`

pkg/testcase/upgradenodereplacement.go
mdrahman-suse: np will update

pkg/testcase/upgradenodereplacement.go
mdrahman-suse: this approach does not work as it tries to find the old ips, which got deleted as part of tf destroy...

pkg/testcase/upgradenodereplacement.go
mdrahman-suse: So we already have `DeleteInstance` in aws.go, which I have used in this func so I don't think it should live there as it deals with cluster stuff... one of the reason I wanted to have a `helper`/`support` dir that will live along with or close to the testcases

entrypoint/upgradecluster/upgradenodereplacement_test.go
```diff
@@ -80,6 +79,12 @@ var _ = Describe("Test:", func() {
 			testcase.TestServiceLoadBalancer(false, true)
 		})
 	}
+
+	if flags.Destroy {
```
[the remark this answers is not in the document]
mdrahman-suse: Actually adding it on the test makes sense as it is applicable to only this test, so keeping it as is

entrypoint/restartservice/restartservice_test.go
```diff
@@ -101,8 +101,8 @@ var _ = Describe("Test:", func() {
 		})
 	}
 
-	It("Displays Cluster details", func() {
-		testcase.TestDisplayClusterDetails()
+	It("Displays Cluster details after tests", func() {
+		testcase.DisplayClusterDetails()
```
ShylajaDevadiga, line 105: 1. Is this change related to this PR? If not its unnecessary for the reviewer to chase this change in the PR context.
2. Why do we need DisplayClusterDetails, why not call the underlying methods instead?

fmoral2: actually this should not live here since its not a test, not a specific test helper.

should live on shared and not also be used as a `It` spec, since its not a actual test or feat

entrypoint/upgradecluster/upgradenodereplacement_test.go
```diff
@@ -80,6 +79,10 @@ var _ = Describe("Test:", func() {
 			testcase.TestServiceLoadBalancer(false, true)
 		})
 	}
+
+	It("Deletes nodes created by aws cli", func() {
+		testcase.DeleteAWSInstance(cluster)
```
fmoral2, line 84: should live on shared and not also be used as a It spec, since its not a actual test or feat

pkg/testcase/upgradenodereplacement.go
fmoral2: upgradeerplacement_test.go
```


.....



var _ = AfterAll(func() {
	ips := append(cluster.ServerIPs, cluster.AgentIPs...)
	for _, ip := range ips {
		err := aws.DeleteInstance(ip)
		Expect(err).ToNot(HaveOccurred())
	}
})


```

fmoral2: suite
```
var (
	kubeconfig string
	flags      *customflag.FlagConfig
	cluster    *shared.Cluster
	aws        *c.Client
)

func TestMain(m *testing.M) {
	flags = &customflag.ServiceFlag
	flag.Var(&flags.InstallMode, "installVersionOrCommit", "Upgrade with version or commit")
	flag.Var(&flags.Channel, "channel", "channel to use on upgrade")
	flag.Var(&flags.Destroy, "destroy", "Destroy cluster after test")
	flag.Var(&flags.SUCUpgradeVersion, "sucUpgradeVersion", "Version for upgrading using SUC")
	flag.Parse()

	_, err := config.AddEnv()
	if err != nil {
		shared.LogLevel("error", "error adding env vars: %w\n", err)
		os.Exit(1)
	}

	kubeconfig = os.Getenv("KUBE_CONFIG")
	if kubeconfig == "" {
		// gets a cluster from terraform.
		cluster = shared.ClusterConfig()
	} else {
		// gets a cluster from kubeconfig.
		cluster = shared.KubeConfigCluster(kubeconfig)
	}

	aws, err = c.AddClient(cluster)

	os.Exit(m.Run())
}
```

pkg/testcase/upgradenodereplacement.go
fmoral2: https://github.com/rancher/distros-test-framework/pull/166#discussion_r1788063328

entrypoint/restartservice/restartservice_test.go
mdrahman-suse: Updated

pkg/testcase/upgradenodereplacement.go
mdrahman-suse: Updated

fmoral2 (dismissed):

mdrahman-suse:
Updated please review

entrypoint/upgradecluster/upgradenodereplacement_test.go
```diff
@@ -4,15 +4,17 @@ package upgradecluster
...
-	"github.com/rancher/distros-test-framework/pkg/customflag"
+	"github.com/rancher/distros-test-framework/pkg/aws"
 	"github.com/rancher/distros-test-framework/pkg/testcase"
+	"github.com/rancher/distros-test-framework/shared"
 
 	. "github.com/onsi/ginkgo/v2"
 )
 
-var _ = Describe("Test:", func() {
+var _ = Describe("Upgrade Node Replcement Test:", Ordered, func() {
```
fmoral2, line 17: typo on `Replcement`

entrypoint/upgradecluster/upgradenodereplacement_test.go
```diff
@@ -80,8 +82,35 @@ var _ = Describe("Test:", func() {
...
 		})
 	}
+
+	AfterAll(func() {
+		deleteAWSResources()
+	})
+		
 })
 
+func deleteAWSResources() {
```
fmoral2, line 92: can we add that on suite file? just to keep those files as clean as possible, without any logics ?

fmoral2: also on suite if we can create aws client on main and pass it through like we do with cluster,flags etc... just to keep the pattern of initializing dependencies on main.

mdrahman-suse: so I tried doing that, not sure why but it seems to fetch the node external ip from the previous cluster

entrypoint/upgradecluster/upgradenodereplacement_test.go
mdrahman-suse: Updated

scripts/delete_resources.sh
```diff
@@ -263,13 +263,13 @@ if [ "${RESOURCES}" = "" ]; then
     fi
 
     #Validate path to the tfvars file
-    if [[ ! -f "${$CONFIG_DIR}"/"$PRODUCT_NAME".tfvars ]]; then
+    if [[ ! -f "${CONFIG_DIR}"/"$PRODUCT_NAME".tfvars ]]; then
```
fmoral2, line 266: tks for seeing that!

entrypoint/upgradecluster/upgradenodereplacement_test.go
fmoral2: Also cannot call it on the suite as the suite applies to all the test and this func is needed only for this test

>  agreed

fmoral2: thats really odd, cause the `aws.AddClient` has nothing to do with cluster.IPS stuff.

fmoral2: but thats ok we can look at that later

fmoral2 (approved):

entrypoint/upgradecluster/upgradenodereplacement_test.go
mdrahman-suse: Yeah I was confused as well as to why it was not working like that

fmoral2: sorry ....really not trying to be picky , but this is like we are doing something that we fought not do to until today lol.....adding logic on  IT specs entrypoint.....

can we figure it out a way do call use this on the logic place? since you are adding this as a specific helper for node replace, can we add this on `/pkg/testcase/nodereplacement.go` ? 

i know that adding there will be a lit bit of trick , but sure is doable

fmoral2: @mdrahman-suse

mdrahman-suse: Let me try

mdrahman-suse: Updated, please take a look @fmoral2

fmoral2: sorry for all . let me try again 


i wanted to mean that it should not live on  `/entrypoint/...`  on a IT spec , since its not a test or nor a feature, and even adding there a just a after something was polluting that file , which in theory should be like a file for " outsiders" to see whats been tested on a feature/flow without needing to code on actual code 
not saying that i like it, just saying its purpose.


since we are treating that ( and i agree ) as a upgrade node replacement helper , as other ones like 
`delErr := deleteRemainServer(serverLeaderIP, awsDependencies)`
`func nodeReplaceServers(`
`serverJoin()`
and etc, this should live on the same file or as closest as possible.


- Adding that as a aws utils does not suite better , because the idea of a client pkg, is kind of  "provide an interface and methods of something, so you can use it wherever its needed" so you probably should be exporting this `DeleteEC2Instances()` as a method, not as a util func, here util func would be util or helpers for the pkg itself, not for  outside , similar to this here 
- https://github.com/rancher/distros-test-framework/pull/177/files#diff-d050ce959548e8086404b5d8d53e4f2ddce5917e83a082106bd34416b306244b

And adding that as a method would not make much sense cause we already have a method exported `DELETEINSTANCE`


Thats why where it would be fitting nicer is on the actual test case , because there you are importing a client to make use of it ( aws) and based on methods provided ( DELETE INSTANCE) you can use it and create a custom helper for your needs `DeleteEC2Instances()`

mdrahman-suse: So if we add the func in the same file, when and where should we call the func?

mdrahman-suse:
Updated please review @fmoral2

entrypoint/upgradecluster/upgradenodereplacement_test.go
```diff
@@ -6,13 +6,13 @@ import (
 	"fmt"
 
 	"github.com/rancher/distros-test-framework/pkg/assert"
-	"github.com/rancher/distros-test-framework/pkg/customflag"
 	"github.com/rancher/distros-test-framework/pkg/testcase"
+	"github.com/rancher/distros-test-framework/pkg/testcase/helper"
```
ShylajaDevadiga, line 10: Can we call it helpers instead?

mdrahman-suse: Changes it to `support`

ShylajaDevadiga (approved):

pkg/testcase/support/awsutils.go
```diff
@@ -1,14 +1,16 @@
-package aws
+package support
```
fmoral2, line 1: do you mind the file be something like: 
- `aws.go`  or `instances.go`.  `ec2s.go` , `infra.go`. something on this path ?

fmoral2: @mdrahman-suse

fmoral2: we dont need utils in the name, since its already in a support/helper pkg

pkg/testcase/upgradenodereplacement.go
```diff
@@ -27,13 +28,14 @@ func TestUpgradeReplaceNode(cluster *shared.Cluster, flags *customflag.FlagConfi
...
 	resourceName := os.Getenv("resource_name")
+	fmt.Println("resource-name: ", resourceName)
 	awsDependencies, err := aws.AddClient(cluster)
 	Expect(err).NotTo(HaveOccurred(), "error adding aws nodes: %s", err)
 
 	// create server names.
 	var serverNames, instanceServerIds, newExternalServerIps, newPrivateServerIps []string
 	for i := 0; i < len(cluster.ServerIPs); i++ {
-		serverNames = append(serverNames, fmt.Sprintf("%s-distros-qa-server-replace-%d", resourceName, i+1))
+		serverNames = append(serverNames, fmt.Sprintf("%s-%s-server-replace%d", resourceName, resourceTag, i+1))
```
fmoral2, line 38: sorry we dont need `resourceTag`

I forgot to remove, since its already set on aws.
https://github.com/rancher/distros-test-framework/blob/4f65fe8b7223db7bd9bfabea68e92165e8ccc47a/pkg/aws/aws.go#L321

mdrahman-suse: I actually removed it from there and added another tag called `Team: distros-qa`... else it will be like `server-replace1-distros-qa`...

mdrahman-suse: Done

pkg/testcase/support/awsutils.go
mdrahman-suse: Updated

PR #168, opened 2024-09-23
test script and jenkinsfile for delete resources

aganesh-suse:
#### Proposed Changes ####

Update to the delete_resources.sh script - to take in a comma separated resource prefix name list. 
Retains the previous method of reading the tfvars resource_name var as well. The help documents these options in detail. 

Adding Jenkinsfile_delete_resources file that the jenkins job will call. This then will call the script and pass the resource prefix comma separated for deletion. 

#### Testing ####

Checklist:


scripts/Jenkinsfile_delete_resources
```diff
@@ -0,0 +1,58 @@
...
+        if (it.value && it.value.trim() != "") {
+          paramsMap << "$it.key=$it.value"
+        }
+      }
+      withCredentials([
+
+        string(credentialsId: 'AWS_ACCESS_KEY_ID', variable: 'AWS_ACCESS_KEY_ID'),
+        string(credentialsId: 'AWS_SECRET_ACCESS_KEY', variable: 'AWS_SECRET_ACCESS_KEY'),
+      ]) {
+        withAWS(region:'us-east-2') {
```
mdrahman-suse, line 32: I think we should make it user defined via a variable and have a default set there

aganesh-suse: fixed/not using this anymore

scripts/delete_resources.sh
```diff
@@ -1,149 +1,232 @@
...
+      echo "Deleting Route53 record ${NAME} for prefix $1"
+      CHANGE_STATUS=$(aws route53 change-resource-record-sets --hosted-zone-id "${R53_ZONE_ID}" --change-batch file://"${PWD}"/payload.json)
+      echo "CHANGE STATUS: 
+      $CHANGE_STATUS"
+      STATUS_ID=$(echo "$CHANGE_STATUS" | grep Id: | cut -d " " -f 4)
+      echo "Status ID: $STATUS_ID"
+      #Get status from the change
+      aws route53 wait resource-record-sets-changed --id "$STATUS_ID"
+      echo "Successfully deleted Route53 record ${NAME}: status: ${STATUS_ID}"
+      exit 1
```
aganesh-suse, line 150: intentional exit during testing in jenkins. should be removed.

scripts/delete_resources.sh
```diff
@@ -1,149 +1,240 @@
...
+      "
+      exit 1
+      ;;
+  esac
+done
+
+if [ "${RESOURCES}" = "" ]; then
+    echo "Working with local .env and .tfvars file to get resource name for deletion"
+    #Get resource name from tfvarslocal && change name to make more sense in this context
+    PRODUCT_NAME=$(grep ENV_PRODUCT <./config/.env | cut -d= -f2 | tr -d ' "')  # Split string based on delimiter =
```
fmoral2, line 204: just as a best practice and  even for safety, not adding comments sideways? lets keeps the comment always above

scripts/delete_resources.sh
```diff
@@ -1,149 +1,240 @@
...
+    if echo "${PRODUCT_NAME}" | grep -q "ENV_PRODUCT"; then
+      PRODUCT_NAME=$(echo "${PRODUCT_NAME}" | cut -d ":" -f 2)  # Split string based on delimiter :
```
fmoral2, line 206: from where we are getting that its coming with a ` : `  ?

scripts/delete_resources.sh
fmoral2, line 204: since we are refactoring the whole script can we also fix this? 
if we have 2 vars with the same name not commented , exit with an error and if   one of those is commented just return the correct one 

```
#ENV_PRODUCT=rke2

ENV_PRODUCT=k3s
ENV_TFVARS=k3s.tfvars
```

or 
```
ENV_PRODUCT=rke2

ENV_PRODUCT=k3s
ENV_TFVARS=k3s.tfvars
```
something like 
```
    PRODUCT_NAME=$(grep -v '^#' ./config/.env | grep ENV_PRODUCT | cut -d= -f2 | tr -d ' "')

  if  $PRODUCT_NAME -gt 1  then exit 

```

scripts/delete_resources.sh
```diff
@@ -1,149 +1,240 @@
...
+    fi
+
+    #Get resource name from tfvars file and validate
+    RESOURCE_NAME=$(grep resource_name <./config/"$PRODUCT_NAME".tfvars | cut -d= -f2 | tr -d ' "')
+    if [[ -z "$RESOURCE_NAME" ]]; then
+      echo "No resource name found for: $PRODUCT_NAME.tfvars file"
+      exit 1
+    fi
+
+    printf "This is going to delete all AWS resources with the prefix %s. Continue (yes/no)? " "$RESOURCE_NAME"
```
fmoral2, line 226: we also need to fix that when i call it with resource name , its not prompting me if i want to proceed with this name, its just going wild and free and go do his stuff. this is too dangerous.
```
 bash scripts/delete_resources.sh -r ANYNAME

## For prefix name: ANYNAME:
No ec2 instances found with prefix: ANYNAME. Nothing to delete.
No db instances found with prefix ANYNAME. Nothing to delete.
No db clusters found with prefix ANYNAME. Nothing to delete.


```

scripts/delete_resources.sh
fmoral2: same for tfvars

scripts/delete_resources.sh
```diff
@@ -1,149 +1,240 @@
...
+    else
+      echo "Exiting: No resources deleted as per user input. Please delete the resources manually"
+      exit 1
+    fi
 else
-  echo "Exiting: No resources deleted as per user input. Please delete the resources manually"
-  exit 1
-fi
\ No newline at end of file
+    for i in $(echo "${RESOURCES}" | tr "," "\n")
```
fmoral2, line 235: for sake of performance the for loop should be ideally done in the functions.
so you should pass the whole list to delete_all 

and then, any reason why doing this sequentially ? its going to take a huge time running this for a lot of resource names ( since we have that option now) 

we should pass the list to each fucntion, inside the function do a loop with all resource names and on the  caller we should do somthing like 
```
delete_all_resources () {
 delete_ec2_instances "$1" &
  delete_db_resources "$1" &
  delete_lb_resources "$1" &
  delete_target_groups "$1" &
  delete_route_s3 "$1" &

    wait
}
```

scripts/delete_resources.sh
```diff
@@ -1,149 +1,240 @@
...
+}
 
-  #Loop through the target group ARNs and delete the target groups
-  for TG_ARN in $TG_ARN_LIST; do
-    echo "Deleting target group $TG_ARN"
-    aws elbv2 delete-target-group --target-group-arn "$TG_ARN"
-  done
-
-
+delete_route_s3 () {
```
fmoral2, line 101: route 53

scripts/delete_resources.sh
fmoral2: if by any change i do , locally or jenkins 
```
 bash scripts/delete_resources.sh -r a
```

it will delete a bunch of stuff from other people also right

scripts/delete_resources.sh
```diff
@@ -1,149 +1,240 @@
...
+  delete_target_groups "$1"
+  delete_route_s3 "$1"
+}
+
+
+while getopts r:h OPTION
+do 
+  case "${OPTION}"
+    in
+    r) RESOURCES=${OPTARG};;
```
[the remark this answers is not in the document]
aganesh-suse: jenkins jobs get created only with lower case values. AFAIK the jobs will fail if we had upper case or special characters(i could be wrong). But if they had upper case when they manually created a setup, changing to lower case would prevent them from using this to delete that setup. i feel putting that restriction should not be on this script. the user should know what they want to delete and provide that.

scripts/delete_resources.sh
aganesh-suse: fixed

scripts/delete_resources.sh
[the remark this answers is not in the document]
aganesh-suse: the worst that will happen, is we wont find resources for that name and we wont do anything. but in the case of having resource names with capitals - it will still look for it and delete them if we want to. there is value in that.

scripts/delete_resources.sh
aganesh-suse: yaml format values use a colon. if by mistake a person edits the .env file manually and puts a yaml format here instead of = it should still work. circumventing manual human errors.

scripts/delete_resources.sh
aganesh-suse: uhmm... how do u make jenkins job interactive? like you would go onto the console page and then enter yes on that page?

aganesh-suse: we could potentially have a dry run option maybe? get a param from jenkins - say dry run true, and in here, we could add another var -d, which will only maybe print the resource list and not do the actual delete?

aganesh-suse: from all my reading, jenkins can take in another param from you - say "yes" and pass that via jenkinsfile to the interactive script say ./delete_resources.sh -r abc << input". this is of no use to us, if we are going to provide a yes anyway when we want to delete.

scripts/delete_resources.sh
aganesh-suse: the deletes happen pretty fast. i hv created like 20 30 resources per group, and the max it has run is maybe 2 mins. the timeout for the job is 30 mins right now. i want it to log the for this prefix - these resources are being deleted - and they should be deleted together in a group. i dont want all ec2 instances all prefix one set, all db clusters all prefix etc. its for easier readability and follow up to  make sure everything got deleted for one prefix first. its log readability that is more important here than performance of a for loop.

scripts/delete_resources.sh
fmoral2: not seeing fixed

scripts/delete_resources.sh
fmoral2: nice ok

scripts/delete_resources.sh
fmoral2: ok

scripts/delete_resources.sh
fmoral2: honestly idk, but at least in jenkins we have a momentary history of the run and etc...locally we dont.
at least the script needs to check if , running != jenkins, prompt the question first.

in jenkins then its mandatory to run with `RESOURCE_PREFIX_LIST` ?

scripts/delete_resources.sh
aganesh-suse: missed push. sorry fixed now. :)

scripts/delete_resources.sh
aganesh-suse: yes - it is mandatory to run with RESOURCE_PREFIX_LIST having a value in jenkins. if you provided empty value, the job will abort. when you provide -r value locally or in the jenkins - you are manually typing the value. we are not reading something from a place you might not know about. this is an intentional choice on a person to go type the exact value/letters they want to delete.

aganesh-suse: if you run the script without providing the -r option - that still does the yes/no check for confirmation. that is the only place when you are not aware of what the prefix name is. since you did not manually type this and ask for deletion.

scripts/delete_resources.sh
aganesh-suse: fixed

scripts/delete_resources.sh
aganesh-suse: fixed

scripts/delete_resources.sh
aganesh-suse: added the dry-run option as well.
fixed.

mdrahman-suse (approved):
LGTM

fmoral2 (dismissed):

mdrahman-suse (approved):
LGTM

scripts/Dockerfile.build
```diff
@@ -24,7 +24,10 @@ RUN apk update && \
     export VERIFY_CHECKSUM=false && \
     curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash && \
     rm -rf tmp && \
-    mkdir -p /root/.ssh
+    mkdir -p /root/.ssh && \
+    mkdir -p /root/.aws
+
+RUN aws --version   # Just to make sure its installed alright
```
fmoral2, line 30: can we add the comment above?

fmoral2: sorry OCD

fmoral2 (approved):

PR #194, opened 2024-11-27
Secrets-Encryption - restart primary etcd first in a 3 etcd setup

aganesh-suse:
#### Proposed Changes ####

The primary etcd node needs to be restarted first, before restarting the rest of the etcd/control plane nodes. 
So, added a shared method to get the primary etcd node ip. GetPrimaryEtcdIp in cluster.go
Post restart of the primary etcd node, the rest of the servers will be restarted - etcd first, control plane next. 

#### Testing ####

Checklist:


pkg/testcase/secretsencrypt.go
```diff
@@ -47,6 +47,8 @@ func secretsEncryptOps(action, product, cpIP string, nodes []shared.Node) {
 	if (action == "reencrypt") || (action == "rotate-keys") {
 		shared.LogLevel("DEBUG", "reencrypt op needs some time to complete - Sleep for 20 seconds before service restarts")
 		time.Sleep(20 * time.Second) // Wait for reencrypt action to complete before restarting services.
+	} else {
```
fmoral2, line 50: instead adding more sleeps, can we refactor secrets encrypts to use k8s pkg?

fmoral2: we have a lot of places here that we can remove all sleeps and add validations/waits  from there

mdrahman-suse:
Can you please update the suite to fail when the flag `secrets-encryption: true` is not passed on the server_flags?

aganesh-suse:
> Can you please update the suite to fail when the flag `secrets-encryption: true` is not passed on the server_flags? Also we dont support 1.27 anymore so the code for that flow can be removed

Only in the case of k3s, do we need the secrets-encryption: true value in the server flags. We ahve a check for this in the test suite here: https://github.com/rancher/distros-test-framework/blob/main/entrypoint/secretsencrypt/encryption_suite_test.go#L56 
Removed the check for 1.27/1.26 in both the testcase and the test suite files.

pkg/testcase/secretsencrypt.go
aganesh-suse: We need some minimum sleeps - since the actual prepare/rotate/reencrypt ops happen in the background - there is no quantifier/verifier for this - and the validation/waits of pods/nodes are not representative of this. This minimum wait is unavoidable. We have checks for pods/nodes as well.

entrypoint/secretsencrypt/encryption_suite_test.go
```diff
@@ -56,17 +55,6 @@ var _ = BeforeSuite(func() {
 		Expect(os.Getenv("server_flags")).To(ContainSubstring("secrets-encryption:"),
 			"ERROR: Add secrets-encryption:true to server_flags for this test")
```
mdrahman-suse, line 56: Should exit out using `os.Exit(1)` as the test will continue to run even if this fails

pkg/testcase/secretsencrypt.go
```diff
@@ -44,40 +41,60 @@ func secretsEncryptOps(action, product, cpIP string, nodes []shared.Node) {
...
 	verifyActionStdOut(action, stdOutput)
-	if (action == "reencrypt") || (action == "rotate-keys") {
-		shared.LogLevel("DEBUG", "reencrypt op needs some time to complete - Sleep for 20 seconds before service restarts")
-		time.Sleep(20 * time.Second) // Wait for reencrypt action to complete before restarting services.
+
+	shared.LogLevel("debug", "secrets-encrypt ops need extra time to complete - Sleep for 30 seconds before service restarts")
+	time.Sleep(30 * time.Second)
+
+	// Restart Primary Etcd Node First
+	etcdIp, getError := shared.GetPrimaryEtcdIp(product)
```
mdrahman-suse, line 49: What if the primary is a control-plane node, will it work in that scenario?

entrypoint/secretsencrypt/encryption_suite_test.go
aganesh-suse: Fixed.

pkg/testcase/secretsencrypt.go
aganesh-suse: The primary node will most definitely be an etcd node. It can be only etcd or a combination of etcd,cp. That said, the GetPrimaryEtcdIp method reads the config.yaml file to get the primary node ip. The server IP we provide there - will always be the primary node ip - be it the config.yaml on a secondary etcd node/cp node/agent node. And a lack of the server line - means it is the primary node itself. So we can reference the config.yaml file from any node to still figure out the primary node.

pkg/testcase/secretsencrypt.go
```diff
@@ -30,10 +29,8 @@ func TestSecretsEncryption() {
 	secretsEncryptOps("rotate", product, cpIp, nodes)
 	secretsEncryptOps("reencrypt", product, cpIp, nodes)
 
-	if strings.Contains(os.Getenv("TEST_TYPE"), "both") {
-		shared.LogLevel("info", "TEST: 'NEW' Secrets Encryption method")
-		secretsEncryptOps("rotate-keys", product, cpIp, nodes)
-	}
+	shared.LogLevel("info", "TEST: 'NEW' Secrets Encryption method")
```
mdrahman-suse, line 32: Do we need to keep this? as this is the only method that is supported ATM?

pkg/testcase/secretsencrypt.go
mdrahman-suse: Yeah lets keep it simple `GetPrimaryNodeIP`

pkg/testcase/secretsencrypt.go
```diff
@@ -44,43 +41,57 @@ func secretsEncryptOps(action, product, cpIP string, nodes []shared.Node) {
...
 }
 
+func restartServerAndWait(ip, product string) {
+	nodearr := []string{ip}
+	nodeIP, errRestart := shared.ManageService(product, "restart", "server", nodearr)
+	Expect(errRestart).NotTo(HaveOccurred(), "error restart service for node: "+nodeIP)
+	// Little lag needed between node restarts to avoid issues.
+	shared.LogLevel("debug", "Sleep for 30 seconds before service restarts between servers")
+	time.Sleep(30 * time.Second)
+	waitEtcdErr := shared.WaitForPodsRunning(10, 3)
```
mdrahman-suse, line 89: We should also wait for the node to be ready?

pkg/testcase/secretsencrypt.go
```diff
@@ -44,43 +41,57 @@ func secretsEncryptOps(action, product, cpIP string, nodes []shared.Node) {
...
 	secretEncryptStatus, errGetStatus := waitForHashMatch(cpIP, product)
 	Expect(errGetStatus).NotTo(HaveOccurred(), "error getting secret-encryption status")
 	verifyStatusStdOut(action, secretEncryptStatus)
 
-	errLog := logEncryptionFileContents(nodes, product)
+	errLog := logEncryptionFileContents(nodes, action, product)
 	Expect(errLog).NotTo(HaveOccurred())
 }
 
+func restartServerAndWait(ip, product string) {
```
mdrahman-suse, line 82: 👍

pkg/testcase/secretsencrypt.go
```diff
@@ -90,16 +101,16 @@ func waitForHashMatch(cpIP, product string) (string, error) {
 	for i := 1; i <= times; i++ {
 		secretEncryptStatus, errGetStatus = shared.SecretEncryptOps("status", cpIP, product)
 		if errGetStatus != nil {
-			shared.LogLevel("DEBUG", "error getting secret-encryption status. Retry.")
+			shared.LogLevel("debug", "error getting secret-encryption status. Retry.")
 		}
 		if secretEncryptStatus != "" && strings.Contains(secretEncryptStatus, "All hashes match") {
-			shared.LogLevel("DEBUG", "Total sleep time before hashes matched: %d seconds", i*int(defaultTime))
+			shared.LogLevel("debug", "Total sleep time before hashes matched: %d seconds", i*int(defaultTime))
```
mdrahman-suse, line 107: Let rephrase this message as we don't actually wait initially or if you want to reflect the same then lets add the wait after for loop

pkg/testcase/secretsencrypt.go
aganesh-suse: it helps to segregate the output - when u read the logs - like both the reencrypt and rotate-keys outputs look similar.. you can search for the TEST keyword and go back and forth. its just for easier traversal of the logs. I can remove if it is not desirable.

aganesh-suse: it was easier to search through logs - since both reencrypt and rotate-keys outputs look exactly the same. But I can remove this.

shared/cluster.go
```diff
@@ -337,6 +337,36 @@ func GetNodesByRoles(roles ...string) ([]Node, error) {
 	return nodes, nil
 }
 
+// GetPrimaryEtcdIp returns the primary etcd node public/external IP as a string, and error if not found.
+// provide product k3s|rke2 string as input.
+func GetPrimaryEtcdIp(product string) (string, error) {
```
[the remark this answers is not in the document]
aganesh-suse: removed.

pkg/testcase/secretsencrypt.go
aganesh-suse: removed

pkg/testcase/secretsencrypt.go
mdrahman-suse: One way can be to check for status of `reencrypt` or `rotate_keys`.. once they have status of `finished` then we can proceed which can help with removing this hard coded sleeps

pkg/testcase/secretsencrypt.go
mdrahman-suse: Yes and actually reflect that it waited for x seconds... coz I see even if you dont wait for x seconds right now it will print that it waited for 10 (default) seconds, which is not true

pkg/testcase/secretsencrypt.go
```diff
@@ -22,65 +24,72 @@ func TestSecretsEncryption() {
 	errSecret := shared.CreateSecret("secret1", "default")
 	Expect(errSecret).NotTo(HaveOccurred(), "error creating secret")
 
-	shared.LogLevel("info", "TEST: 'CLASSIC' Secrets Encryption method")
-
 	index := len(nodes) - 1
 	cpIp := nodes[index].ExternalIP
```
mdrahman-suse, line 28: Will it be always true that the last index in nodes array will be the CP node?

shared/product.go
```diff
@@ -51,6 +51,7 @@ func ManageService(product, action, nodeType string, ips []string) (string, erro
 	}
 
 	for _, ip := range ips {
+		LogLevel("debug", "Performing %s on %s", action, ip)
```
mdrahman-suse, line 54: nit: `Performing systemctl %s on IP: %s`

pkg/testcase/secretsencrypt.go
aganesh-suse: yes. we either have a setup with etcd + cp nodes. or 1 etcd 2 cp, 3 etcd, 2cp etc. we get the nodes by roles - etcd first and next cp. we wont have a setup without a cp node. and it will always be there on the last node - since we get them by roles in that order.

shared/product.go
aganesh-suse: fixed

pkg/testcase/secretsencrypt.go
aganesh-suse: I changed the for loop to start from 0 to < times. that way, if it matched right away - it will matched after 0 seconds. and not 10 seconds default.

fmoral2 (dismissed):

entrypoint/secretsencrypt/encryption_suite_test.go
```diff
@@ -51,23 +52,14 @@ func TestSecretsEncryptionSuite(t *testing.T) {
...
 }
 
-var _ = BeforeSuite(func() {
-	if cluster.Config.Product == "k3s" {
-		Expect(os.Getenv("server_flags")).To(ContainSubstring("secrets-encryption:"),
-			"ERROR: Add secrets-encryption:true to server_flags for this test")
+func validateSecretsEncryptFlag() {
+	if cfg.Product == "k3s" {
+		if !strings.Contains(os.Getenv("server_flags"), "secrets-encryption:") {
+			shared.LogLevel("error", "Add secrets-encryption:true to server_flags for this test")
```
mdrahman-suse, line 58: `Add secrets-encryption: true`

pkg/testcase/secretsencrypt.go
```diff
@@ -22,84 +24,91 @@ func TestSecretsEncryption() {
...
-		secretsEncryptOps("rotate-keys", product, cpIp, nodes)
-	}
+	secretsEncryptOps("prepare", product, cluster.ServerIPs[0], cpIp, nodes)
+	secretsEncryptOps("rotate", product, cluster.ServerIPs[0], cpIp, nodes)
+	secretsEncryptOps("reencrypt", product, cluster.ServerIPs[0], cpIp, nodes)
+	secretsEncryptOps("rotate-keys", product, cluster.ServerIPs[0], cpIp, nodes)
 }
 
-func secretsEncryptOps(action, product, cpIP string, nodes []shared.Node) {
+func secretsEncryptOps(action, product, primaryEtcdIp, cpIP string, nodes []shared.Node) {
```
mdrahman-suse, line 36: nit: `primaryNodeIP`

pkg/testcase/secretsencrypt.go
```diff
@@ -22,84 +24,91 @@ func TestSecretsEncryption() {
...
-	if (action == "reencrypt") || (action == "rotate-keys") {
-		shared.LogLevel("DEBUG", "reencrypt op needs some time to complete - Sleep for 20 seconds before service restarts")
-		time.Sleep(20 * time.Second) // Wait for reencrypt action to complete before restarting services.
-	}
+
+	shared.LogLevel("debug", "secrets-encrypt ops need to complete - Sleep for 30 seconds before service restarts")
+	time.Sleep(30 * time.Second)
+
+	// Restart Primary Etcd Node First
+	restartServerAndWait(primaryEtcdIp, product)
```
mdrahman-suse, line 49: Can we add debug log here before and after restart like `Restart service -> Starting...` `Restart service -> Completed!`

pkg/testcase/secretsencrypt.go
```diff
@@ -22,84 +24,91 @@ func TestSecretsEncryption() {
...
 	cpIp := nodes[index].ExternalIP
-	secretsEncryptOps("prepare", product, cpIp, nodes)
-	secretsEncryptOps("rotate", product, cpIp, nodes)
-	secretsEncryptOps("reencrypt", product, cpIp, nodes)
 
-	if strings.Contains(os.Getenv("TEST_TYPE"), "both") {
-		shared.LogLevel("info", "TEST: 'NEW' Secrets Encryption method")
-		secretsEncryptOps("rotate-keys", product, cpIp, nodes)
-	}
+	secretsEncryptOps("prepare", product, cluster.ServerIPs[0], cpIp, nodes)
```
mdrahman-suse, line 30: Can we add debug logging here before and after each action similar to https://github.com/rancher/distros-test-framework/pull/194/files#r1882119964

pkg/testcase/secretsencrypt.go
aganesh-suse: no - we need sleeps before we restart the servers and in between restart of servers - during this time, the status wont say finished. only post restart of the servers will the status say finished. so these sleeps are unavoidable. we do have test for checking the status of reencrypt and rotate-keys to finished in the logEncryptionFileContents method.  and the secrets-encrypt status output.

entrypoint/secretsencrypt/encryption_suite_test.go
aganesh-suse: Some day, we may want to test 'secrets-encryption: false' for an issue or something, say. The odds of that happening are really minimal. But if we put 'true' in, then the test wont run at all, and we will miss out an opportunity to reuse some existing automation. Thats why i did not add the true keyword there. Also, if someone has an extra space between 'secrets-encryption:  true'  would it fail then - and we check for contains 1 space - it would fail again - but falsely? We shouldnt limit this check to the exact value.

pkg/testcase/secretsencrypt.go
aganesh-suse: The log line is inside the secretsEncryptOps method: https://github.com/rancher/distros-test-framework/blob/main/pkg/testcase/secretsencrypt.go#L40
Added a debug log line at the end of the method action step completed as well.

pkg/testcase/secretsencrypt.go
aganesh-suse: fixed

entrypoint/secretsencrypt/encryption_suite_test.go
mdrahman-suse: I meant add an space on line 58 between `secrets-encryption:` and `true` so that it says `Add secrets-encryption: true to server_flags for this test`

pkg/testcase/secretsencrypt.go
mdrahman-suse: I am not seeing this change that you have added...

mdrahman-suse (approved):

fmoral2 (approved):

pkg/testcase/secretsencrypt.go
aganesh-suse: https://github.com/rancher/distros-test-framework/pull/194/files#diff-d178807df3cc2529cbf5b801c72ee6a584847d796a2c5693c7785b0e45f9c7f8R104

PR #196, opened 2024-12-04
update image to avoid rate limit

ShylajaDevadiga:
This PR is to update image  for local-path-provisioner test from rancher/mirrored-xxx so that it does not get rate limited

### DIFF
--- a/workloads/amd64/local-path-provisioner.yaml
+++ b/workloads/amd64/local-path-provisioner.yaml
@@ -39,7 +39,7 @@ spec:
     spec:
       containers:
         - name: volume-test
-          image: nginx:stable-alpine
+          image: rancher/mirrored-library-nginx:1.27.2-alpine
           imagePullPolicy: IfNotPresent
           volumeMounts:
             - name: volv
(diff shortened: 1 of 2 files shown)


workloads/amd64/local-path-provisioner.yaml
```diff
@@ -49,4 +49,4 @@ spec:
       volumes:
         - name: volv
           persistentVolumeClaim:
-            claimName: local-path-pvc
```
mdrahman-suse, line 52: nit: new line for EOF

mdrahman-suse:
I see v1.27.3 in this PR: https://github.com/k3s-io/k3s/pull/11401/files Can/Should we be using that @ShylajaDevadiga?

ShylajaDevadiga:
> I see v1.27.3 in this PR: https://github.com/k3s-io/k3s/pull/11401/files Can/Should we be using that @ShylajaDevadiga? Or its a different image?

That's the latest one we have in mirrored list

mdrahman-suse (approved):

aganesh-suse (approved):

PR #204, opened 2024-12-20
Sonobuoy Conformance Results

VestigeJ:
#### Proposed Changes ####

Adding the ability to run sonobuoy and get the results from the run of conformance results

the command to run it exists within the makefile and it has been edited to accommodate the long timeout window that conformance results frequently hit within go/ginkgo expectations.

#### Types of Changes ####

It can be a quick test by altering the launchSonobuoyTests("quick") but the default is to run the full conformance suite from upstream. Primarily to be used during patch validation runs but it could be used more frequently in other ways to quickly check a clusters status. Although if you have a cluster running the fastest and easiest way to run conformance is to do it the simple way upstream does it with a simple yaml deployment. 

```
---
apiVersion: v1
kind: Namespace
metadata:
  name: conformance
---
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    component: conformance
  name: conformance-serviceaccount
  namespace: conformance
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  labels:
    component: conformance
  name: conformance-serviceaccount-role
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: conformance-serviceaccount
subjects:
- kind: ServiceAccount
  name: conformance-serviceaccount
  namespace: conformance
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    component: conformance
  name: conformance-serviceaccount
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - '*'
- nonResourceURLs:
  - '/metrics'
  - '/logs'
  - '/logs/*'
  verbs:
  - 'get'
---
apiVersion: v1
kind: Pod
metadata:
  name: e2e-conformance-test
  namespace: conformance
spec:
  containers:
  - name: conformance-container
    image: registry.k8s.io/conformance-amd64:v1.31.2 #EDIT THIS FOR YOUR CLUSTER
    imagePullPolicy: IfNotPresent
    env:
    - name: E2E_FOCUS
      value: "\\[Conformance\\]"
    - name: E2E_SKIP
      value: ""
    - name: E2E_PROVIDER
      value: "skeleton"
    - name: E2E_PARALLEL
      value: "false"
    - name: E2E_VERBOSITY
      value: "4"
    volumeMounts:
    - name: output-volume
      mountPath: /tmp/results
  volumes:
  - name: output-volume
    hostPath:
      path: /tmp/results
  restartPolicy: Never
  serviceAccountName: conformance-serviceaccount
```

#### Testing ####

Checklist:
 not yet 

✅ 
✅ 
✅ 
✅


pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+			GinkgoT().Errorf("error: %v", err)
+			return
+		}
+	}
+}
+
+func ConformanceTest(cluster *shared.Cluster) {
+	verifyClusterNodes(cluster)
+	installConformanceBinary()
+	launchSonobuoyTests("certified-conformance")
```
mdrahman-suse, line 49: We should add a flag for this if it can be user defined

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+	// launchSonobuoyTests("quick")
+	testResultTar := checkStatusGetResults(cluster)
+	fmt.Println("testResultTar: ", testResultTar)
```
mdrahman-suse, line 52: use LogLevel func instead of print

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+		fmt.Println("cluster does not meet the minimum requirements to run conformance tests")
+		return false
+	} else {
+		return true
+	}
+}
+
+func installConformanceBinary() {
+	sonobuoyVersion := customflag.ServiceFlag.External.SonobuoyVersion
+	err := shared.SonobuoyMixedOS("install", sonobuoyVersion)
```
mdrahman-suse, line 69: May be rename the func to be `shared.InstallSonobuoy`

config/optional_files/get_sonobuoy.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+get_sono() {
+    has_bin wget
+    _arch
+    arch=$(if [ "$(uname -m)" = "x86_64" ]; then echo "amd64"; else echo "arm64"; fi)
+    wget https://github.com/vmware-tanzu/sonobuoy/releases/download/v0.57.1/sonobuoy_0.57.1_linux_"${arch}".tar.gz
```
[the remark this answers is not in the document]
VestigeJ: ah sorry this is an artifact from running commands - the real script does live in scripts - I kept the existing name mixedos_sonobuoy.sh to not interfere with the mixed os test yet. The script update does account for accepting a different version.

VestigeJ:
I'll be running golangci-lint run a lot more before creating PRs in the future.... Apologies

entrypoint/sonobuoyconformance/sonobuoyconformance_suite_test.go
```diff
@@ -0,0 +1,56 @@
+package sonobuoyconformance
```
mdrahman-suse, line 1: Can we rename the package to `conformance`, so suite will be `conformance_suite_test` and test file will be `sonobuoy_test.go`

entrypoint/sonobuoyconformance/my-sonobuoy-plugins
mdrahman-suse, line 1: Can we delete this file (after sonobouy gets deleted) or ignore committing to git?

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+
+func ConformanceTest(cluster *shared.Cluster) {
+	verifyClusterNodes(cluster)
+	installConformanceBinary()
+	launchSonobuoyTests("certified-conformance")
+	// refactor later to accept quick or certified-conformance
+	testResultTar := checkStatusGetResults(cluster)
+	shared.LogLevel("info", "%s", "testResultTar: "+testResultTar)
+	// need to do cilium force failures to test
+	rerunFailedTests(testResultTar)
```
mdrahman-suse, line 53: Can we add condition based on cilium?

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+	parseResults(testResultTar)
+	cleanupTests()
+}
+
+func verifyClusterNodes(cluster *shared.Cluster) bool {
```
mdrahman-suse, line 58: Should add this validation before the cluster is created, Can we add this in the suite?

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			GinkgoT().Errorf("error: %v", err)
+			return
+		}
+	}
+}
+
+func ConformanceTest(cluster *shared.Cluster) {
+	verifyClusterNodes(cluster)
+	installConformanceBinary()
```
mdrahman-suse, line 47: Can we add logs before each func call?

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+
+func parseResults(testResultTar string) {
+	cmd := "sonobuoy results  " + testResultTar
+	res, err := shared.RunCommandHost(cmd)
+	Expect(err).NotTo(HaveOccurred(), "failed cmd: "+cmd)
+	Expect(res).Should(ContainSubstring("Status: passed"))
+	shared.LogLevel("info", "%s", "sonobuoy results: "+res)
+}
+
+// func exportResultsToS3() {}
```
mdrahman-suse, line 127: Can you add TODO for this?

pkg/testcase/networkconnectivity.go
```diff
@@ -109,3 +110,65 @@ func testCrossNodeService(services, ports, expected []string) error {
 
 	return nil
 }
+
+func TestEndpointReadiness(cluster *shared.Cluster) {
+	var err error
+	var wg sync.WaitGroup
+	commands := []string{
```
mdrahman-suse, line 117: `commands` -> `endpoints`

pkg/testcase/networkconnectivity.go
```diff
@@ -109,3 +110,65 @@ func testCrossNodeService(services, ports, expected []string) error {
...
+
+	for _, serverIP := range servers {
+		for _, endpoint := range commands {
+			wg.Add(1)
+			go func(serverIP, endpoint string) {
+				defer wg.Done()
+				fmt.Println("Checking endpoint", endpoint, "on server", serverIP)
+				err = assert.CheckComponentCmdNode(endpoint, serverIP, "ok")
+				if err != nil {
+					fmt.Printf("Error checking endpoint %s on server %s: %v\n", endpoint, serverIP, err)
```
mdrahman-suse, line 145: `fmt.Println` -> `shared.LogLevel`

pkg/testcase/networkconnectivity.go
```diff
@@ -109,3 +110,65 @@ func testCrossNodeService(services, ports, expected []string) error {
...
+				}
+			}(serverIP.ExternalIP, endpoint)
+		}
+	}
+	wg.Wait()
+	Expect(err).NotTo(HaveOccurred(), err)
+}
+
+func Testk8sAPIReady(cluster *shared.Cluster) {
```
mdrahman-suse, line 154: can these be done using the k8s client?

pkg/testcase/networkconnectivity.go
```diff
@@ -109,3 +110,65 @@ func testCrossNodeService(services, ports, expected []string) error {
...
+
+func TestEndpointReadiness(cluster *shared.Cluster) {
+	var err error
+	var wg sync.WaitGroup
+	commands := []string{
+		"sudo curl -sk http://127.0.0.1:10248/healthz",  // kubelet
+		"sudo curl -sk http://127.0.0.1:10249/healthz",  // kube-proxy
+		"sudo curl -sk https://127.0.0.1:10257/healthz", // kube-controller
+		"sudo curl -sk https://127.0.0.1:10258/healthz", // cloud-controller
+		"sudo curl -sk https://127.0.0.1:10259/healthz", // kube-scheduler
```
mdrahman-suse, line 122: Are these static ports for the components for both the products?

pkg/testcase/networkconnectivity.go
```diff
@@ -109,3 +110,65 @@ func testCrossNodeService(services, ports, expected []string) error {
...
+		"sudo curl -sk  " + fmt.Sprintf("--cert /var/lib/rancher/%s/server/tls/client-ca.crt",
+			cluster.Config.Product) + fmt.Sprintf(" --key  /var/lib/rancher/%s/server/tls/client-ca.key",
+			cluster.Config.Product) + " https://127.0.0.1:6443/healthz",
```
mdrahman-suse, line 125: can we create vars here?
```
tlsPath := fmt.Sprintf("/var/lib/rancher/%s/server/tls/", cluster.Config.Product)

"sudo curl -sk  --cert " + tlsPath + "client-ca.crt" + 
" --key " + tlsPath + "client-ca.key" + 
" https://127.0.0.1:6443/healthz"
```

pkg/testcase/networkconnectivity.go
```diff
@@ -109,3 +110,65 @@ func testCrossNodeService(services, ports, expected []string) error {
...
+	}
+
+	servers, err := shared.GetNodesByRoles("control-plane")
```
mdrahman-suse, line 128: nit: `servers` -> `cpNodes`

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+
+func installConformanceBinary() {
+	sonobuoyVersion := customflag.ServiceFlag.External.SonobuoyVersion
+	err := shared.InstallSonobuoy("install", sonobuoyVersion)
+	Expect(err).NotTo(HaveOccurred())
+}
+
+func launchSonobuoyTests(testMode string) {
+	// not doing anything different yet if the status is running from the previous attempts
+	cmds := "kubectl get namespace sonobuoy --kubeconfig=" + shared.KubeConfigFile
```
mdrahman-suse, line 75: nit: use shorthand `namespace` -> `-n`
Not sure what the comment means, can you clarify?

pkg/testcase/networkconnectivity.go
```diff
@@ -109,3 +110,65 @@ func testCrossNodeService(services, ports, expected []string) error {
...
+	}
+	// I want the external IPs of the shared.GetNodesByRoles("control-plane") nodes
+	fmt.Println("Servers:", servers)
+
+	for _, serverIP := range servers {
+		for _, endpoint := range commands {
+			wg.Add(1)
+			go func(serverIP, endpoint string) {
+				defer wg.Done()
+				fmt.Println("Checking endpoint", endpoint, "on server", serverIP)
```
mdrahman-suse, line 142: https://github.com/rancher/distros-test-framework/pull/204/files#r1894404274

scripts/install_sonobuoy.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# Usage: sh mixedos_sonobuoy.sh runs with default values
```
mdrahman-suse, line 3: update as per script name

scripts/install_sonobuoy.sh
```diff
@@ -0,0 +1,56 @@
...
+# Usage: sh mixedos_sonobouy.sh delete
+
+action=${1:-install}
+version=${2:-0.57.2}
+arch=${3:-amd64}
+
+# adopt golang error handling in bash check variables are passed in appropriately - if not return appropriate error message
+
+installation(){
+    echo "Installing sonobuoy version ${version} for mixedos validation"
```
mdrahman-suse, line 14: can you separate this on another func that this is needed only for mixedos validation... may be add a parameter to be passed in

scripts/install_sonobuoy.sh
```diff
@@ -0,0 +1,56 @@
...
+    if [ ! -d "my-sonobuoy-plugins" ]; 
+    then
+        echo "Cloning repo: https://github.com/phillipsj/my-sonobuoy-plugins.git"
+        git clone https://github.com/phillipsj/my-sonobuoy-plugins.git
+    fi
+    wait
+    echo "Downloading sonobouy installer..."
+    if [[ $(command -v wget) ]]; then
+        wget -q https://github.com/vmware-tanzu/sonobuoy/releases/download/v"${version}"/sonobuoy_"${version}"_linux_"${arch}".tar.gz -O sonobuoy.tar.gz
```
mdrahman-suse, line 23: TODO: Add retry logic in case the wget fails

shared/cluster.go
```diff
@@ -250,15 +248,15 @@ func FetchIngressIP(namespace string) (ingressIPs []string, err error) {
 	return ingressIPs, nil
 }
 
-// SonobuoyMixedOS Executes scripts/mixedos_sonobuoy.sh script.
+// SonobuoyMixedOS Executes scripts/install_sonobuoy.sh script.
```
mdrahman-suse, line 251: Update as per func name

shared/cluster.go
```diff
@@ -688,3 +686,21 @@ func WaitForPodsRunning(defaultTime time.Duration, attempts uint) error {
...
 	)
 }
+func ExtractKubeImageVersion() string {
+	_, serverVersion, err := Product()
+	if err != nil {
+		LogLevel("warn", "error from RunCommandHost: %v\nwith res: Retrying...", err)
+		return "error"
+	}
+	re := regexp.MustCompile(`v(\d+\.\d+\.\d+)`)
+	fmt.Println("serverVersion: ", strings.Split(serverVersion, "\n"))
```
mdrahman-suse, line 696: https://github.com/rancher/distros-test-framework/pull/204/files#r1894404274

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func checkStatusGetResults(cluster *shared.Cluster) string {
+	// sonobuoy's output is becoming unreliable for status checks observe remaining count incorrect at 404
+	// 	sono status
+	//          PLUGIN     STATUS   RESULT   COUNT                                PROGRESS
+	//             e2e   complete   passed       1   Passed:  0, Failed:  0, Remaining:404
+	//    systemd-logs   complete   passed       2
+	// Sonobuoy has completed. Use `sonobuoy retrieve` to get results
+	cmd := "sonobuoy status --kubeconfig=" + shared.KubeConfigFile
```
mdrahman-suse, line 96: I think we can separate the status check and get results in to their own func so that it is simpler to understand the error returned in case of failures

mdrahman-suse:
> I'll be running golangci-lint run a lot more before creating PRs in the future.... Apologies

No worries, you can just run `make pre-commit` before pushing the changes

pkg/testcase/conformance.go
VestigeJ: I hear you but I disagree - over 95% of the time anyone reaches for sonobuoy within automation it's to run conformance, other test suites and sig-* groups should be run manually.

entrypoint/sonobuoyconformance/sonobuoyconformance_suite_test.go
VestigeJ: I disagree because conformance isn't isolated to only running sonobuoy there are multiple ways of running conformance tests that don't require sonobuoy.

entrypoint/sonobuoyconformance/my-sonobuoy-plugins
VestigeJ: Sorry which file are you referring to? The scripts are all part of the repo no?

pkg/testcase/conformance.go
VestigeJ: When writing kubectl commands on the cli I use the shorthand variants but for code clarity where it's less likely to be read unless someone is editing or attempting to understand a function I use the longform of kubectl commands.

The comment addresses that this isn't doing different functionality if a re-used cluster is already running sonobuoy tests. This could target future functionality that works with previously run conformacne tests on a cluster but either is checking different results, re-running different suites, etc.

pkg/testcase/conformance.go
VestigeJ: Done will be in my next push

pkg/testcase/networkconnectivity.go
VestigeJ: Yes for default configurations unless of course someone intentionally alters them with configuration values

scripts/install_sonobuoy.sh
VestigeJ: I'm tempted to leave it for now because of how we can refactor the mixed os validation later on. In truth the mixed os validation is just creating a workload for a windows node and then creating an additional workload for a linux node. Under the hood this test is a simple workload deploy with node affinity labels to get each node type. the plugin for sonobuoy here is obscuring that it's pulling down separate images/workloads for each node type.

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+	shared.LogLevel("info", "%s", "testResultTar: "+testResultTar)
+	// need to do cilium force failures to test
+	rerunFailedTests(testResultTar)
+	parseResults(testResultTar)
+	cleanupTests()
+}
+
+func verifyClusterNodes(cluster *shared.Cluster) bool {
+	if cluster.NumAgents < 1 && cluster.NumServers < 1 {
+		shared.LogLevel("error", "%s", "cluster does not meet the minimum requirements to run conformance tests")
```
ShylajaDevadiga, line 60: Can we add One agent One server to the comment to display minimum requirement?

scripts/install_sonobuoy.sh
VestigeJ: Done got some retry logic written will push shortly

pkg/testcase/cluster.go
```diff
@@ -49,37 +47,14 @@ func TestBuildCluster(cluster *shared.Cluster) {
 	}
 }
 
-// TestSonobuoyMixedOS runs sonobuoy tests for mixed os cluster (linux + windows) node.
-func TestSonobuoyMixedOS(deleteWorkload bool) {
-	sonobuoyVersion := customflag.ServiceFlag.External.SonobuoyVersion
-	err := shared.SonobuoyMixedOS("install", sonobuoyVersion)
+// TestDisplayClusterDetails used to display cluster details.
+func TestDisplayClusterDetails() {
```
ShylajaDevadiga, line 51: Are we not using this here? https://github.com/rancher/distros-test-framework/blob/main/pkg/template/helper.go#L79

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,137 @@
...
+		return false
+	}
+
+	return true
+}
+
+func installConformanceBinary() {
+	sonobuoyVersion := customflag.ServiceFlag.External.SonobuoyVersion
```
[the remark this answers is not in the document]
VestigeJ: Yeah I actually think we should get rid of passing the sonobuoy version as it doesn't change often and should always just be kept latest especially for conformance tests.

shared/cluster.go
VestigeJ: Voting we keep the script named install_sonobuoy.sh as it's descriptive of what it's doing.

entrypoint/sonobuoyconformance/sonobuoyconformance_suite_test.go
VestigeJ: done - good idea thanks for explaining

pkg/testcase/conformance.go
VestigeJ: done

pkg/testcase/cluster.go
VestigeJ: This method is still exposed publicly from the conformance file

pkg/testcase/conformance.go
VestigeJ: testing this refactor locally before pushing it

pkg/testcase/conformance.go
mdrahman-suse: As we have linter for word length and also for consistency, lets keep using the shorthand as we have used in other places

scripts/install_sonobuoy.sh
mdrahman-suse: The reason I asked to have it separate so that we dont install / download unnecessary stuff for the regular sonobuoy tests and only install what is needed by the respective use cases...

shared/cluster.go
mdrahman-suse: I meant changing `SonobuoyMixedOS` -> `InstallSonobuoy` in this line

VestigeJ:
Just posting for visibility with all of the refactoring that's been going on 

```
PASSED! Sonobuoy Conformance Tests... Starts Up with no issues

•INFO[2025-01-09T12:00:54-08:00] 

Cluster nodes:                             
ip.us-east-2.compute.internal   Ready   control-plane,etcd,master   3m26s   v1.31.4+k3s1   1.1.1.1     1.1.1.1    SUSE Linux Enterprise Server 15 SP5   5.4.3.21-150500.53-default   containerd://1.7.23-k3s2
ip.us-east-2.compute.internal    Ready   <none>                      28s     v1.31.4+k3s1  1.1.1.3     1.1.1.3    SUSE Linux Enterprise Server 15 SP5   5.4.3.21-150500.53-default   containerd://1.7.23-k3s2


PASSED! Sonobuoy Conformance Tests... Validates Node

•INFO[2025-01-09T12:00:55-08:00] 

Cluster pods:                              
kube-system   coredns-ccb96694c-kfbfd                   1/1   Running     0     3m21s   10.42.0.4   ip.us-east-2.compute.internal   <none>   <none>
kube-system   helm-install-traefik-crd-2h64b            0/1   Completed   0     3m21s   10.42.0.3   ip.us-east-2.compute.internal   <none>   <none>
kube-system   helm-install-traefik-s9kts                0/1   Completed   1     3m21s   10.42.0.6   ip.us-east-2.compute.internal   <none>   <none>
kube-system   local-path-provisioner-5cf85fd84d-wzqjh   1/1   Running     0     3m21s   10.42.0.5   ip.us-east-2.compute.internal   <none>   <none>
kube-system   metrics-server-5985cbc9d7-ptd6d           1/1   Running     0     3m21s   10.42.0.2   ip.us-east-2.compute.internal   <none>   <none>
kube-system   svclb-traefik-6653f630-5ppg5              2/2   Running     0     28s     10.42.1.2   ip.us-east-2.compute.internal    <none>   <none>
kube-system   svclb-traefik-6653f630-hp89x              2/2   Running     0     2m53s   10.42.0.7   ip.us-east-2.compute.internal   <none>   <none>
kube-system   traefik-57b79cf995-bvclj                  1/1   Running     0     2m53s   10.42.0.8   ip.us-east-2.compute.internal   <none>   <none>


PASSED! Sonobuoy Conformance Tests... Validate Pods

•INFO[2025-01-09T12:00:55-08:00] installing sonobuoy binary                   
INFO[2025-01-09T12:01:06-08:00] checking namespace existence                 
serverVersion:  [k3s version v1.31.4+k3s1 (a562d090) go version go1.22.9]
re: v(\d+\.\d+\.\d+)
match:  [v1.31.4 1.31.4]
serverVersionReturnValue :  v1.31.4
INFO[2025-01-09T12:01:22-08:00] checking status of running tests             
INFO[2025-01-09T12:02:54-08:00] getting sonobuoy results                     
INFO[2025-01-09T12:02:56-08:00] testResultTar: 202501092001_sonobuoy_44454f32-35b9-46e0-97dd-32005ffba79d.tar.gz 
INFO[2025-01-09T12:02:56-08:00] re-running tests that failed from previous run 
INFO[2025-01-09T12:02:56-08:00] parsing sonobuoy results                     
INFO[2025-01-09T12:02:56-08:00] sonobuoy results: Plugin: e2e
Status: passed
Total: 6609
Passed: 5
Failed: 0
Skipped: 6604

Plugin: systemd-logs
Status: passed
Total: 2
Passed: 2
Failed: 0
Skipped: 0

Run Details:
API Server version: v1.31.4+k3s1
Node health: 2/2 (100%)
Pods health: 12/12 (100%)
Errors detected in files:
Errors:
10 podlogs/kube-system/metrics-server-5985cbc9d7-ptd6d/logs/metrics-server.txt
 1 podlogs/sonobuoy/sonobuoy-e2e-job-40663c74973f4742/logs/e2e.txt
 1 podlogs/kube-system/helm-install-traefik-crd-2h64b/logs/helm.txt
 1 podlogs/kube-system/helm-install-traefik-s9kts/logs/helm.txt
Warnings:
1 podlogs/sonobuoy/sonobuoy-e2e-job-40663c74973f4742/logs/e2e.txt
1 podlogs/sonobuoy/sonobuoy/logs/kube-sonobuoy.txt 
INFO[2025-01-09T12:02:56-08:00] cleaning up cluster conformance tests and deleting sonobuoy namespace 

PASSED! Sonobuoy Conformance Tests... Validates the releases conformance with upstream requirements

•

Ran 4 of 4 Specs in 139.027 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestConformance (139.03s)
PASS
ok      github.com/rancher/distros-test-framework/entrypoint/conformance        525.544s

```

fmoral2:
can u add a section of this, on how to use or any important infos under /docs/testing.md ?

entrypoint/conformance/conformance_suite_test.go
```diff
@@ -0,0 +1,68 @@
...
+	os.Exit(m.Run())
+}
+
+func TestConformance(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecs(t, "Run Conformance Suite")
+}
+
+var _ = AfterSuite(func() {
```
fmoral2, line 52: this test will have a qase automatic report? if yes, we need to add in here like other suits

entrypoint/conformance/conformance_suite_test.go
```diff
@@ -0,0 +1,68 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(status).To(Equal("cluster destroyed"))
+	}
+})
+
+func verifyClusterNodes(cluster *shared.Cluster) bool {
+	shared.LogLevel("info", "verying cluster configuration matches minimum requirements for conformance tests")
+	if cluster.NumAgents < 1 && cluster.NumServers < 1 {
+		shared.LogLevel("error", "%s", "cluster must at least consist of 1 server and 1 agent")
+		return false
```
fmoral2, line 64: here we probably want to kill everything if we are not passing the min right ?

so use `os.Exit(1)` instead of bool

entrypoint/conformance/conformance_suite_test.go
```diff
@@ -0,0 +1,68 @@
...
+	kubeconfig = os.Getenv("KUBE_CONFIG")
+	if kubeconfig == "" {
+		// gets a cluster from terraform.
+		cluster = shared.ClusterConfig()
+	} else {
+		// gets a cluster from kubeconfig.
+		cluster = shared.KubeConfigCluster(kubeconfig)
+	}
+
+	verifyClusterNodes(cluster)
```
fmoral2, line 41: we need to check that, before cluster creation.

so this function call needs to go right after `add.env`

entrypoint/validatecluster/validatecluster_test.go
```diff
@@ -38,6 +38,10 @@ var _ = Describe("Test:", func() {
 		testcase.TestServiceNodePort(true, true)
 	})
 
+	It("Checks the healthz endpoints", func() {
+		testcase.TestEndpointReadiness(cluster)
```
fmoral2, line 42: why adding this here?

pkg/testcase/cluster.go
```diff
@@ -49,40 +47,14 @@ func TestBuildCluster(cluster *shared.Cluster) {
 	}
 }
 
-// TestSonobuoyMixedOS runs sonobuoy tests for mixed os cluster (linux + windows) node.
-func TestSonobuoyMixedOS(deleteWorkload bool) {
-	sonobuoyVersion := customflag.ServiceFlag.External.SonobuoyVersion
-	err := shared.SonobuoyMixedOS("install", sonobuoyVersion)
+func TestDisplayClusterDetails() {
```
fmoral2, line 50: canu delete this function please? its not being used anywhere

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+
+func installConformanceBinary() {
+	shared.LogLevel("info", "installing sonobuoy binary")
+	err := shared.InstallSonobuoy("install")
+	Expect(err).NotTo(HaveOccurred())
+}
+
+func launchSonobuoyTests(testMode string) {
+	shared.LogLevel("info", "checking namespace existence")
+	// not doing anything different yet if the status is running from the previous attempts
```
fmoral2, line 65: what does exactly this comment means?

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return "timed out waiting for sonobuoy to complete after 170 minutes nearly 3 hours"
+}
+
+func getResults(cluster *shared.Cluster) string {
+	shared.LogLevel("info", "getting sonobuoy results")
+	cmd := "sonobuoy retrieve --kubeconfig=" + shared.KubeConfigFile
+	res, err := shared.RunCommandHost(cmd)
+	Expect(err).NotTo(HaveOccurred())
+	// sonobuoy's output is becoming unreliable for status checks observe remaining count incorrect at 404
```
fmoral2, line 97: add those type of info on docs please

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+	Eventually(func() string {
+		res, err := shared.RunCommandHost(cmd)
+		Expect(err).NotTo(HaveOccurred())
+		return res
+	}, "170m", "45s").Should(ContainSubstring("Sonobuoy has completed"))
+
+	return "timed out waiting for sonobuoy to complete after 170 minutes nearly 3 hours"
+}
+
+func getResults(cluster *shared.Cluster) string {
```
fmoral2, line 92: cluster parameter is not being used

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func checkStatus() string {
+	shared.LogLevel("info", "checking status of running tests")
+	cmd := "sonobuoy status --kubeconfig=" + shared.KubeConfigFile
+	Eventually(func() string {
+		res, err := shared.RunCommandHost(cmd)
+		Expect(err).NotTo(HaveOccurred())
+		return res
+	}, "170m", "45s").Should(ContainSubstring("Sonobuoy has completed"))
```
fmoral2, line 87: shouldnt we increase a bit this interval of "45s" ? cause this is running every 45s but we do know that at least for 40 min we dont need to check right, maybe we should improve this logic

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if strings.Contains(res, "Error from server (NotFound): namespaces \"sonobuoy\" not found") {
+		cmd := "sonobuoy run --kubeconfig=" + shared.KubeConfigFile +
+			" --mode=" + testMode + " --kubernetes-version=" + shared.ExtractKubeImageVersion()
+		_, err := shared.RunCommandHost(cmd)
+		Expect(err).NotTo(HaveOccurred())
+	}
+}
+
+func checkStatus() string {
```
fmoral2, line 80: you are returning a string but never using on the function call 

you can add your error message
`"timed out waiting for sonobuoy to complete after 170 minutes nearly 3 hours"` on the Eventually itself soemthing like
```
	Eventually(func() string {
		res, err := shared.RunCommandHost(cmd)
		return res

	}, "170m", "45s").Should(ContainSubstring("Sonobuoy has completed"),"timed out waiting for sonobuoy to complete after 170 minutes nearly 3 hours"

}
```

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+func ConformanceTest(cluster *shared.Cluster, testName string) {
+	installConformanceBinary()
+	// refactor later to accept quick or certified-conformance
+	launchSonobuoyTests(testName)
+	checkStatus()
+	testResultTar := getResults(cluster)
+	shared.LogLevel("info", "%s", "testResultTar: "+testResultTar)
+	// TODO(VestigeJ): 2021-08-26
+	// need to do cilium force failures to test
+	rerunFailedTests(testResultTar)
```
fmoral2, line 52: every running we are waiting to have the need to re-run failed ones?

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+
+func rerunFailedTests(testResultTar string) {
+	shared.LogLevel("info", "re-running tests that failed from previous run")
+	cmd := "sonobuoy run --rerun-failed=" + testResultTar + " --kubeconfig=" + shared.KubeConfigFile
+	res, err := shared.RunCommandHost(cmd)
+	Expect(err).To(HaveOccurred(), "failed cmd: "+cmd)
+	Expect(res).Should(ContainSubstring("no tests failed for plugin"))
+}
+
+func parseResults(testResultTar string) {
```
fmoral2, line 115: https://github.com/rancher/distros-test-framework/pull/204/files#r1913267227

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+	shared.LogLevel("info", "%s", "sonobuoy results: "+res)
+}
+
+// TODO(VestigeJ): 2021-08-26
+// func exportResultsToS3() {}
+// export results to s3
+// if destroy is false keep results in s3 bucket
+// send results to s3 bucket with deletion rules
+
+func cleanupTests() {
```
fmoral2, line 130: https://github.com/rancher/distros-test-framework/pull/204/files#r1913267227

scripts/install_sonobuoy.sh
```diff
@@ -0,0 +1,74 @@
...
+# Usage: sh install_sonobouy.sh install 0.56.17 arm64
+# Usage: sh install_sonobouy.sh delete
+
+action=${1:-install}
+version=${2:-0.57.2}
+arch=${3:-amd64}
+mixed_plugins_url="git clone https://github.com/phillipsj/my-sonobuoy-plugins.git"
+sonobuoy_url="https://github.com/vmware-tanzu/sonobuoy/releases/download/v${version}/sonobuoy_${version}_linux_${arch}.tar.gz"
+max_retries=5
+retry_delay=11
```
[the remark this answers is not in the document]
fmoral2: same for `max_retries`

scripts/install_sonobuoy.sh
```diff
@@ -0,0 +1,74 @@
...
+# adopt golang error handling in bash check variables are passed in appropriately - if not return appropriate error message
+
+download_retry(){
+  i=1
+ 
+  until $1 || [ $i -gt $max_retries ]; do
```
fmoral2, line 19: while loop does not make more sense/clarity here ?
also can we reassign the $1 to a named var here? for better understanding

scripts/install_sonobuoy.sh
```diff
@@ -0,0 +1,74 @@
...
+        download_retry "${mixed_plugins_url}"
+    fi
+    wait
+    echo "Downloading sonobouy installer..."
+    if [[ $(command -v wget) ]]; then
+        download_retry "wget -q ${sonobuoy_url} -O sonobuoy.tar.gz"
+    elif [[ $(command -v curl) ]]; then
+        download_retry "curl -s ${sonobuoy_url} --output sonobuoy.tar.gz"
+        wait
+        sleep 10
```
fmoral2, line 45: using curl we need to wait and sleep and using wget no need?

scripts/install_sonobuoy.sh
```diff
@@ -0,0 +1,74 @@
...
+    else
+        echo "Unable to use wget or curl to download sonobuoy installer, consider a networking error or an under configured OS if this error persists"
+        wait
```
fmoral2, line 48: why waiting and sleep after an error ?

entrypoint/validatecluster/validatecluster_test.go
VestigeJ: This was prior work for checking the cluster endpoints were reporting as operational - it doesn't need to be in this PR but I was trying to get it in finally.

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+	cleanupTests()
+}
+
+func installConformanceBinary() {
+	shared.LogLevel("info", "installing sonobuoy binary")
+	err := shared.InstallSonobuoy("install")
+	Expect(err).NotTo(HaveOccurred())
+}
+
+func launchSonobuoyTests(testMode string) {
```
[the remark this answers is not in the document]
VestigeJ: The second half of this function is launching the sonobuoy conformance checks - the first half is checking if the sonobuoy namespace exists or is already running on a given cluster.

pkg/testcase/conformance.go
VestigeJ: Removed note

pkg/testcase/conformance.go
VestigeJ: I could increase it to every 10 minutes, you don't want the cluster waiting around for the check to complete any longer than necessary which was why i left the check interval 45 seconds. So we would know it was completed sooner after it finished instead of waiting for a ten minute internal to see that it completed.

pkg/testcase/conformance.go
VestigeJ: It is not a long wait to do this and this automatically resolves any flakey tests except for the cilium tests which are known/expected conformance failures at this time. It makes more sense to have sonobuoy retry any tests that do fail which someone would have to do anyways in response to conformance tests failing. This way just ensures that whomever is responding to failed conformance tests can know they were already retried in case they were flakey.

pkg/testcase/networkconnectivity.go
VestigeJ: Pushing that out of scope for now

scripts/install_sonobuoy.sh
VestigeJ: $1 is the local boolean of the until loop and is shellcheck compliant.

scripts/install_sonobuoy.sh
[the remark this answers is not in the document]
VestigeJ: The variable locations here are for ease of edits in the future if necessary they're all shellcheck compliant norms.

scripts/install_sonobuoy.sh
VestigeJ: The script isn't downloading the mixedos plugins unless the test suite passes it so there isn't additional items downloaded that aren't necessary. If it's conformance only sonobuoy is installed - if it's mixedos both sonobuoy and the mixedos plugin are required and installed.

scripts/install_sonobuoy.sh
VestigeJ: They _**_**should**_**_ function identically on all systems but we can try just wait which will finish when the command process finishes.

VestigeJ:
RKE2 with Cilium 

```

File: pkg/testcase/conformance.go (hunk 35 above)
Comment (fmoral2) [reply pkg/testcase/conformance.go]: totally agree, thats why said we should improve the logic. both scenarios are not good
checking every 10m or every 45s.

we should add a between , like always at first wait longer ( since it will never finishes before 30 min , example ) and then after a good amount of time , check less bigger intervals etc

File: pkg/testcase/conformance.go (hunk 37 above)
Comment (fmoral2) [reply pkg/testcase/conformance.go:51]: yeah sure , but cant we just simply on the `get results()` parse the `res` check if there is  failed ones there and if yes, then we call rerun ?

File: pkg/testcase/networkconnectivity.go [hunk 45]
```diff
@@ -109,3 +110,65 @@ func testCrossNodeService(services, ports, expected []string) error {
...
+		"sudo curl -sk http://127.0.0.1:10249/healthz",  // kube-proxy
+		"sudo curl -sk https://127.0.0.1:10257/healthz", // kube-controller
+		"sudo curl -sk https://127.0.0.1:10258/healthz", // cloud-controller
+		"sudo curl -sk https://127.0.0.1:10259/healthz", // kube-scheduler
+		"sudo curl -sk  " + listenPort + " https://127.0.0.1:6443/healthz",
+	}
+
+	controlPlaneNodes, err := shared.GetNodesByRoles("control-plane")
+	if err != nil {
+		// handle the error, e.g. log or return an error
```
fmoral2, line 132: dont forget to fix what your comment is saying

pkg/testcase/networkconnectivity.go
```diff
@@ -109,3 +110,65 @@ func testCrossNodeService(services, ports, expected []string) error {
...
+		err := assert.CheckComponentCmdNode(
+			"kubectl get --raw='http://127.0.0.1/readyz?verbose'",
+			serverIP,
+			"readyz check passed",
+		)
+		Expect(err).NotTo(HaveOccurred(), err)
+	}
+}
+
+func Testk8sAPILive(cluster *shared.Cluster) {
```
fmoral2, line 165: if not using those funcs anywhere, remove for now

entrypoint/validatecluster/validatecluster_test.go
fmoral2: sure thats awesome, but whats the point checking those in a running cluster that is probably all working fine since on this suite before this call we are just validating cluster ip and nodeport ? 

+ should we also already add etcd endpoint health? 
+ this function i would also suggest to accept individual endpoints like

i wanna call the function to check all or just 1 or 2

`func TestEndpointReadiness(cluster *shared.Cluster, endpoints ...string)`
and then instead a simple list with all, create a map to return all endpoint or only the choosen ones

pkg/testcase/conformance.go
fmoral2: yeah sure , but cant we just simply on the `get results()` parse the `res` check if there is  failed ones there and if yes, then we call rerun ?

entrypoint/validatecluster/validatecluster_test.go
fmoral2: and if going on that route, this also could be a aux or cluster function , instead a "Test"

VestigeJ: I'll probably stash this just so conformance can proceed

pkg/testcase/conformance.go
VestigeJ: It's been expanded to 10 minutes which would only result in 17 checks for now and still resolve results in a reasonable amount of time for now.

pkg/testcase/conformance.go
VestigeJ: This essentially performs identically to what you're recommending although not in the same steps, instead of handling the strings emitted by the results sonobuoy itself already parses those results which would add code complexity. I'm voting we leave the logic in sonobuoy for this type of conformance test as it's likely more conclusive.

pkg/testcase/networkconnectivity.go
VestigeJ: Will be stashing network checks to get conformance in.

pkg/testcase/networkconnectivity.go
VestigeJ: Will be stashing network checks to get conformance in.

.gitignore
```diff
@@ -11,3 +11,5 @@ workloads/*/dynamic-ingressroute.yaml
 /config/*.env.*
 /config/*.tfvars
 tmp/
+entrypoint/conformance/my-sonobuoy-plugins
+entrypoint/conformance/*.tar.gz
```
mdrahman-suse, line 15: nit: new line

pkg/testcase/conformance.go
VestigeJ: In this case the shorthand would be kubectl get ns sonobuoy because I'm getting the resource 'namespace' not resources within -n sonobuoy. I'd like to leave it considering it's still under the line length limit from the linter.

pkg/testcase/conformance.go
fmoral2: sorry i am not understanding.....

what am suggesting is only  re-run if actually find failed ones or in cases u are already expecting faillures ( e.g cilium )

for instance, you are returning on line 106 if it is cilium right?  why are you waiting a function call `rerunFailedTests()` go inside it, just to check if it is cilium and if yes , return back? 

makes no much sense u get it?

u are always calling re-run but in the real usage it will be only few times needed , i get the idea "oh but if there is a failure ( which if some frequence occur ) we are already taking care of that in the logic ( because u are always re-running, execpt for cilium) 

so my suggestion is , your code should do what needs to do when needs to

```

getResults() ---  checkRessults() --- failed ones ?  yes, its cilium ? dont re-run  
```
```

getResults() ---  checkRessults() --- failed ones ?   yes ,   its not cilium ? re-run 
```
```
getResults() ---  checkRessults() --- failed ones ? no, dont re-run 
```


- this way your code is clear, dont spent unnecessary time and resources and its more maintainable

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,136 @@
...
+func getResults() string {
+	shared.LogLevel("info", "getting sonobuoy results")
+	cmd := "sonobuoy retrieve --kubeconfig=" + shared.KubeConfigFile
+	res, err := shared.RunCommandHost(cmd)
+	Expect(err).NotTo(HaveOccurred())
+
+	return res
+}
+
+func rerunFailedTests(testResultTar string) {
```
fmoral2, line 97: do u mind separating code blocks?
 soemthing like

```
func rerunFailedTests(testResultTar string) {
	ciliumExpectedFailures := `
		[sig-network] Services should serve endpoints on same port and different protocols
	 	Services should be able to switch session affinity for service with type clusterIP
		Services should have session affinity work for service with type clusterIP`

	if !strings.Contains(os.Getenv("cni"), "cilium") {
		shared.LogLevel("info", "Cilium has known issues with conformance tests, skipping re-run")
		shared.LogLevel("info", "ciliumExpectedFailures: %s", ciliumExpectedFailures)

		return
	}

	shared.LogLevel("info", "re-running tests that failed from previous run")

	cmd := "sonobuoy run --rerun-failed=" + testResultTar + " --kubeconfig=" + shared.KubeConfigFile
	res, err := shared.RunCommandHost(cmd)
	Expect(err).To(HaveOccurred(), "failed cmd: "+cmd)
	Expect(res).Should(ContainSubstring("no tests failed for plugin"))
}


```

fmoral2: on the function bellow as well please, is really hard to read code like that

scripts/install_sonobuoy.sh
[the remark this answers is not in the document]
fmoral2: 1- if it should be " edited" it should be parametrized right ?  not manually edit in the script

> 2- "they're all shellcheck compliant norms."
thats not my point, i am saying in a reading/coding perspective.

shared/cluster.go
```diff
@@ -688,3 +686,23 @@ func WaitForPodsRunning(defaultTime time.Duration, attempts uint) error {
...
+	_, serverVersion, err := Product()
+	if err != nil {
+		LogLevel("error", "error from RunCommandHost: %v\nwith res: Retrying...", err)
+		os.Exit(1)
+	}
+	re := regexp.MustCompile(`v(\d+\.\d+\.\d+)`)
+	fmt.Println("serverVersion: ", strings.Split(serverVersion, "\n"))
+	fmt.Println("re:", re)
+	match := re.FindStringSubmatch(serverVersion)
+	fmt.Println("match: ", match)
```
fmoral2, line 700: if you want to print use logs please in all those println

shared/cluster.go
```diff
@@ -688,3 +686,23 @@ func WaitForPodsRunning(defaultTime time.Duration, attempts uint) error {
...
+	if match == nil {
+		LogLevel("error", "%s failed to resolve to server version string: %s", serverVersion, err)
+		os.Exit(1)
```
fmoral2, line 703: you cant return os.exit() here because it will execution

shared/cluster.go
```diff
@@ -688,3 +686,23 @@ func WaitForPodsRunning(defaultTime time.Duration, attempts uint) error {
 		}),
 	)
 }
+
+func ExtractKubeImageVersion() string {
+	_, serverVersion, err := Product()
+	if err != nil {
+		LogLevel("error", "error from RunCommandHost: %v\nwith res: Retrying...", err)
+		os.Exit(1)
```
fmoral2, line 694: you cant return os.exit() here because it will execution

entrypoint/conformance/sonobuoy_test.go
```diff
@@ -0,0 +1,45 @@
...
+	It("Validate Pods", func() {
+		testcase.TestPodStatus(
+			cluster,
+			assert.PodAssertRestart(),
+			assert.PodAssertReady())
+	})
+
+	It("Validates the releases conformance with upstream requirements", func() {
+		testcase.ConformanceTest(cluster, "certified-conformance")
+		// testcase.ConformanceTest(cluster, "quick")
```
mdrahman-suse, line 35: comment

entrypoint/conformance/conformance_suite_test.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		shared.LogLevel("error", "error adding env vars: %w\n", err)
+		os.Exit(1)
+	}
+	verifyClusterNodes()
+
+	kubeconfig = os.Getenv("KUBE_CONFIG")
+	if kubeconfig == "" {
+		cluster = shared.ClusterConfig()
+	}
```
mdrahman-suse, line 37: if kubeconfig is not empty then we should use 
```
cluster = shared.KubeConfigCluster(kubeconfig)
```

entrypoint/conformance/conformance_suite_test.go
```diff
@@ -0,0 +1,77 @@
...
+	os.Exit(m.Run())
+	shared.LogLevel("error", "cluster must at least consist of 1 server and 1 agent")
```
mdrahman-suse, line 39: this should go inside `verifyClusterNodes()` func

shared/cluster.go
VestigeJ: When I'm unable to get the server image version I do want to exit the run because the entire suite won't run correctly. I could use a little more clarification on your comment.

entrypoint/conformance/conformance_suite_test.go
```diff
@@ -0,0 +1,100 @@
...
+		shared.LogLevel("error", "error converting no_of_agent_nodes to int: %w", err)
+		os.Exit(1)
+	}
+
+	if serverNum < 1 && agentNum < 1 {
+		shared.LogLevel("error", "%s", "cluster must at least consist of 1 server and 1 agent")
+		os.Exit(1)
+	}
+
+	return true
```
mdrahman-suse, line 99: return is not needed as its not used

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,138 @@
...
+		Expect(err).NotTo(HaveOccurred(), "failed cmd: "+cmd)
+		err = shared.InstallSonobuoy("delete")
+		if err != nil {
+			GinkgoT().Errorf("error: %v", err)
+			return
+		}
+	}
+}
+
+func ConformanceTest(_ *shared.Cluster, testName string) {
```
mdrahman-suse, line 44: suggest removing this `_ *shared.Cluster` if its not being used anywhere in this func

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,138 @@
...
+	Expect(err).NotTo(HaveOccurred(), "failed output: "+res)
+
+	cmd = "sonobuoy retrieve --kubeconfig=" + shared.KubeConfigFile
+	testResultTar, err := shared.RunCommandHost(cmd)
+	Expect(err).NotTo(HaveOccurred(), "failed cmd: "+cmd)
+
+	cmd = "sonobuoy results  " + testResultTar
+	res, err = shared.RunCommandHost(cmd)
+	Expect(err).NotTo(HaveOccurred(), "failed cmd: "+cmd)
+	Expect(res).Should(ContainSubstring("Plugin: mixed-workload-e2e\nStatus: passed\n"))
```
mdrahman-suse, line 30: Can probably re-use `getResults` and `parseResults` func here

mdrahman-suse (commented):
Added some comments, else LGTM

pkg/testcase/conformance.go
```diff
@@ -0,0 +1,138 @@
...
+	if deleteWorkload {
+		cmd = "sonobuoy delete --all --wait --kubeconfig=" + shared.KubeConfigFile
+		_, err = shared.RunCommandHost(cmd)
+		Expect(err).NotTo(HaveOccurred(), "failed cmd: "+cmd)
+		err = shared.InstallSonobuoy("delete")
+		if err != nil {
+			GinkgoT().Errorf("error: %v", err)
+			return
+		}
+	}
```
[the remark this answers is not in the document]
VestigeJ: Initially some months ago I specifically didn't want to change the existing mixedOS test functionality, but good call out now.

entrypoint/conformance/conformance_suite_test.go
```diff
@@ -0,0 +1,99 @@
...
+	if kubeconfig == "" {
+		// gets a cluster from terraform.
+		cluster = shared.ClusterConfig(cfg)
+	} else {
+		// gets a cluster from kubeconfig.
+		cluster = shared.KubeConfigCluster(kubeconfig)
+	}
+
+	os.Exit(m.Run())
+	os.Exit(1)
```
fmoral2, line 50: this code line  is unreachable

PR #207, opened 2025-01-09
Fix/upgrade manual

mdrahman-suse:
#### Testing ####

Checklist:


pkg/testcase/upgrademanual.go
```diff
@@ -22,66 +28,92 @@ func TestUpgradeClusterManual(cluster *shared.Cluster, k8sClient *k8s.Client, ve
...
+			{
+				Service:  product,
+				Action:   "restart",
+				NodeType: nodeType,
+			},
+		}
+		output, err := ms.ManageService(ip, restart)
+		if output != "" {
+			Expect(output).To(ContainSubstring("active "),
+				fmt.Sprintf("error starting %s service for %s node ip: %s", product, nodeType, ip))
```
fmoral2, line 92: starting or restarting?

pkg/testcase/upgrademanual.go
```diff
@@ -22,66 +28,92 @@ func TestUpgradeClusterManual(cluster *shared.Cluster, k8sClient *k8s.Client, ve
...
-			shared.LogLevel("info", "Restarting %s: %s", nodeType, ip)
-			err := shared.RestartCluster(cluster.Config.Product, ip)
-			if err != nil {
-				return
-			}
-		}(ip, upgradeCommand)
+		output, err = ms.ManageService(ip, status)
+		if output != "" {
+			Expect(output).To(ContainSubstring("active "),
+				fmt.Sprintf("error checking status %s service for %s node ip: %s", product, nodeType, ip))
```
VestigeJ, line 102:
```
shared.LogLevel("error", "") // ?
```

pkg/testcase/upgrademanual.go
```diff
@@ -22,66 +28,92 @@ func TestUpgradeClusterManual(cluster *shared.Cluster, k8sClient *k8s.Client, ve
...
-	if !ok {
-		return errors.New("cluster is not healthy")
+	if product == "k3s" {
+		shared.LogLevel("info", "Waiting for 1 mins after installing upgrade...")
+		time.Sleep(1 * time.Minute)
+		ms := shared.NewManageService(3, 1)
+		output, err := ms.ManageService(ip, status)
+		if output != "" {
+			Expect(output).To(ContainSubstring("active "),
+				fmt.Sprintf("error starting %s service for %s node ip: %s", product, nodeType, ip))
```
VestigeJ, line 114:
```
shared.LogLevel("error", "") // ?
```

pkg/testcase/upgrademanual.go
```diff
@@ -22,66 +28,92 @@ func TestUpgradeClusterManual(cluster *shared.Cluster, k8sClient *k8s.Client, ve
...
+		}
+		Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("error starting %s service on %s", product, ip))
```
VestigeJ, line 116:
```
shared.LogLevel("error", "") // ?
```

pkg/testcase/daemonset.go
```diff
@@ -17,6 +18,12 @@ func TestDaemonset(applyWorkload, deleteWorkload bool) {
 		workloadErr = shared.ManageWorkload("apply", "daemonset.yaml")
 		Expect(workloadErr).NotTo(HaveOccurred(), "Daemonset manifest not deployed")
 	}
+
+	getDeamonset := "kubectl get pods -n test-daemonset" +
+		" --field-selector=status.phase=Running --kubeconfig="
```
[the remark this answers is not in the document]
mdrahman-suse: It is being passed on the next line

pkg/testcase/upgrademanual.go
mdrahman-suse: should `running`, fixing it

pkg/testcase/upgrademanual.go
mdrahman-suse: Added expect here so no need to log, will update the expect message

pkg/testcase/upgrademanual.go
mdrahman-suse: https://github.com/rancher/distros-test-framework/pull/207#discussion_r1937980446

shared/manageservice.go
```diff
@@ -35,9 +36,14 @@ func NewManageService(maxRetries uint, retryDelay time.Duration) *ManageService
 //
 // If action is "rotate", it will rotate the certificate for the given service.
 // If action is "status", it will return the status of the service.
-func (ms *ManageService) ManageService(ip string, actions []ServiceAction) (string, error) {
+func (ms *ManageService) ManageService(ip string, actions []ServiceAction, explicitDelay time.Duration) (string, error) {
 	for _, act := range actions {
-		LogLevel("debug", "Starting %s on %s@%s", act.Action, act.Service, ip)
+		if explicitDelay > 0 {
```
fmoral2, line 41: hey, as i mentioned yesterday, this should be handle at `ServiceAction` level not here....because if you do this here, where i am going to use like: 

 
 lets say i am in another test that needs this explicit for upgrade but then for other usages that i might have i dont need to, but even when i dont need to , since at the creation ( manage service level ) i passed this parameter as not null or 0 , for all actions i perform later this will sleep before runs, and i might not want to


so that being said, if you add this on the ServiceAction level, we can for each action we perform , explicit wait or not , not forcing all actions to wait for this



- would also suggest to use `	<-time.After(ms.ExplicityDelay)` isntead of sleep for being more accurate on what we are doing

mdrahman-suse: Updated

pkg/testcase/daemonset.go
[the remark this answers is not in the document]
ShylajaDevadiga: To add to Justin's comment, for better readability could we have this `getDeamonset+shared.KubeConfigFile` set like in rest of the commands?

pkg/testcase/pod.go
```diff
@@ -23,15 +24,24 @@ func TestPodStatus(
...
+	cmd := "kubectl get pods -A --field-selector=status.phase!=Running | " +
+		"kubectl get pods -A --field-selector=status.phase=Pending"
+	Eventually(func(g Gomega) bool {
 		pods, err := shared.GetPods(false)
 		g.Expect(err).NotTo(HaveOccurred())
 		g.Expect(pods).NotTo(BeEmpty())
 
+		res, _ := shared.RunCommandHost(cmd + " --kubeconfig=" + shared.KubeConfigFile)
+		if res != "" {
+			shared.LogLevel("info", "Pods not Running or Pending: \n%s", res)
```
ShylajaDevadiga, line 36: Waiting for the pod status to become "Running".

pkg/testcase/pod.go
```diff
@@ -65,6 +75,28 @@ func getPrivatePods(cluster *shared.Cluster) (podDetails string) {
...
+		if res != "" {
+			shared.LogLevel("info", "Waiting for pods: \n%s", res)
+			return false
+		}
+		for i := range pods {
+			processPodStatus(cluster, g, &pods[i], podAssertRestarts, podAssertReady)
+		}
+
+		return true
+	}, "600s", "10s").Should(BeTrue(), "failed to process pods status")
```
ShylajaDevadiga, line 97: Could we consider rephrasing it to 'Pod is not in the desired state,' since it seems that its the pod's status rather than the processing of it?

mdrahman-suse: Updated

pkg/testcase/pod.go
mdrahman-suse: Updated

ShylajaDevadiga (approved):

PR #209, opened 2025-01-27
fix cluster restore and cert rotate

fmoral2:
#### Testing ####

Checklist:


docs/examples/.env.example
```diff
@@ -1,34 +1,54 @@
 # Framework config variables
+
```
mdrahman-suse, line 2: Can we mention which are required and which are optional vars?

entrypoint/clusterrestore/clusterrestore_suite_test.go
```diff
@@ -30,7 +29,8 @@ func TestMain(m *testing.M) {
 	var err error
 	flags = &customflag.ServiceFlag
 	flag.Var(&flags.Destroy, "destroy", "Destroy cluster after test")
-	flag.StringVar(&flags.S3Flags.Bucket, "s3Bucket", "distros_qa", "s3 bucket to store snapshots")
+	flag.Var(&flags.Channel, "channel", "channel to use on install")
+	flag.StringVar(&flags.S3Flags.Bucket, "s3Bucket", "distros1", "s3 bucket to store snapshots")
```
mdrahman-suse, line 33: Can we provide like `distrosqa` or `distros-qa`?

docs/examples/.env.example
fmoral2: i believe i did that right

fmoral2: `### Variables that would need to change for different products,test cases,pem key ####`

fmoral2: `##### Variables that would not need to change ######`

entrypoint/clusterrestore/clusterrestore_suite_test.go
fmoral2: changed

docs/examples/.env.example
mdrahman-suse: It mentions what need to change, adding required vs optional sections would make it more clear in my opinion

pkg/testcase/certrotate.go
```diff
@@ -10,27 +10,64 @@ import (
...
-	ip, rotateError := shared.CertRotate(product, ips)
-	Expect(rotateError).NotTo(HaveOccurred(),
-		fmt.Sprintf("error running certificate rotate for %s service on %s", product, ip))
-
-	ip, startError := shared.ManageService(product, "start", "server", ips)
-	Expect(startError).NotTo(HaveOccurred(),
-		fmt.Sprintf("error starting %s service for node ip: %s", product, ip))
+func certRotate(ms *shared.ManageService, product string, ips []string) {
+	for _, ip := range ips {
+		actions := []shared.ServiceAction{
```
mdrahman-suse, line 42: Not sure if we should be passing it as an array... previous implementation was more descriptive

fmoral2: this is for cases u wanna pass more than one action
whats not descriptive here?

fmoral2: the older implementation was buggy.  receiving list of IPS instead of actions. Thats the "main " change.
now we can do in order the action we need on that node

docs/examples/.env.example
fmoral2: sure added

pkg/testcase/certrotate.go
fmoral2: then only after all "done"we proceed to next node. OF course there is other ways of doing that, but for me actually passing a slice its the more descriptive and clear one. since its more visual as well

mdrahman-suse: Sounds good, thnx!

mdrahman-suse (approved):
LGTM

VestigeJ (approved):

PR #212, opened 2025-02-05
Feat/ipv6only

mdrahman-suse:
#### Testing ####

Checklist:


shared/kubeconfig.go
```diff
@@ -153,7 +153,7 @@ func NewLocalKubeconfigFile(newServerIP, resourceName, product, localPath string
 
 // decodeKubeConfig decodes the kubeconfig and writes it to a local /tmp file.
 func decodeKubeConfig(kubeConfig string) (string, error) {
-	dec, err := base64.StdEncoding.DecodeString(kubeConfig)
+	dec, err := base64.StdEncoding.DecodeString(strings.ReplaceAll(kubeConfig, " ", ""))
```
fmoral2, line 156: nice improv :)

docs/examples/k3s.tfvars.example
```diff
@@ -1,7 +1,7 @@
 ## K3S variables -- Fill these in with your desired values.
 
-k3s_version    = "v1.27.4+k3s1"
-k3s_channel    = "latest"
+install_version    = "v1.27.4+k3s1"
```
fmoral2, line 3: so related to this.... this is something i have been talking like more than a year, since we are moving to more generic vars i think we should have only `one place` as source of truth for all `config vars/ env vars` like only in .env 
-  on that way is less prone to errors, forgetting stuff, and better management 
-  (+) on that way we can load vars/env in any provider freely , not just on TF

thoughts ?

mdrahman-suse: definitely, I am totally on board with generalizing some of these vars, hence I am moving more towards this approach...
So far I think we can do
```
install_mode
install_version
install_channel
install_method
```
For upgrades, we can just replace theses with corresponding values in upgrade*...

shared/kubeconfig.go
mdrahman-suse: Yeah got an error when copying b64 coded string because of spaces :)

docs/examples/k3s.tfvars.example
fmoral2: humm honestly i would say ,maybe we should create a "i dont remember the name of the process " thing that we used to do for bigger structural changes? because i feel that if we are already like touching it, we should have a single PR and work for that, and literally get ride of  all  tf.vars files.

wdyt?

mdrahman-suse: Yep agree, but for now as this is a new module, we can implement the change and will add todo in the code (jenkins file) once we update it. I will also create an issue for applying the change later. https://github.com/rancher/distros-test-framework/issues/216

mdrahman-suse: Updated docs and reverted changes

Makefile
```diff
@@ -120,3 +120,4 @@ pre-commit:
 	@go vet ./...
 	@golangci-lint run --tests ./...
 	@shellcheck modules/airgap/setup/*.sh
+	@shellcheck modules/ipv6only/scripts/*.sh
```
fmoral2, line 123: can we create a `make shell check` ? just to not start getting bigger and bigger this one

shared/airgap.go
```diff
@@ -289,8 +289,8 @@ func UpdateRegistryFile(cluster *Cluster, flags *customflag.FlagConfig) (err err
 	return nil
 }
 
-// DisplayAirgapClusterDetails executes and prints kubectl get nodes,pods on bastion.
-func DisplayAirgapClusterDetails(cluster *Cluster) {
+// LogClusterDetailsViaProxy executes and prints kubectl get nodes,pods on bastion.
+func LogClusterDetailsViaProxy(cluster *Cluster) {
```
fmoral2, line 293: can we name more accurately , idk something like `LogClusterDetailsFromBastion` ? 
although it behaves like a proxy , but i think can bring confusion

docs/examples/k3s.tfvars.example
VestigeJ: adr architecture decision record

pkg/testcase/ipv6only.go
```diff
@@ -0,0 +1,17 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/rancher/distros-test-framework/pkg/aws"
+	"github.com/rancher/distros-test-framework/pkg/testcase/support"
+	"github.com/rancher/distros-test-framework/shared"
+)
+
+func TestIPv6Only(cluster *shared.Cluster, awsClient *aws.Client) {
+	shared.LogLevel("info", "Setting up %s cluster on ipv6 only nodes...", cluster.Config.Product)
+	err := support.ConfigureIPv6OnlyNodes(cluster, awsClient)
+	Expect(err).NotTo(HaveOccurred(), err)
```
fmoral2, line 13: can we add a break line here? after your expect >?

pkg/testcase/pod.go
```diff
@@ -47,8 +47,8 @@ func TestPodStatus(
 	Expect(err).NotTo(HaveOccurred())
 }
 
-// TestAirgapClusterPodStatus test the status of the pods in the private cluster using custom assert functions.
-func TestAirgapClusterPodStatus(
+// TestPodStatusViaProxy test the status of the pods in the private cluster using custom assert functions.
+func TestPodStatusViaProxy(
```
fmoral2, line 51: same comment from Log via proxy

pkg/testcase/node.go
```diff
@@ -51,8 +51,8 @@ func TestNodeStatus(
 	Expect(err).NotTo(HaveOccurred())
 }
 
-// TestAirgapClusterNodeStatus test the status of the nodes in the private cluster using 2 custom assert functions.
-func TestAirgapClusterNodeStatus(
+// TestNodeStatusViaProxy test the status of the nodes in the private cluster using 2 custom assert functions.
+func TestNodeStatusViaProxy(
```
fmoral2, line 55: same comment from Log via proxy

fmoral2 (approved):

fmoral2 (approved):

PR #215, opened 2025-02-07
add rke2 e2e test report to qase.io

fmoral2:
#### Proposed Changes ####
- this will be ran daily at '0 21 * * *' on gh actions 
- golang and lint updated to fix know issue from  < 1.23  - which could causes data loss on for loops https://github.com/rancher/distros-test-framework/actions/runs/13197167919

#### Testing ####

Checklist:


cmd/qase/main.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/rancher/distros-test-framework/shared"
+)
+
+var (
+	fileName string
+	product  string
+)
+
+func main() {
+	flag.StringVar(&fileName, "f", "", "path to rke2/k3s e2e tests log file")
```
mdrahman-suse, line 17: Should we add a default filepath here and then append the product later on the path?

cmd/qase/main.go
```diff
@@ -0,0 +1,42 @@
...
+	flag.StringVar(&product, "p", "", "product name")
+	flag.Parse()
+
+	if product == "" {
+		shared.LogLevel("error", "-p flag is required")
```
mdrahman-suse, line 22: Suggest to add what the flag means here

cmd/qase/main.go
fmoral2: its there on the `usage` field

cmd/qase/main.go
fmoral2: https://github.com/rancher/distros-test-framework/pull/215/files#r1949047680

VestigeJ (approved):

ShylajaDevadiga (approved):

PR #220, opened 2025-02-18
e2e job

fmoral2:
#### Proposed Changes ####
-  adding e2e job to also grab from report dir
- adding.a quick check on tf to at least wait first instance to be ready

#### Testing ####

Checklist:

### DIFF
--- a/scripts/e2e_report.sh
+++ b/scripts/e2e_report.sh
@@ -5,18 +5,26 @@ set -e
 trap 'echo "Error on line $LINENO: $BASH_COMMAND"' ERR
 
 create_binary() {
-  go build -o processreport ./cmd/qase/main.go
+if ! command -v go &> /dev/null; then
+  echo "Error: golang is not installed"
+  exit 1
+fi
+
+  echo "Building process report binary..."
+  go build -o ./processreport ./cmd/qase/main.go
 }
 
 run_qase() {
+  echo "Processing $latest_log for $PRODUCT..."
   ./processreport -f "$latest_log" -p "$PRODUCT"
 }
 
+# Init variables.
 PRODUCT=
 latest_log=$(find  ./report -type f -name "rke2_*.log" -o -name "k3s_*.log" | sort -r | head -n 1)
 
 if [[ ! "$latest_log" =~ ^./report/(rke2|k3s)_.*\.log$ ]]; then
-  echo "Invalid log file name: $latest_log"
+  echo "Error: Invalid log file name: $latest_log"
   exit 1
 fi
 
@@ -27,17 +35,12 @@ else
 fi
 
 if [ -f "$latest_log" ]; then
-  echo "Uploading $latest_log to Qase"
+  echo "Found log file: $latest_log"
   create_binary
   run_qase
+  echo "Upload complete"
+  rm -f processreport
 else
-  echo "No log file found for $PRODUCT"
+  echo "Error: No log file found for $PRODUCT"
   exit 1
 fi
-
-cd ../report
-rm -f processreport
-
-
-
-


scripts/e2e_report.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+PS4='+(${LINENO}): '
+set -e
+trap 'echo "Error on line $LINENO: $BASH_COMMAND"' ERR
+
+create_binary() {
+  echo "Building processreport binary..."
+  go build -o processreport ./cmd/qase/main.go
```
mdrahman-suse, line 9: Can we add a check if go is available? else it will fail...

scripts/e2e_report.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+run_qase() {
+  echo "Processing $latest_log for $PRODUCT..."
+  ./processreport -f "$latest_log" -p "$PRODUCT"
```
mdrahman-suse, line 14: suggest adding check for the folder

modules/rke2/master/instances_server.tf
```diff
@@ -247,6 +247,24 @@ resource "aws_instance" "master2-ha" {
...
+  connection {
+    type        = "ssh"
+    user        = var.aws_user
+    host        = aws_instance.master.public_ip
+    private_key = file(var.access_key)
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "while ! kubectl get nodes | grep Ready; do sleep 5; done",
```
mdrahman-suse, line 262: Do we need the same for k3s? and also for the agent node for single node setup?

fmoral2: i think we should yeah add for k3s, just to be consistent . will push that and for the server +agent setup

scripts/e2e_report.sh
fmoral2: good , will do tks

scripts/e2e_report.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+# Init variables.
+PRODUCT=
```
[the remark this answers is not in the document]
fmoral2: yep it is. no warns

scripts/e2e_report.sh
fmoral2: good, will add also

fmoral2:
updated @mdrahman-suse

modules/k3s/master/instances_server.tf
```diff
@@ -270,6 +270,24 @@ resource "aws_instance" "master2-ha" {
...
+  connection {
+    type        = "ssh"
+    user        = var.aws_user
+    host        = aws_instance.master.public_ip
+    private_key = file(var.access_key)
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "while ! kubectl get nodes | grep Ready; do sleep 5; done",
```
mdrahman-suse, line 285: what is the timeout on this wait?

modules/rke2/master/instances_server.tf
mdrahman-suse: What is the timeout on this wait?

scripts/e2e_report.sh
mdrahman-suse: right, I missed that

mdrahman-suse:
One other question... else LGTM 👍

modules/rke2/master/instances_server.tf
fmoral2: good point....i will add for now just a bash timeout then, because adding in tf would also require a lots of refactor and adding in other places for consistency ( + i really dont know if in our version we do have native timeouts available )

fmoral2: that was intermittently working so to not spent much time on something totally not related, removed.

mdrahman-suse (approved):
LGTM

ShylajaDevadiga (approved):

PR #233, opened 2025-04-10
Add tags on resources, update ipv6 func

mdrahman-suse:
#### Proposed Changes ####

Update resource tags
Update ipv6 only funcs

#### Testing ####

Checklist:


VestigeJ (approved):
🟢 LGTM no public ipv6 though?

pkg/testcase/support/installer.go
```diff
@@ -0,0 +1,158 @@
+package support
```
fmoral2, line 1: nice!! so since we are refactoring and adding a new whole file to do that ( which we already do basically the same on node replace, and probably will do in more places ) can we add this on shared please and update usages?

pkg/testcase/support/installer.go
```diff
@@ -0,0 +1,158 @@
...
+	InstallOrEnable string
+}
+
+func buildInstallCmd(cluster *shared.Cluster, nodeType, ip string) (cmd string) {
+	product := cluster.Config.Product
+	args := ScriptArgs{
+		NodeOS:          os.Getenv("node_os"),
+		FQDN:            cluster.FQDN,
+		MasterIP:        cluster.ServerIPs[0],
+		Token:           token,
```
fmoral2, line 39: the var token here should be better create a new one for this file ( because here we using in one way and in the file it gets declared, using in another way )

pkg/testcase/support/installer.go
```diff
@@ -0,0 +1,158 @@
...
+		PublicIPv4:      ip,
+		PrivateIPv4:     "",
+		PublicIPv6:      "",
+		InstallChannel:  os.Getenv("install_channel"),
```
fmoral2, line 43: also for a broader usage i would say for us not rely that much on os env calls....
+ i cant see the actual gain using a struct to that , to basically just assign a bunch of os env to it
we should securely and properly load that somewhere only once, then populates the struct to be safely access where needed

pkg/testcase/support/installer.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return cmd
+}
+
+func buildK3SCmd(args *ScriptArgs, product, nodeType string) (cmd string) {
+	var script string
+	if nodeType == "master" && token == "" {
+		script = fmt.Sprintf("%v_%v.sh", product, nodeType)
+		cmd = fmt.Sprintf("sudo chmod +x %v; ", script)
+		cmd += fmt.Sprintf(
```
fmoral2, line 75: since on this file we are going to use a lot of strings replacement 
 can we create alike a helper func ?
 
something like `buildScriptCmd(script string, args ...string) string`  then we call this func something like 
```
script == 
 fmt.Sprintf("join_%s_%s.sh", product, nodeType)
	cmdArgs = []string{
           args.something
           args.other
}

buildScriptCmd(script, cmdArgs...)

```

pkg/testcase/support/installer.go
mdrahman-suse: Yeah sure can do, I was on the fence for leaving it here vs shared

pkg/testcase/support/installer.go
mdrahman-suse: Probably can store as part of loading tf vars

modules/ipv6only/scripts/prepare.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
aganesh-suse, line 2: nit: we could add some comments here explaining what this script is preparing.

aganesh-suse (approved):
LGTM

pkg/testcase/support/installer.go
mdrahman-suse: It has been removed as we need to discuss on how to configure these env/tf params to be reused on test runtime

pkg/testcase/support/installer.go
mdrahman-suse: https://github.com/rancher/distros-test-framework/pull/233#discussion_r2054391313

pkg/testcase/support/installer.go
mdrahman-suse: https://github.com/rancher/distros-test-framework/pull/233#discussion_r2054391313

fmoral2 (approved):

VestigeJ (approved):

PR #242, opened 2025-04-30
Add support to test secrets-encryption-provider

mdrahman-suse:
#### Testing ####

Checklist:

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -57,8 +57,11 @@ Please set the server_flags in .tfvars file for k3s:
 server_flags   = "secrets-encryption: true\n"
 ```
 
-For versions 1.26 and 1.27 - we run the traditional tests only: prepare/rotate/reencrypt (TEST_TYPE gets set to 'classic' in env var. We use this to determine which tests to run.)
-For versions 1.28 and greater - we run both the traditional tests and new method - rotate-keys (TEST_TYPE gets set to 'both' in env var)
+The test tag `secretsEncryptMethod` is optional which is set to `rotate-keys` by default. Other values that is accepted are `reencrypt`, which performs prepare/rotate/reencrypt operations and `both` which performs `prepare/rotate/reencrypt` and `rotate-keys` operations.
+Additionally starting v1.30.12 and above, support for secretbox has been added only for `rotate-keys` and below can be added on the `server_flags` to test that feature.
+```
+secrets-encryption-provider: secretbox
+```
 
 Note/TODO: k3s external db fails working with etcd only node. Refer: https://docs.k3s.io/datastore/ha
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -57,7 +57,7 @@ Please set the server_flags in .tfvars file for k3s:
 server_flags   = "secrets-encryption: true\n"
 ```
 
-The test tag `secretsEncryptMethod` is optional which is set to `rotate-keys` by default. Other values that is accepted are `reencrypt`, which performs prepare/rotate/reencrypt operations and `both` which performs `prepare/rotate/reencrypt` and `rotate-keys` operations.
+The test tag `secretsEncryptMethod` is optional which is set to `both` by default, which performs both `prepare/rotate/reencrypt` and `rotate-keys` operations. Other values that is accepted are `reencrypt`, which performs prepare/rotate/reencrypt operations and `rotate-keys` which performs only `rotate-keys` operations.
 Additionally starting v1.30.12 and above, support for secretbox has been added only for `rotate-keys` and below can be added on the `server_flags` to test that feature.
 ```
 secrets-encryption-provider: secretbox
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -57,7 +57,7 @@ Please set the server_flags in .tfvars file for k3s:
 server_flags   = "secrets-encryption: true\n"
 ```
 
-The test tag `secretsEncryptMethod` is optional which is set to `both` by default, which performs both `prepare/rotate/reencrypt` and `rotate-keys` operations. Other values that is accepted are `reencrypt`, which performs prepare/rotate/reencrypt operations and `rotate-keys` which performs only `rotate-keys` operations.
+The test tag `secretsEncryptMethod` is optional which is set to `both` by default, which performs both `prepare/rotate/reencrypt` and `rotate-keys` operations. Other values that is accepted are `classic`, which performs `prepare/rotate/reencrypt` operations and `rotate-keys` which performs only `rotate-keys` operations.
 Additionally starting v1.30.12 and above, support for secretbox has been added only for `rotate-keys` and below can be added on the `server_flags` to test that feature.
 ```
 secrets-encryption-provider: secretbox
--- a/entrypoint/secretsencrypt/encryption_suite_test.go
+++ b/entrypoint/secretsencrypt/encryption_suite_test.go
@@ -19,12 +19,15 @@ var (
 	qaseReport = os.Getenv("REPORT_TO_QASE")
 	kubeconfig string
 	cluster    *shared.Cluster
+	flags      *customflag.FlagConfig
 	cfg        *config.Env
 	err        error
 )
 
 func TestMain(m *testing.M) {
-	flag.Var(&customflag.ServiceFlag.Destroy, "destroy", "Destroy cluster after test")
+	flags = &customflag.ServiceFlag
+	flag.Var(&flags.Destroy, "destroy", "Destroy cluster after test")
+	flag.StringVar(&flags.SecretsEncrypt.Method, "secretsEncryptMethod", "rotate-keys", "method to perform secrets encryption")
 	flag.Parse()
 
 	cfg, err = config.AddEnv()
--- a/entrypoint/secretsencrypt/encryption_suite_test.go
+++ b/entrypoint/secretsencrypt/encryption_suite_test.go
@@ -27,7 +27,7 @@ var (
 func TestMain(m *testing.M) {
 	flags = &customflag.ServiceFlag
 	flag.Var(&flags.Destroy, "destroy", "Destroy cluster after test")
-	flag.StringVar(&flags.SecretsEncrypt.Method, "secretsEncryptMethod", "rotate-keys", "method to perform secrets encryption")
+	flag.StringVar(&flags.SecretsEncrypt.Method, "secretsEncryptMethod", "both", "method to perform secrets encryption")
 	flag.Parse()
 
 	cfg, err = config.AddEnv()
@@ -61,6 +61,12 @@ func validateSecretsEncryptFlag() {
 			shared.LogLevel("error", "Add secrets-encryption:true to server_flags for this test")
 			os.Exit(1)
 		}
+	}
+
+	if strings.Contains(os.Getenv("server_flags"), "secretbox") &&
+		flags.SecretsEncrypt.Method != "rotate-keys" {
+		shared.LogLevel("info", "secretbox provider is supported only with rotate-keys operation")
+		flags.SecretsEncrypt.Method = "rotate-keys"
 	}
 }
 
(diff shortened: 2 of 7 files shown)


VestigeJ (approved):

pkg/testcase/secretsencrypt.go
```diff
@@ -2,15 +2,17 @@ package testcase
...
 
+	"github.com/rancher/distros-test-framework/pkg/customflag"
 	"github.com/rancher/distros-test-framework/shared"
 
 	. "github.com/onsi/gomega"
 )
 
-func TestSecretsEncryption(cluster *shared.Cluster) {
+func TestSecretsEncryption(cluster *shared.Cluster, flags *customflag.FlagConfig) {
 	Expect(cluster.Status).To(Equal("cluster created"))
```
fmoral2, line 16: i know thats not your changes, but do you know why are we validating these 2 lines here again? we already do that on test build cluster

pkg/testcase/secretsencrypt.go
```diff
@@ -26,27 +28,58 @@ func TestSecretsEncryption(cluster *shared.Cluster) {
 
 	index := len(nodes) - 1
 	cpIp := nodes[index].ExternalIP
-	shared.LogLevel("info", "TEST: Old Method of Secrets-Encryption")
-	secretsEncryptOps("prepare", product, cluster.ServerIPs[0], cpIp, nodes)
-	secretsEncryptOps("rotate", product, cluster.ServerIPs[0], cpIp, nodes)
-	secretsEncryptOps("reencrypt", product, cluster.ServerIPs[0], cpIp, nodes)
-	shared.LogLevel("info", "TEST: New Method of Secrets-Encryption")
-	secretsEncryptOps("rotate-keys", product, cluster.ServerIPs[0], cpIp, nodes)
+	err = performSecretsEncryption(flags.SecretsEncrypt.Method, product, cluster.ServerIPs[0], cpIp, nodes)
```
fmoral2, line 31: do u mind create another error here? instead of shadowing the one from line 23 ?

just as best practice, i know that in practice will have no big impact here, just to maintain error handling clean, since the errors would not be related to each other here

pkg/testcase/secretsencrypt.go
```diff
@@ -26,27 +28,58 @@ func TestSecretsEncryption(cluster *shared.Cluster) {
...
+	Expect(err).NotTo(HaveOccurred(), "error performing secrets-encryption")
 }
 
-func secretsEncryptOps(action, product, primaryNodeIp, cpIP string, nodes []shared.Node) {
-	shared.LogLevel("info", "TEST: Secrets-Encryption: %v starts", action)
+func performSecretsEncryption(method, product, primaryNodeIp, cpIP string, nodes []shared.Node) (err error) {
```
fmoral2, line 35: only if you prefer but what u think reading like that instead?

```
func performSecretsEncryption(method, product, primaryNodeIp, cpIP string, nodes []shared.Node) (err error) {
	actions := []string{"prepare", "rotate", "reencrypt", "rotate-keys"}

	switch method {
	case "reencrypt":
		shared.LogLevel("info", "Performing secrets-encryption using prepare, rotate and reencrypt...")
		for _, action := range actions[:len(actions)-1] {
			err = secretsEncryptOps(action, product, primaryNodeIp, cpIP, nodes)
			if err != nil {
				return shared.ReturnLogError("error performing secrets-encrypt %s operation on node: %s!", action, cpIP)
			}
		}

	case "rotate-keys":
		shared.LogLevel("info", "Performing secrets-encryption using rotate-keys...")
		action := actions[len(actions)-1]
		err = secretsEncryptOps(action, product, primaryNodeIp, cpIP, nodes)
		if err != nil {
			return shared.ReturnLogError("error performing secrets-encrypt %s operation on node: %s!", action, cpIP)
		}

	case "both":
		for _, action := range actions {
			err = secretsEncryptOps(action, product, primaryNodeIp, cpIP, nodes)
			if err != nil {
				return shared.ReturnLogError("error performing secrets-encrypt %s operation on node: %s!", action, cpIP)
			}
		}

	default:
		return shared.ReturnLogError("unsupported method %s! Supported methods are: reencrypt, rotate-keys and both", method)
	}

	return nil
}
```

pkg/testcase/secretsencrypt.go
```diff
@@ -159,7 +197,17 @@ func verifyStatusStdOut(action, stdout string) {
 	}
 }
 
+func verifyStatusProvider(stdout string) {
```
fmoral2, line 200: since in the TestSecretsEncryption function we receive already the cluster , you can pass along and use  Cluster.config.ServerFlags

fmoral2: same on function bellow

pkg/testcase/secretsencrypt.go
mdrahman-suse: Sorry I did not get the point here.. the pasted code is the same

pkg/testcase/secretsencrypt.go
mdrahman-suse: Updated

fmoral2 (dismissed):

pkg/testcase/secretsencrypt.go
```diff
@@ -159,7 +195,21 @@ func verifyStatusStdOut(action, stdout string) {
...
+func verifyStatusProvider(serverFlags, stdout string) {
+	if strings.Contains(serverFlags, "secrets-encryption-provider: secretbox") {
+		Expect(stdout).To(ContainSubstring("XSalsa20-POLY1305"))
+	} else {
+		Expect(stdout).To(ContainSubstring("AES-CBC"))
+	}
+}
+
 func logEncryptionFileContents(nodes []shared.Node, action, product string) error {
+	serverFlags := os.Getenv("server_flags")
```
aganesh-suse, line 212: did we want to pass the server flags here too instead of getting from OS env?

docs/testing.md
```diff
@@ -57,8 +57,11 @@ Please set the server_flags in .tfvars file for k3s:
 server_flags   = "secrets-encryption: true\n"
 ```
 
-For versions 1.26 and 1.27 - we run the traditional tests only: prepare/rotate/reencrypt (TEST_TYPE gets set to 'classic' in env var. We use this to determine which tests to run.)
-For versions 1.28 and greater - we run both the traditional tests and new method - rotate-keys (TEST_TYPE gets set to 'both' in env var)
+The test tag `secretsEncryptMethod` is optional which is set to `rotate-keys` by default. Other values that is accepted are `reencrypt`, which performs prepare/rotate/reencrypt operations and `both` which performs `prepare/rotate/reencrypt` and `rotate-keys` operations.
```
aganesh-suse, line 60: we should have 'both' by default. unless the other method is taking off the testing.

docs/testing.md
```diff
@@ -57,8 +57,11 @@ Please set the server_flags in .tfvars file for k3s:
...
+Additionally starting v1.30.12 and above, support for secretbox has been added only for `rotate-keys` and below can be added on the `server_flags` to test that feature.
+```
+secrets-encryption-provider: secretbox
```
aganesh-suse, line 63: if 'secrets-encryption-provider: secretbox' has been set -  we could fail and let the user know what the values allowed are. (we hard switch to rotate-keys value if we want.)

entrypoint/secretsencrypt/encryption_suite_test.go
```diff
@@ -19,12 +19,15 @@ var (
...
 	cfg        *config.Env
 	err        error
 )
 
 func TestMain(m *testing.M) {
-	flag.Var(&customflag.ServiceFlag.Destroy, "destroy", "Destroy cluster after test")
+	flags = &customflag.ServiceFlag
+	flag.Var(&flags.Destroy, "destroy", "Destroy cluster after test")
+	flag.StringVar(&flags.SecretsEncrypt.Method, "secretsEncryptMethod", "rotate-keys", "method to perform secrets encryption")
 	flag.Parse()
```
aganesh-suse, line 32: we should have func that does a flag check here - for 
if 
secrets-encryption-provider: secretbox
then flags.SecretsEncrypt.Method should be rotate-keys only. 
(regardless of what the values are for default)
Also check if secrets-encryption-provider: secretbox should be only for versions over 1.30.12 - equivalent 1.31.x and 1.32.x etc.

entrypoint/secretsencrypt/encryption_suite_test.go
```diff
@@ -19,12 +19,15 @@ var (
...
+	flags      *customflag.FlagConfig
 	cfg        *config.Env
 	err        error
 )
 
 func TestMain(m *testing.M) {
-	flag.Var(&customflag.ServiceFlag.Destroy, "destroy", "Destroy cluster after test")
+	flags = &customflag.ServiceFlag
+	flag.Var(&flags.Destroy, "destroy", "Destroy cluster after test")
+	flag.StringVar(&flags.SecretsEncrypt.Method, "secretsEncryptMethod", "rotate-keys", "method to perform secrets encryption")
```
aganesh-suse, line 30: https://github.com/rancher/distros-test-framework/pull/242/files#r2089621941

pkg/testcase/secretsencrypt.go
mdrahman-suse: can do

entrypoint/secretsencrypt/encryption_suite_test.go
mdrahman-suse: So for the version check, it will be tricky as we then wont be able to validate using commits, although we hardly use it but prefer not to move away from it. I added a check to override the method for secretbox

docs/testing.md
mdrahman-suse: Updated

mdrahman-suse:
@aganesh-suse please review

pkg/testcase/secretsencrypt.go
```diff
@@ -26,27 +24,59 @@ func TestSecretsEncryption(cluster *shared.Cluster) {
...
+			if err != nil {
+				return shared.ReturnLogError("error performing secrets-encrypt %s operation on node: %s!", action, cpIP)
+			}
+		}
+	case "rotate-keys":
+		shared.LogLevel("info", "Performing secrets-encryption using rotate-keys...")
+		action := actions[len(actions)-1]
+		err = secretsEncryptOps(action, product, serverFlags, primaryNodeIp, cpIP, nodes)
+		if err != nil {
+			return shared.ReturnLogError("error performing secrets-encrypt %s operation on node: %s!", action, cpIP)
```
ShylajaDevadiga, line 48: Can we instead return "% secret-encrypt %s on node: %s failed", product, action, cpIP?  Provides clarity on what failed instead of notes format.  Can we log the actual err?

pkg/testcase/secretsencrypt.go
```diff
@@ -26,27 +24,59 @@ func TestSecretsEncryption(cluster *shared.Cluster) {
...
+	shared.LogLevel("info", "Secrets-Encryption %v action starting...", action)
 
 	_, errStatusB4 := shared.SecretEncryptOps("status", cpIP, product)
-	Expect(errStatusB4).NotTo(HaveOccurred(), "error getting secret-encryption status before action")
+	Expect(errStatusB4).NotTo(HaveOccurred(), "error: getting secret-encryption status before action")
 
 	stdOutput, err := shared.SecretEncryptOps(action, cpIP, product)
-	Expect(err).NotTo(HaveOccurred(), "error: secret-encryption: %v", action)
+	if err != nil {
+		return shared.ReturnLogError("error: performing secret-encryption: %v", action)
```
ShylajaDevadiga, line 72: if err != nil we would want content of the err to be logged.

pkg/testcase/secretsencrypt.go
```diff
@@ -102,27 +139,25 @@ func restartServerAndWait(ip, product string) {
 	}
 }
 
-func waitForHashMatch(cpIP, product string) (string, error) {
+func waitForHashMatch(cpIP, product string) (stdOut string, err error) {
```
ShylajaDevadiga, line 142: secretEncryptStatus instead of stdOut? stdout is not exactly but kind of a reserved word

mdrahman-suse:
@ShylajaDevadiga Updated with comments addressed, please review

endawkins (approved):
LGTM

PR #246, opened 2025-05-12
Fix/privreg test

mdrahman-suse:
#### Proposed Changes ####

- Add more static timeout for process to finish before installing podman
- Update handling registry_url flag

#### Testing ####

Checklist:

### DIFF
--- a/entrypoint/airgap/airgap_suite_test.go
+++ b/entrypoint/airgap/airgap_suite_test.go
@@ -50,7 +50,7 @@ func TestMain(m *testing.M) {
 func validateAirgap() {
 	serverFlags := os.Getenv("server_flags")
 	cniSlice := []string{"calico", "flannel"}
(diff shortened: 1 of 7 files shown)


entrypoint/airgap/airgap_suite_test.go
```diff
@@ -70,10 +70,16 @@ func validateAirgap() {
 		os.Exit(1)
 	}
 
-	if (cfg.Product == "k3s" && strings.Contains(serverFlags, "protect")) ||
```
fmoral2, line 73: sorry i only saw the above now....do you mind pull env info from where its already loaded?

cluster config and the cfg, err = config.AddEnv() already have all u need....you dont need to add extra (lots ) of os. calls to os.getenv

fmoral2 (approved):

aganesh-suse (approved):

PR #253, opened 2025-06-17
Remove helm install from build image to container

mdrahman-suse:
#### Proposed Changes ####

Removing helm install when creating image and moved to run the install on the container on test runtime

#### Testing ####

Checklist:

### DIFF
--- a/entrypoint/deployrancher/rancher_suite_test.go
+++ b/entrypoint/deployrancher/rancher_suite_test.go
@@ -84,12 +84,22 @@ func validateRancher() {
 		}
 	}
 
+	// Install helm
+	res, err := shared.InstallHelm()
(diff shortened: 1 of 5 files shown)


entrypoint/deployrancher/rancher_suite_test.go
```diff
@@ -84,8 +84,15 @@ func validateRancher() {
 		}
 	}
 
+	// Install helm
+	res, err := shared.InstallHelm()
```
ShylajaDevadiga, line 88: We don't do anything with the result, so _ ?

aganesh-suse (approved):

aganesh-suse (dismissed):

PR #254, opened 2025-06-19
Feature: Airgap cluster with Tarball and Windows agent

mdrahman-suse:
#### Testing ####

Checklist:

### DIFF
--- a/pkg/testcase/support/airgapwindows.go
+++ b/pkg/testcase/support/airgapwindows.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/rancher/distros-test-framework/pkg/customflag"
 	"github.com/rancher/distros-test-framework/shared"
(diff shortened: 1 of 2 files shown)


pkg/testcase/support/airgapwindows.go
```diff
@@ -30,6 +31,9 @@ func InstallOnAirgapAgentsWindows(cluster *shared.Cluster, airgapMethod string)
 		_, err := CmdForPrivateNode(cluster, cmd, agentIP)
 		Expect(err).To(BeNil(), err)
 	}
+
+	shared.LogLevel("info", "Sleep for 30 seconds while Windows node joins...")
```
fmoral2, line 35: should we add a simple retry to check that for a while? cause it also maybe for any unknown reason takes more than 30s, then the sleep would not be enough

fmoral2 (approved):

aganesh-suse (approved):
LGTM

PR #255, opened 2025-06-23
Add support for profile: etcd flag

mdrahman-suse:
#### Proposed Changes ####

Added support for new flag `profile: etcd` that allows hardening for etcd only
Ref: https://github.com/rancher/rke2/issues/8127

Likely needs doc update

#### Types of Changes ####

Updated rke2 install scripts in modules, not applicable for k3s

#### Testing ####

Checklist:


ShylajaDevadiga (approved):

modules/install/rke2_master.sh
```diff
@@ -132,6 +132,9 @@ EOF
       useradd -r -c "etcd user" -s /sbin/nologin -M etcd -U
     fi
   fi
+  if [ -n "$server_flags" ] && [[ "$server_flags" == *"etcd"* ]]; then
+    useradd -r -c "etcd user" -s /sbin/nologin -M etcd -U
```
fmoral2, line 136: arent we already adding that always(except micro)  on line 132 ?

modules/install/join_rke2_agent.sh
```diff
@@ -90,6 +90,9 @@ EOF
     fi
     systemctl restart systemd-sysctl
   fi
+  if [ -n "$worker_flags" ] && [[ "$worker_flags" == *"etcd"* ]]; then
+    useradd -r -c "etcd user" -s /sbin/nologin -M etcd -U
```
aganesh-suse, line 94: we wont have etcd in a worker node right?

aganesh-suse: https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/kubernetes-clusters-in-rancher-setup/checklist-for-production-ready-clusters/recommended-cluster-architecture

aganesh-suse:
we do not need this PR. the etcd user gets added for all OS but slemicro in the master nodes. and agent node should not have etcd in it.

modules/install/rke2_master.sh
mdrahman-suse: Yeah but that is only when profile is cis... for this scenario the profile is etcd so needed a separate flow

modules/install/join_rke2_agent.sh
mdrahman-suse: we dont but we allow the profile: etcd flag on worker as well

mdrahman-suse:
> we do not need this PR. the etcd user gets added for all OS but slemicro in the master nodes. and agent node should not have etcd in it.

We need it as we are sending a new flag that needs to be handled

aganesh-suse (approved):
1. Please update the branch. 
2. Also, test with slemicro. Am not sure if the slemicro needs the user added or the group is enough. Kindly check the same as well. 
LGTM otherwise.

mdrahman-suse:
Addressed as part of https://github.com/rancher/distros-test-framework/pull/300

Issue #245, opened 2025-05-02
Check nodename while taking snapshot on s3


ShylajaDevadiga:
After creating snapshot in s3, get the snapshotfiles and validate the nodename is listed correctly. 

```
ec2-user@ip-172-31-25-8:~> kubectl get etcdsnapshotfiles |grep s3 |grep 21-163
s3-on-demand-ip-172-31-21-163.us-east-2.compute.internal-1746221568-01d128    on-demand-ip-172-31-21-163.us-east-2.compute.internal-1746221568   ip-172-31-21-163.us-east-2.compute.internal   s3://sonobuoy-results/on-demand-ip-172-31-21-163.us-east-2.compute.internal-1746221568                               3723296    2025-05-02T21:32:48Z

ec2-user@ip-172-31-25-8:~> kubectl get etcdsnapshotfiles |grep s3 |grep 28-179
s3-on-demand-ip-172-31-28-179.us-east-2.compute.internal-1746221507-e4186b    on-demand-ip-172-31-28-179.us-east-2.compute.internal-1746221507   s3    
```

Issue #252, opened 2025-06-17
Add sestatus checks when rpm based OS is used


ShylajaDevadiga:
When rpm based OS is used let's report sestatus, rke2-selinux, container-selinux (rpm -qa|grep selinux)versions in the end of the test run.

Issue #244, opened 2025-05-02
Add reporting to Qase after test run is complete


mdrahman-suse:
Capture info and add the below logs to Qase after test execution is completed (regardless of pass or fail). Currently its adding the logs only when it fails
The logs should include
- Product version
- tfvars info
- cluster details
- cluster config

PR #264, opened 2025-07-31
Jjb qase.summary

fmoral2:
#### Proposed Changes ####
https://github.com/rancher/distros-test-framework/issues/244 

- adding report summary after all runs reported to qase

- add https://github.com/rancher/distros-test-framework/issues/252
 
- add fix and validation for  zypper os lock
```
      echo "Checking for package manager locks if so, removing them."
```

- add wait for instances 
```
aws ec2 wait instance-status-ok --region ${var.region} --instance-ids ${self.id}"
```

- fixed pod call returning value when was not correct on `pkg/testcase/pod.go`

- fix https://github.com/rancher/distros-test-framework/issues/245 

#### Testing ####

Checklist:

### DIFF
--- a/shared/clusterconfig.go
+++ b/shared/clusterconfig.go
@@ -258,7 +258,7 @@ func addClusterFromKubeConfig(nodes []Node) (*Cluster, error) {
 			Version:             getVersion(),
 			ServerFlags:         getFlags("server"),
 			WorkerFlags:         getFlags("worker"),
-			Channel:             getChanell(product),
+			Channel:             getChannel(product),
 			InstallMethod:       os.Getenv("install_method"),
 			InstallMode:         os.Getenv("install_mode"),
 			DataStore:           os.Getenv("datastore_type"),
@@ -290,6 +290,8 @@ func addClusterFromKubeConfig(nodes []Node) (*Cluster, error) {
 	}, nil
 }
 
+// getVersion retrieves the install version from environment variables.
+// used for addClusterFromKubeConfig().
 func getVersion() string {
 	installVersion := os.Getenv("install_version")
 	installVersionEnv := os.Getenv("INSTALL_VERSION")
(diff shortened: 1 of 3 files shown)


shared/terraform.go
```diff
@@ -76,22 +81,75 @@ func loadTFconfig(
...
+
+	if c.Config.Product == "rke2" {
+		c.Config.InstallMethod = terraform.GetVariableAsStringFromVarFile(t, varDir, "install_method")
 	}
+	c.Config.InstallMode = terraform.GetVariableAsStringFromVarFile(t, varDir, "install_mode")
 
 	return c, nil
 }
 
+// TODO: aux functions for loading data while we dont standardize from one source of truth,
```
fmoral2, line 102: here

modules/install/join_k3s_agent.sh
aganesh-suse, line 1: We should also do a slemicro test. just want to make sure these changes work just fine there as well.

shared/clusterconfig.go
```diff
@@ -239,3 +208,140 @@ func DestroyCluster(cfg *config.Env) (string, error) {
...
+	}
+
+	if flagsEnv != "" {
+		return flagsEnv
+	}
+
+	return ""
+}
+
+func getChanell(product string) string {
```
mdrahman-suse, line 323: `getChannel`

shared/clusterconfig.go
```diff
@@ -239,3 +208,140 @@ func DestroyCluster(cfg *config.Env) (string, error) {
...
+	if c != "" {
+		return c
+	}
+
+	channel := os.Getenv(product + "_channel")
+	if channel != "" {
+		return channel
+	}
+
+	return "stable"
```
mdrahman-suse, line 334: Should default to `testing`?

aganesh-suse (approved):

PR #271, opened 2025-08-22
add asset rke2 pkg count

fmoral2:
#### Testing ####

Checklist:


scripts/captain.sh
```diff
@@ -160,6 +160,33 @@ verify_release_asset_count_k3s () {
...
+     # expect 60 cause 2 assets are the zipped and tar.gz source code
+    EXPECTED_ASSETS_COUNT="${EXPECTED_ASSETS_COUNT:-60}"
+
+    printf '\n==== VERIFY RKE2 PACKAGING ASSETS FOR RKE2 VERSION: %s ====\n' "${VERSION}"
+
+    JSON=$(
+      curl -sS -H "Accept: application/vnd.github+json" \
+        "https://api.github.com/repos/rancher/rke2-packaging/releases?per_page=100"
+    )
+
```
aganesh-suse, line 173: nit: can we add a call to debug_log line here - which will print the exact query used for us to know if needed.

mdrahman-suse:
Can we change the name of the script to be like `asset_checker` or something more meaningful... `captain.sh` sounds weird... anyone should be able to run this not only captain

PR #304, opened 2026-01-09
add call to etcd user for cis

mdrahman-suse:
#### Testing ####

Checklist:

### DIFF
--- a/modules/install/join_rke2_agent.sh
+++ b/modules/install/join_rke2_agent.sh
@@ -72,15 +72,17 @@ update_config() {
   cat /etc/rancher/rke2/config.yaml
 }
 
-cis_setup() {
-  [[ -z "$worker_flags" || "$worker_flags" != *"cis"* ]] && return 0
-
-  cis_sysctl_file="/etc/sysctl.d/60-rke2-cis.conf"
-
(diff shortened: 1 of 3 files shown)


aganesh-suse (dismissed):

aganesh-suse:
if we are calling the ensure_etcd_user function anyway, is the if condition check still needed (line 155/162/119 in different files): 
```
  if [[ -n "$server_flags" ]] && [[ "$server_flags" == *"etcd"* ]]; then
      ensure_etcd_user
  fi
```

mdrahman-suse:
> if we are calling the ensure_etcd_user function anyway, is the if condition check still needed (line 155/162/119 in different files):
> 
> ```
>   if [[ -n "$server_flags" ]] && [[ "$server_flags" == *"etcd"* ]]; then
>       ensure_etcd_user
>   fi
> ```

If there is a cis flag then yes but we need one for etcd flag because if the `profile: etcd`, then we should not apply cis and then that func will not be called

ShylajaDevadiga (dismissed):

ShylajaDevadiga (approved):

endawkins (approved):

PR #322, opened 2026-03-31
Sec.fixes

fmoral2:
#### Testing ####

Checklist:


modules/install/join_k3s_agent.sh
```diff
@@ -119,19 +121,80 @@ install_k3s() {
...
-  fi
+  url="https://get.k3s.io"
+  install=("$install_mode=$version")
+
+  [[ -n "$channel" ]] && install+=("INSTALL_K3S_CHANNEL=$channel")
+  [[ "$install_or_enable" == "install" ]] && install+=("INSTALL_K3S_SKIP_ENABLE=true")
+  [[ "$install_mode" == "INSTALL_K3S_COMMIT" ]] && [[ -n "${github_token:-}" ]] && install+=("GITHUB_TOKEN=$github_token")
+
+# Download and execute an install script with basic validation.
+safe_install() {
```
fmoral2, line 141: .

fmoral2:
https://github.com/rancher/rancher-security/issues/1765

modules/airgap/setup/bastion_prepare.sh
```diff
@@ -47,7 +98,10 @@ install_kubectl() {
   else
       KUBE_ARCH="amd64"
   fi
-  curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/$KUBE_ARCH/kubectl" && \
+  KUBECTL_VERSION=$(curl -L -s https://dl.k8s.io/release/stable.txt)
```
manuelbuil: According to Gui and Chris guidelines, we should pin the Kubectl version and hardcode the SHA256 instead of downloading them

modules/ipv6only/scripts/prepare.sh
```diff
@@ -9,6 +9,10 @@ if [ "$arch" = "aarch64" ]; then
 else
     KUBE_ARCH="amd64"
 fi
-curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/$KUBE_ARCH/kubectl" && \
+
+KUBECTL_VERSION=$(curl -L -s https://dl.k8s.io/release/stable.txt)
```
manuelbuil: same here

fmoral2: https://github.com/rancher/distros-test-framework/pull/322/changes/0bda6f58383ab599900983002ebe6311c761e671

macedogm (commented):
I see that the functions `safe_install()` and `safe_download_install()` are defined multiple times. Would it make sense to have a helper script to reuse the same code in different scripts?

fmoral2:
> I see that the functions `safe_install()` and `safe_download_install()` are defined multiple times. Would it make sense to have a helper script to reuse the same code in different scripts?

doing that would require adding files and calls to the terraform so that we can reference the function in another file , and then the trouble of adding that and needing to have more tests around the framework intgrating with jenkins and etc its not worthy for this task

as a future enhancement yes

modules/airgap/setup/get_artifacts.sh
```diff
@@ -141,13 +227,15 @@ get_cni_assets() {
...
   url=$(get_url)
   echo "Download Windows assets using url: $url"
-  download_retry "wget $url/sha256sum-amd64.txt"
-  download_retry "wget $url/rke2-images.windows-amd64.txt"
-  download_retry "wget $url/rke2.windows-amd64.tar.gz"
-  download_retry "wget -O rke2-install.ps1 https://raw.githubusercontent.com/rancher/rke2/master/install.ps1"
+  download_retry wget $url/sha256sum-amd64.txt
+  download_retry wget $url/rke2-images.windows-amd64.txt
+  download_retry wget $url/rke2.windows-amd64.tar.gz
+  download_retry wget -O rke2-install.ps1 https://raw.githubusercontent.com/rancher/rke2/${version}/install.ps1
```
mdrahman-suse: Will this work? The docs says to download from master: https://docs.rke2.io/install/quickstart#1-download-the-install-script

scripts/release_checks.sh
```diff
@@ -48,6 +48,16 @@ fi
...
+trap 'rm -f \
+  "${FAILURE_FILE}" \
+  "sys_agent_installers_${RANDOM_INT}" \
+  "upgrade_images_${RANDOM_INT}" \
+  "releases_${RANDOM_INT}" \
+  "rke2_pkg_${RANDOM_INT}" \
+  "sys_agent_${RANDOM_INT}" \
+  "upgrade_${RANDOM_INT}" \
+  "rke2_runtime_${RANDOM_INT}" \
+  "lts_output_${RANDOM_INT}"' EXIT
```
mdrahman-suse, line 60: It looks like we missed the curl commands on this file @fmoral2 ?

modules/install/rke2_master.sh
```diff
@@ -179,19 +179,65 @@ install_rke2() {
   fi
```
mdrahman-suse, line 179: Dont we need the GH token here for commit installs?

modules/install/join_rke2_agent.sh
```diff
@@ -172,19 +172,65 @@ install_rke2() {
   fi
```
mdrahman-suse, line 172: https://github.com/rancher/distros-test-framework/pull/322/changes#r3047254389

mdrahman-suse:
@fmoral2 we are using the install scripts in ipv6Only test, so we need to update the GH Token param there as well, no?

modules/install/join_rke2_master.sh
```diff
@@ -186,19 +186,64 @@ install_rke2() {
   fi
```
mdrahman-suse, line 186: https://github.com/rancher/distros-test-framework/pull/322/changes#r3047254389

modules/airgap/setup/get_artifacts.sh
fmoral2: i think i tested it , but i will test again to double check

modules/install/rke2_master.sh
fmoral2: rke2 no

scripts/release_checks.sh
fmoral2: no need , its just a api read

fmoral2:
> @fmoral2 we are using the install scripts in ipv6Only test, so we need to update the GH Token param there as well, no?

@mdrahman-suse removing refrences for gh token since i added here by mistake..... this is not the scope of current PR

modules/airgap/setup/get_artifacts.sh
fmoral2: fixed

scripts/release_checks.sh
mdrahman-suse: But they were flagged here https://github.com/rancher/rancher-security/issues/1765#issue-4138984036, wont it be flagged again?

mdrahman-suse (approved):
Needs the comments resolved from other reviewers, LGTM otherwise

scripts/release_checks.sh
fmoral2: dont think so tehre is no much to do here afaik

ShylajaDevadiga (approved):

PR #327, opened 2026-05-06
Suc plan change and node status timeout update

aganesh-suse:
#### Proposed Changes ####

Problem at hand: for sles16, the system-upgrade jobs fail for one of the nodes with timeout at 15 mins. If we update the plan to wait upto 25 mins, we can see that the job for rke2-server or rke2-agent starts running around 18 to 20 mins time frame and exists quickly after. 
Also, the node status check we do times out at 10 mins. 
So the changes made: 
1. updated the timeout time in the suc plan.yamls
2. updated adding a drain force for all server/agents in the plan.yamls
3. Added an optional timeouts var to the TestNodeStatus and the TestNodeStatusUsingBastion functions that defaults to 600s and can be edited when called - for suc test alone, updated the call to be 1500s - which is upto 25 mins. 

We could potentially do the same for the pod status too if we need to change the timeouts individually from the tests instead of the main function. (Have nt done that - just a thought for potential improvement on this PR)

P.S: Used gemini to get the drain options for the suc plan update. 

#### Testing ####

Checklist:

5. Verify code lint; we should not have errors.

6. Update the documentation if needed.

7. Update makefile and docker run if adding new tests.

8. Run your tests at least 4 times with all configurations needed and possible.

9. If needed test with different os types.


fmoral2 (dismissed):

pkg/testcase/node.go
```diff
@@ -15,12 +15,18 @@ func TestNodeStatus(
 	cluster *shared.Cluster,
 	nodeAssertReadyStatus assert.NodeAssertFunc,
 	nodeAssertVersion assert.NodeAssertFunc,
+	timeouts ...string,
```
mdrahman-suse, line 18: Does this needs to be an array? looks like we are using only the first element anyways. Suggest adding a string var and check for nil.. if not nil then use that, else use the default?

workloads/amd64/rke2-suc-plan-splitroles.yaml
```diff
@@ -64,5 +73,7 @@ spec:
     args: ["prepare", "rke2-server-etcd"]
   drain:
     force: true
+    skipWaitForDeleteTimeout: 30 # Don't wait forever for pods to exit
+    timeout: 300s                # Force the drain to give up after 5 mins
   upgrade:
     image: rancher/rke2-upgrade
```
mdrahman-suse: nit: new line

workloads/amd64/rke2-suc-plan.yaml
```diff
@@ -38,5 +44,7 @@ spec:
     args: ["prepare", "rke2-server"]
   drain:
     force: true
+    skipWaitForDeleteTimeout: 30 # Don't wait forever for pods to exit
+    timeout: 300s                # Force the drain to give up after 5 mins
   upgrade:
     image: rancher/rke2-upgrade
```
mdrahman-suse: nit: new line

pkg/testcase/node.go
aganesh-suse: It is to make a variable arguments or make the function a `variadic function`. This way we dont have to edit all the calls made to the function as well to have either empty string or nil value passed to the function. From what i read online these are the rules for this type of usage: 
AI notes on the same: 
```
In Go, you can handle optional or variable arguments using variadic functions. These are defined by placing an ellipsis (...) before the parameter's type in the function signature
Key Characteristics
Optional by Default: 
When you define a variadic parameter, it is perfectly legal to call the function without passing any arguments for that parameter.
Must Be Last: The variadic parameter must always be the final parameter in your function's signature.
Handled as a Slice: Inside the function, the variadic arguments are treated as a slice of the specified type.
Nil if Empty: If zero arguments are passed, the variadic parameter will be nil. You should always check its length using len() before accessing specific indices to avoid runtime panics.
```

mdrahman-suse (dismissed):

endawkins (approved):
LGTM

mdrahman-suse (approved):

PR #329, opened 2026-05-12
Top command

endawkins:
#### Testing ####

Checklist:

### DIFF
--- a/entrypoint/rebootinstances/rebootinstances_test.go
+++ b/entrypoint/rebootinstances/rebootinstances_test.go
@@ -29,6 +29,10 @@ var _ = Describe("Test:", func() {
 			assert.PodAssertReady())
 	})
 
+	It("Verifies node CPU usage does not exceed 80% before reboot", func() {
+		testcase.TestNodeCPUUsageBelowThreshold(80, true, true)
+	})
+
 	It("Verifies ClusterIP Service", func() {
 		testcase.TestServiceClusterIP(true, true)
 	})
--- a/entrypoint/rebootinstances/rebootinstances_test.go
+++ b/entrypoint/rebootinstances/rebootinstances_test.go
@@ -53,8 +53,8 @@ var _ = Describe("Test:", func() {
 		)
 	})
 
-	It("Verifies node CPU usage stays below 80% after reboot", func() {
-		testcase.TestNodeCPUUsageBelowThreshold(80, true, true)
+	It("Verifies node CPU usage does not exceed 80% after reboot", func() {
+		testcase.TestNodeCPUUsageBelowThreshold(81, true, true)
 	})
 
 	It("Verifies ClusterIP Service after reboot", func() {
--- a/entrypoint/rebootinstances/rebootinstances_test.go
+++ b/entrypoint/rebootinstances/rebootinstances_test.go
@@ -29,9 +29,9 @@ var _ = Describe("Test:", func() {
 			assert.PodAssertReady())
 	})
 
-	// It("Verifies node CPU usage does not exceed 80% before reboot", func() {
-	// 	testcase.TestNodeCPUUsageBelowThreshold(80, true, true)
-	// })
+	It("Verifies node CPU usage does not exceed 80% before reboot", func() {
+		testcase.TestNodeCPUUsageBelowThreshold(80, true, true)
+	})
 
 	It("Verifies ClusterIP Service", func() {
 		testcase.TestServiceClusterIP(true, true)
--- a/entrypoint/rebootinstances/rebootinstances_test.go
+++ b/entrypoint/rebootinstances/rebootinstances_test.go
@@ -53,6 +53,10 @@ var _ = Describe("Test:", func() {
 		)
 	})
 
+	It("Verifies node CPU usage stays below 80% after reboot", func() {
+		testcase.TestNodeCPUUsageBelowThreshold(80, true, true)
+	})
+
 	It("Verifies ClusterIP Service after reboot", func() {
 		testcase.TestServiceClusterIP(true, true)
 	})
--- a/entrypoint/rebootinstances/rebootinstances_test.go
+++ b/entrypoint/rebootinstances/rebootinstances_test.go
@@ -29,9 +29,9 @@ var _ = Describe("Test:", func() {
 			assert.PodAssertReady())
 	})
 
-	It("Verifies node CPU usage does not exceed 80% before reboot", func() {
-		testcase.TestNodeCPUUsageBelowThreshold(80, true, true)
-	})
+	// It("Verifies node CPU usage does not exceed 80% before reboot", func() {
+	// 	testcase.TestNodeCPUUsageBelowThreshold(80, true, true)
+	// })
 
 	It("Verifies ClusterIP Service", func() {
 		testcase.TestServiceClusterIP(true, true)
--- a/entrypoint/rebootinstances/rebootinstances_test.go
+++ b/entrypoint/rebootinstances/rebootinstances_test.go
@@ -30,7 +30,7 @@ var _ = Describe("Test:", func() {
 	})
 
 	It("Verifies node CPU usage does not exceed 80% before reboot", func() {
-		testcase.TestNodeCPUThreshold(80, false, false)
+		testcase.TestNodeCPUThreshold(80, true, false)
 	})
 
 	It("Verifies ClusterIP Service", func() {
--- a/entrypoint/rebootinstances/rebootinstances_test.go
+++ b/entrypoint/rebootinstances/rebootinstances_test.go
@@ -58,7 +58,7 @@ var _ = Describe("Test:", func() {
 	})
 
 	It("Verifies node CPU usage does not exceed 80% after reboot", func() {
-		testcase.TestNodeCPUUsageBelowThreshold(81, true, true)
+		testcase.TestNodeCPUUsageBelowThreshold(80, true, true)
 	})
 
 	It("Verifies ClusterIP Service after reboot", func() {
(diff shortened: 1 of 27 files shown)


pkg/testcase/node.go
```diff
@@ -135,3 +137,103 @@ func TestNodeMetricsServer(applyWorkload, deleteWorkload bool) {
...
 	}
 }
+
+// TestNodeCPUUsageBelowThreshold fails when any node exceeds the provided CPU percentage.
+func TestNodeCPUUsageBelowThreshold(maxCPUPercent int, applyWorkload, deleteWorkload bool, timeouts ...string) {
+	var workloadErr error
+	if applyWorkload {
+		shared.LogLevel("info", "Deploying metrics-server...")
+		workloadErr = shared.ManageWorkload("apply", "metrics-server.yaml")
+		Expect(workloadErr).To(BeNil())
```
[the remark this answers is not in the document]
endawkins: This is not a user-configurable parameter.

mdrahman-suse:
Thanks @endawkins do we want to add this test post-upgrade?

pkg/aws/ec2.go
```diff
@@ -350,7 +350,7 @@ func (c Client) waitForInstanceRunning(instanceId string) error {
 	}
 
 	ticker := time.NewTicker(15 * time.Second)
-	timeout := time.After(5 * time.Minute)
+	timeout := time.After(10 * time.Minute)
```
mdrahman-suse: Any reason we are increasing this wait time?

entrypoint/rebootinstances/rebootinstances_test.go
```diff
@@ -29,6 +29,10 @@ var _ = Describe("Test:", func() {
 			assert.PodAssertReady())
 	})
 
+	It("Verifies node CPU usage does not exceed 80% before reboot", func() {
```
mdrahman-suse, line 32: I would suggest running this test while keeping the workloads and cleaning up only after the reboot. May be deploy the daemonset only post reboot and keep the other workloads running?

entrypoint/rebootinstances/rebootinstances_test.go
```diff
@@ -29,6 +29,10 @@ var _ = Describe("Test:", func() {
...
+		testcase.TestNodeCPUUsageBelowThreshold(80, false, false)
```
mdrahman-suse: Also we can shorten the func name `TestNodeCPUThreshold`. Also do we want to know memory threshold as well?

entrypoint/restartservice/restartservice_test.go
```diff
@@ -29,6 +29,10 @@ var _ = Describe("Test:", func() {
 			assert.PodAssertReady())
 	})
 
+	It("Verifies node CPU usage does not exceed 80% before service restarts", func() {
```
mdrahman-suse, line 32: I would suggest running this test while keeping the workloads and cleaning up only after the restart. May be deploy the daemonset only post restart and keep the other workloads running?

pkg/testcase/node.go
```diff
@@ -135,3 +138,114 @@ func TestNodeMetricsServer(applyWorkload, deleteWorkload bool) {
...
+	Expect(err).To(BeNil())
+	shared.LogLevel("info", "Test metrics-server workload pod is running")
+
+	timeout := "120s"
+	if len(timeouts) > 0 && timeouts[0] != "" {
+		timeout = timeouts[0]
+	}
+
+	Eventually(func(g Gomega) bool {
+		shared.LogLevel("info", "Querying node CPU usage with 'kubectl top node --no-headers'...")
```
mdrahman-suse: May be move it above eventually to call it once? And then log something like `Failed to get valid stats, Retrying...`

pkg/testcase/node.go
```diff
@@ -135,3 +138,114 @@ func TestNodeMetricsServer(applyWorkload, deleteWorkload bool) {
...
+		topNodeCmd := "kubectl top node --kubeconfig=" + shared.KubeConfigFile + " --no-headers"
+		res, err := shared.RunCommandHost(topNodeCmd)
+		g.Expect(err).To(BeNil())
+		g.Expect(strings.TrimSpace(res)).NotTo(Equal(""), "kubectl top node returned no data")
+		shared.LogLevel("info", "kubectl top node query completed:\n%s", res)
```
mdrahman-suse: Also this can be moved out or add debug logging

pkg/testcase/node.go
mdrahman-suse: Also this can be moved out or add debug logging

mdrahman-suse:
@endawkins Also can you check if it works with `debug: true` on config.yaml?

entrypoint/rebootinstances/rebootinstances_test.go
endawkins: We may want to add that later but right now this is directly addressing a customer issue for node CPU% usage. However, they do tend to act independently of one another so low CPU% usage doesn't necessarily mean low memory% usage.

entrypoint/restartservice/restartservice_test.go
endawkins: The task was to add a check both after initial install (due to potential spikes of CPU% usage) as well as after the cluster-disruption. But I can change to have all of the workloads running before the restart and be cleaned up after. Should I do the same with reboot?

pkg/testcase/node.go
endawkins: This logging already exists at line 210 for if the CPU usage information is in an invalid syntax etc, I'll add ```retrtying...``` to that logging.

entrypoint/rebootinstances/rebootinstances_test.go
fmoral2: also from where is this  80% threshold pulled from ? 
in my pov is too high

pkg/testcase/node.go
```diff
@@ -135,3 +138,112 @@ func TestNodeMetricsServer(applyWorkload, deleteWorkload bool) {
...
+
+	if deleteWorkload {
+		shared.LogLevel("info", "Cleaning up test metrics-server workload...")
+		workloadErr = shared.ManageWorkload("delete", "metrics-server.yaml")
+		Expect(workloadErr).To(BeNil())
+		shared.LogLevel("info", "Test workload cleaned up successfully")
+	}
+}
+
+func parseNodeCPUPercentages(output string) (map[string]int, error) {
```
fmoral2: this functiona nd the bellow one, should live in a shared folder/file

entrypoint/rebootinstances/rebootinstances_test.go
fmoral2: i agree with @mdrahman-suse  suggestion, it will be a small addition and a good gain, since its related.

pkg/testcase/node.go
endawkins: Done

shared/cluster.go
```diff
@@ -888,3 +890,70 @@ func CleanupPod(podName string) error {
 
 	return nil
 }
+
+// ParseNodeCPUPercentages parses the output of 'kubectl top node' and returns a map of node names to CPU percentages.
+func ParseNodeCPUPercentages(output string) (map[string]int, error) {
```
fmoral2: this should not be exported right ?

pkg/aws/ec2.go
endawkins: Fixed.

entrypoint/rebootinstances/rebootinstances_test.go
endawkins: This was set based off of conversation with @ShylajaDevadiga but we can discuss a better percentage for the threshold.

pkg/testcase/node.go
endawkins: Fixed.

pkg/testcase/node.go
endawkins: Fixed.

shared/cluster.go
endawkins: fixed.

shared/cluster.go
```diff
@@ -888,3 +890,70 @@ func CleanupPod(podName string) error {
 
 	return nil
 }
+
+// ParseNodeCPUPercentages parses the output of 'kubectl top node' and returns a map of node names to CPU percentages.
```
mdrahman-suse: update to match func name

shared/cluster.go
```diff
@@ -888,3 +890,70 @@ func CleanupPod(podName string) error {
...
+		LogLevel("info", "  Parsed node %s with CPU usage %d%%", nodeName, cpuPercent)
+	}
+
+	LogLevel("info", "Parsed %d lines total, extracted %d nodes", lineCount, len(nodeCPU))
+
+	return nodeCPU, nil
+}
+
+// CheckNodeCPUThreshold returns a list of nodes exceeding maxCPUPercent, or nil if all nodes pass.
+func CheckNodeCPUThreshold(maxCPUPercent int, output string) ([]string, error) {
```
mdrahman-suse, line 934: Wondering if these new func should be in shared or testcase/node ... could also be in testcase/support/node.go, wdyt? cc @fmoral2

mdrahman-suse: As this is specific to cluster nodes only?

entrypoint/rebootinstances/rebootinstances_test.go
endawkins: Fixed

entrypoint/rebootinstances/rebootinstances_test.go
endawkins: Done.

entrypoint/restartservice/restartservice_test.go
endawkins: Done.

shared/cluster.go
endawkins: Keeping as is. Done.

fmoral2 (approved):

aganesh-suse (approved):

PR #333, opened 2026-06-17
fix for runtime env change

endawkins:
#### Testing ####

Checklist:

### DIFF
--- a/entrypoint/deployrancher/rancher_suite_test.go
+++ b/entrypoint/deployrancher/rancher_suite_test.go
@@ -86,22 +86,13 @@ func validateRancher() {
 		}
 	}
 
-	arch, archErr := shared.RunCommandHost("uname -m")
-	if archErr != nil {
-		shared.LogLevel("debug", "unable to detect runner architecture: %v", archErr)
-	}
-	shared.LogLevel(
-		"debug",
-		"helm preflight context: uname -m=%s PATH=%s",
-		strings.TrimSpace(arch),
-		os.Getenv("PATH"),
-	)
+	shared.LogLevel("debug", "helm preflight PATH=%s", os.Getenv("PATH"))
 
-	// Validate helm is available in the runner image.
-	res, err := shared.VerifyHelmAvailable()
+	// Ensure helm is available, installing it from vendored tarball if needed.
+	res, err := shared.InstallHelm()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


entrypoint/deployrancher/rancher_suite_test.go
```diff
@@ -86,11 +86,22 @@ func validateRancher() {
 		}
 	}
 
-	// Install helm
-	res, err := shared.InstallHelm()
+	arch, archErr := shared.RunCommandHost("uname -m")
```
fmoral2: why do u need to pull arch if you are not doing anything usefull with it ?  better avoid 15 lines of code for just a logging that doesnt change any next decision

fmoral2:
Dockerfile Approach

  For a hotfix, I think not fixing via Dockerfile is safer and easier.

  The Dockerfile approach is okay as a longer-term cleanup if the project wants Helm baked into the image, but it has operational risk: image must be rebuilt and every Jenkins/local
  runner must use that new image. If anyone runs deploy-rancher outside that exact image, the test fails earlier than before.

  The safer hotfix is source-level:

  - keep InstallHelm
  - first check command -v helm
  - if missing, install from the existing vendored tarball
  - after installing into ~/bin, prepend ~/bin to PATH in the Go process so later bare helm calls work

shared/aux.go
```diff
@@ -234,20 +237,34 @@ func InstallHelm() (res string, err error) {
...
+		"%v/helm version --short", localbin, BasePath(), arch, arch, localbin, localbin, localbin)
+
+	res, err = RunCommandHost(cmd)
+	if err != nil {
+		return res, err
+	}
+
+	pathEnv := os.Getenv("PATH")
+	homeBin := fmt.Sprintf("%v/bin", homedir)
+	if !strings.Contains(pathEnv, homeBin) {
```
fmoral2: if !slices.Contains(filepath.SplitList(pathEnv), homeBin) {

mdrahman-suse:
@endawkins I have created a PR to have consistent implementation similar to: https://github.com/rancher/distros-test-framework/blob/main/scripts/install_sonobuoy.sh#L16. This PR can be closed
https://github.com/rancher/distros-test-framework/pull/335

endawkins:
Closing as MD has provided a better fix for the same issue.

PR #343, opened 2026-08-25
testing ssh changes

aganesh-suse:
#### Testing ####

Checklist:

### DIFF
--- a/shared/aux.go
+++ b/shared/aux.go
@@ -154,8 +154,8 @@ func RunScp(c *Cluster, ip string, localPaths, remotePaths []string) error {
 	for i, localPath := range localPaths {
 		remotePath := remotePaths[i]
 		scp := fmt.Sprintf(
-			"ssh-keyscan %[1]s >> /root/.ssh/known_hosts && "+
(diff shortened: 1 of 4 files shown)


shared/aux.go
```diff
@@ -154,22 +155,38 @@ func RunScp(c *Cluster, ip string, localPaths, remotePaths []string) error {
 	for i, localPath := range localPaths {
 		remotePath := remotePaths[i]
 		scp := fmt.Sprintf(
-			"ssh-keyscan %[1]s >> /root/.ssh/known_hosts && "+
-				"scp -i %[2]s -o StrictHostKeyChecking=no %[3]s %[4]s@%[1]s:%[5]s",
+			"ssh-keyscan -t rsa,ecdsa,ed25519 %[1]s >> /root/.ssh/known_hosts && "+
```
fmoral2: this will not work, this does not fix the issue.....this only try to reduce the number of "trips " if u want a better approach just look how qa infra branch  is doing that

fmoral2 (changes requested):

